PR #11, opened 2015-02-28
Pull logic from edit.js in lexicon app into comment and comments-right...

pqh2:
* Create comments-right-panel directive that encapsulates the logic for the right hand side of the commenting page.


megahirt:
Patrick, I will let you delete the comments-directive branch on your repo.

PR #51, opened 2016-12-16
Fix navigation menus that caused a horizontal scrollbar	

Nateowami:
Fixes the horizontal scrollbar caused by the menu being hidden off screen and to the right.

The extent of the scrollbar was dependent on the length of the username in the menu. Longer usernames moved the menu farther left, sometimes making the scrollbar disappear.

Also re-adds the dropdown triangles on the menus:
screenshot from 2016-12-16 10-29-27
On the live branch they show up, and the menus fade. On master, as of commit 53ae518795cf2d4812b6b68385242b5bcd1e3365, the little arrows and fading of the menu nolonger happened (That commit is a squashed commit. The squashed commit where the changes of interest are from is 1174e29dafd04349a87512d2b9460cca17b53604).

That commit (1174e29dafd04349a87512d2b9460cca17b53604) also upgrades from jQuery 1.x to 3.x. I'm guessing that could be related, but I'm not going to investigate it more.

### DIFF
--- a/src/Site/views/languageforge/container/languageforge.html.twig
+++ b/src/Site/views/languageforge/container/languageforge.html.twig
@@ -8,20 +8,20 @@
                 {% if app.website.domain == "languageforge.local" %} (Bootstrap 2){% endif %}
             </div>
             <div id="header-nav" class="pull-right">
-                <ul class="sf-menu">
+                <ul class="sf-menu sf-arrows">
                     <li><a href="/">Home</a></li>
                     {% if showHelpButton %}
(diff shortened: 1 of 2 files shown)


cambell-prince:
Wrong fork.  See sil-student-projects.

Nateowami:
This actually was intended for this fork, by the way (since before jterm), but inasmuch as the top nav menu was changed in the BS4 migration, I'm not sure how valuable this is anymore since it would be ephemeral.

PR #53, opened 2016-12-30
feature/deploy: Updated deploy for Xenial desktop and server with Vagrant

cambell-prince:
This is the deploy system tested with the Thailand January 2017 group and related videos.


megahirt:
Cool.  @cambell-prince I assume these are all changes that reflect feedback from TU students who have already setup their dev environment?

PR #65, opened 2017-01-20
feature/jterm2017 bootstrap4

cambell-prince:
@megahirt  Here's the cleaned up bootstrap 4 branch.  There's a few things to tidy on the way, so don't just push the green button!

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/65)


src/Api/Model/Shared/Command/UserCommands.php
```diff
@@ -53,7 +53,7 @@ public static function updateUser($params, $website)
         }
 
         UserCommands::assertUniqueIdentity($user, $params['username'], $params['email']);
-        $user->setProperties(UserModel::$ADMIN_ACCESSIBLE, $params);
+        $user->setProperties(UserModel::ADMIN_ACCESSIBLE, $params);
```
cambell-prince: Need to revert this one, its been merged the wrong way. I plan on removing this commit and will force push.

gulpfile.js
```diff
@@ -678,8 +678,6 @@ gulp.task('build-minify', function () {
     'src/angular-app/bellows/**/*.js',
     'src/angular-app/container/**/*.js',
     'src/angular-app/' + params.applicationName + '/**/*.js',
-    '!src/angular-app/languageforge/rapid-words/**', // ng2
-    '!src/angular-app/languageforge/review-suggest/**', // ng2
     '!src/angular-app/**/*.min.js',
```
cambell-prince: @megahirt This is an exclusion to stop the ng2 from being built in the same way as ng1.  So long as we remember to put the exclusion back in when we do ng2 apps removing the exclusion is fine.

PR #85, opened 2017-02-21
Miscilaneous authentication and mobile fixes

Nateowami:
- Changes "Bad credentials" to "Invalid username or password." I'm not sure if there's a better way to do this, but I don't see one.

- Apparently I made the mobile logged-out header title show up even on large screens in c5bbcedf6e55f1e776d583221b4ccbb81270d8f7. This is now fixed.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/85)

### DIFF
--- a/src/Site/Controller/Auth.php
+++ b/src/Site/Controller/Auth.php
@@ -80,6 +80,7 @@ class Auth extends App
     {
         $this->data['last_username'] = $app['session']->get(Security::LAST_USERNAME);
         $errorMsg = $app['security.last_error']($request);
+        if($errorMsg == 'Bad credentials.') $errorMsg = 'Invalid username or password.';
         if ($errorMsg) {
             $app['session']->getFlashBag()->add('errorMessage', $errorMsg);
             if ($app['session']->has(SecurityContextInterface::AUTHENTICATION_ERROR)) {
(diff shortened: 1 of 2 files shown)


megahirt:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---

*[src/Site/Controller/Auth.php, line 83 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/85#-KdUhd9EVUAPaDnW8sI6:-KdUhd9EVUAPaDnW8sI7:b-q1ivu7) ([raw file](https://github.com/sillsdev/web-languageforge/blob/ae5df70eb3c94db6eab6e4bb516ca43f4c154ec5/src/Site/Controller/Auth.php#L83)):*
> ```PHP
>         $this->data['last_username'] = $app['session']->get(Security::LAST_USERNAME);
>         $errorMsg = $app['security.last_error']($request);
>         if($errorMsg == 'Bad credentials.') $errorMsg = 'Invalid username or password.';
> ```

This is fine with me for now.  It's possible that this message is localizable, in which case we might be able to change the text to our preferred method through the localization method.  But I don't know how that works or if it exists, so this is fine with me at the moment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/85#-:-KdUi9yRbFReACELXs61:bnfp4nl)*

PR #110, opened 2017-05-18
Offline support using app cache [WIP]

Nateowami:
Some functionality working. One test is still broken. Work in progress, not yet ready to merge.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/110)


megahirt:
This is great work Nathaniel.  I'm very glad you've gotten this far.  This offline feature will benefit many people I believe.

---

Reviewed 19 of 19 files at r1.

---

*[src/Api/Service/Sf.php, line 658 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110#-KkPursRDPMmGlFRbhEs:-KkPursRDPMmGlFRbhEt:by0oqbj) ([raw file](https://github.com/sillsdev/web-languageforge/blob/f25574a874b650b6e424773d2eeb854253635767/src/Api/Service/Sf.php#L658)):*
> ```PHP
>     }
> 
>     public function lex_dbeDtoFull($browserId, $projectId, $offset)
> ```

why does this method need projectId while the others don't?  Can't we still get projectId from the session and everything will be OK?  I'd like to keep the API surface consistent at this point, that is no projectId parameter.  As you've already fixed up the JSONRPC call to pass in projectId separately this should in theory work.

---

*[src/Site/views/shared/page/cache.manifest.html.twig, line 58 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110#-KkPvhlQrwNkpGEMFQMk:-KkPvhlQrwNkpGEMFQMl:bz563y6) ([raw file](https://github.com/sillsdev/web-languageforge/blob/f25574a874b650b6e424773d2eeb854253635767/src/Site/views/shared/page/cache.manifest.html.twig#L58)):*
> ```Twig
> 
> # Hard coded because for some reason not everything needed is included above
> /angular-app/bellows/js/bellows.services.project.js?v={{ version }}
> ```

this is probably only necessary because of the xforge.navbarApp.js which has a dependency on bellows.services.  The navbarApp was designed to be a second angularJS app in the frame, independent of the angular-app being served in the main page.  the navbarApp isn't actually even shipped as a feature, so we might get rid of it at some point.  I'm fine leaving this hardcoded along with the navbarApp.js

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110#-:-KkPsKAc37iRO6NkPIAr:bf4iyku)*

Nateowami:

---

*[src/Api/Service/Sf.php, line 658 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110#-KkPursRDPMmGlFRbhEs:-KkTVY24HtC4qSpesb5z:bwrxv9t) ([raw file](https://github.com/sillsdev/web-languageforge/blob/f25574a874b650b6e424773d2eeb854253635767/src/Api/Service/Sf.php#L658)):*

why does this method need projectId while the others don't?  Can't we still get projectId from the session and everything will be OK?  I'd like to keep the API surface consistent at this point, that is no projectId parameter.  As you've already fixed up the JSONRPC call to pass in projectId separately this should in theory work.
</blockquote></details>

Whoops! Good catch; that was left over from experimenting and should not have been committed. Does Reviewable play nicely with rebases? If so I'll do an interactive rebase to remove it. If not I can just push another commit fixing it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110)*

megahirt:

---

*[src/Api/Service/Sf.php, line 658 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110#-KkPursRDPMmGlFRbhEs:-KkTvIRJIqVNaKPLMQf2:b-3a80w9) ([raw file](https://github.com/sillsdev/web-languageforge/blob/f25574a874b650b6e424773d2eeb854253635767/src/Api/Service/Sf.php#L658)):*

Whoops! Good catch; that was left over from experimenting and should not have been committed. Does Reviewable play nicely with rebases? If so I'll do an interactive rebase to remove it. If not I can just push another commit fixing it.
</blockquote></details>

yes, a rebase is fine and reviewable should cope.  Thanks.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110)*

src/Site/Controller/Upload.php
```diff
@@ -53,33 +53,28 @@ public function receive(Application $app, $appType, $mediaType)  { // e.g. 'lf',
                     $response = $api->sfChecks_uploadFile($mediaType, $tmpFilePath);
                 } elseif ($appType == 'lf-lexicon') {
                     $api = new Sf($app);
+                    $jsonResult = array(
+                        'orderedParams' => array(
+                            $mediaType,
+                            $tmpFilePath
+                        ),
+                        'projectId' => $_POST['projectId']
```
Nateowami: As of this morning there was only one failing e2e test. Now there are five, since the fix involved requiring a `projectId`. They're all in `lexicon-new-project.spec.js`. I think they're probably all related to this line, where we try to access projectId, which may not exist (but ought to).

megahirt:
I made some comment below to the effect of "we've taken this require projectId too far in checkPermissions".  In a sense, having it throw in checkPermissions has been nice to show you exactly where in the API calls you haven't injected it, but actually not all API calls should have it.  I think I've covered that in my comment.  So, for the upload controller specifically, the upload audio DOES require a projectId (project context) whereas the upload a LIFT file does not require one and should not (because it doesn't make sense and does NOT have a project context.  Maybe you can come up with a good solution.

---

Reviewed 5 of 8 files at r2.

---

*[src/Api/Service/Sf.php, line 658 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110#-KkPursRDPMmGlFRbhEs:-KkUye8Mo0ecTEy4OiKt:bg7igkv) ([raw file](https://github.com/sillsdev/web-languageforge/blob/f25574a874b650b6e424773d2eeb854253635767/src/Api/Service/Sf.php#L658)):*

yes, a rebase is fine and reviewable should cope.  Thanks.
</blockquote></details>

:lgtm: 

---

*[src/Site/Controller/Upload.php, line 61 at r2](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110#-KkUeLepJEefKPi2JkNR-r2-61:-KkUyx-LEsIUiKlzEmqB:b-795olf) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2c0c9687da49ad1037797a59b988971fd1f38d9b/src/Site/Controller/Upload.php#L61)):*

As of this morning there was only one failing e2e test. Now there are five, since the fix involved requiring a `projectId`. They're all in `lexicon-new-project.spec.js`. I think they're probably all related to this line, where we try to access projectId, which may not exist (but ought to).
</blockquote></details>

Nathaniel, unfortunately I realize now that we've taken the "require projectId" too far.  There are some API methods that outside the scope of the project.  e.g. update user profile or upload a file in the process of creating a new project.   These API methods are outside of the project context and therefore do not have a projectId.  The reasons it's mostly working is likely because the previous projectId is already in the session.  For example, if you visit project A, then go back to My Projects, then click on create new project and get to a new project wizard to upload a LIFT file, the new project wizard has not project context, because the purpose of the wizard is to gather enough data to create a new project (which doesn't yet exist).  Any API calls that that work in this context but still insert a projectId work because they have a previous projectId already in the session, the projectId of project A above.  So, we need to think this through a bit more.  We can't enforce a "you must have a projectId in checkPermissions because not all API methods require a project context.  Does this make sense?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110#-:-KkV-9G6FqLyemw5gTHX:bs94gun)*

Nateowami:

---

*[src/Site/Controller/Upload.php, line 61 at r2](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110#-KkUeLepJEefKPi2JkNR-r2-61:-Kkuq8mmWwv7h_gP8UBc:b7ut9t6) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2c0c9687da49ad1037797a59b988971fd1f38d9b/src/Site/Controller/Upload.php#L61)):*

Nathaniel, unfortunately I realize now that we've taken the "require projectId" too far.  There are some API methods that outside the scope of the project.  e.g. update user profile or upload a file in the process of creating a new project.   These API methods are outside of the project context and therefore do not have a projectId.  The reasons it's mostly working is likely because the previous projectId is already in the session.  For example, if you visit project A, then go back to My Projects, then click on create new project and get to a new project wizard to upload a LIFT file, the new project wizard has not project context, because the purpose of the wizard is to gather enough data to create a new project (which doesn't yet exist).  Any API calls that that work in this context but still insert a projectId work because they have a previous projectId already in the session, the projectId of project A above.  So, we need to think this through a bit more.  We can't enforce a "you must have a projectId in checkPermissions because not all API methods require a project context.  Does this make sense?
</blockquote></details>

I agree, it needs to get pulled out again. That's probably the first thing I'll be working on. Likely after that the E2E tests will pass, but only one way to be sure.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110)*

Nateowami:
Review status: 18 of 19 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/Site/Controller/Upload.php, line 61 at r2](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110#-KkUeLepJEefKPi2JkNR-r2-61:-KkxnV5T876zFzdomtn5:bwmdsjb) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2c0c9687da49ad1037797a59b988971fd1f38d9b/src/Site/Controller/Upload.php#L61)):*

I agree, it needs to get pulled out again. That's probably the first thing I'll be working on. Likely after that the E2E tests will pass, but only one way to be sure.
</blockquote></details>

I've changed this to make it optional. Now working on the one remaining failing test.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110)*

test/app/languageforge/lexicon/newLexProject/e2e/lexicon-new-project.spec.js
```diff
@@ -363,7 +363,6 @@ describe('E2E testing: New Lex Project wizard app', function () {
       expect(page.backButton.isDisplayed()).toBe(false);
       expect(page.initialDataPage.browseButton.isDisplayed()).toBe(true);
       expect(page.progressIndicatorStep3Label.getText()).toEqual('Verify');
-      page.expectFormIsNotValid();
```
Nateowami: I looked into this pretty deeply for a good while before deciding removing the assertion was the best course of action. It appears the form validation logic considers there to be three states: valid, invalid, and somewhere in between (for the condition of no input perhaps?). When one progresses quickly past the "project name" state to the "import project" state the form is marked valid without ever uploading anything (it's possible to skip the step though, so it makes sense to consider it valid).

From the code it appears this was the intended behavior, but it's not what the tests look for, and because validation isn't run until there is user input, it's not usually the behavior. I'm not sure why this wasn't happening before (on my machine it even happens on master, whereas it wasn't on my previous dev machine). It's also inconsistent (it's a race condition).

There's basically the same issue below, where it again asserts that the form is not valid but that it's valid enough to move on.

Nateowami:
E2E tests are passing! :tada: I don't think this is ready to be merged into master though, because the session is still being permanently cached. I think that's the next step, extracting the session from the page, storing the data in IndexedDB so it's available offline, and calling the server to update it when there's a connection.

test/app/languageforge/lexicon/newLexProject/e2e/lexicon-new-project.spec.js
irahopkinson: The intention was that the **Next** button was only green when the data/form was valid to move on. However in some cases the data was not valid but you could still choose to move on. In this instance the button should not be green so as not to draw the user to do that normally. In this particular case when the user can choose to **Skip** (**Next** button relabeled) uploading data, we don't want the button to be green hence the check you have removed.

However because of the race condition, this functionality has always been a problem. I've had several goes at fixing the race condition with varying degrees of success. The current state is probably the worst it has been where it consistently fails so removing the test is the simplest course of action.

Nateowami: That makes sense.

It appears though that `lexicon-new-project.js` is [pretty explicit](https://github.com/sillsdev/web-languageforge/blob/e5b0aafa0f97eb7630839ad92dbe67838f4a5afc/src/angular-app/languageforge/lexicon/new-project/lexicon-new-project.js#L328) about `newProject.initialData` and `newProject.verifyData` both being always valid. I would think that would be the place to add a check and return `neutral()` in the default state. Not sure why that wasn't apparent to me earlier (Unless of course you've tried it and know that doesn't work for some reason, in which case I trust it's more complicated than it appears and won't attempt it again).

megahirt:
Congratulations on getting all tests passing again!  We will merge this into the offline branch upstream.  I would like the few other cleanup comments taken care of before I merge.

---

Reviewed 1 of 19 files at r1, 1 of 1 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/angular-app/bellows/js/bellows.services.activity.js, line 5 at r4](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110#-Kl1Xws1a3wI8XjhduMA:-Kl1Xws1a3wI8XjhduMB:b1y9l2r) ([raw file](https://github.com/sillsdev/web-languageforge/blob/4fc0a0ee4dfa12744a593d880d3f94c5564387de/src/angular-app/bellows/js/bellows.services.activity.js#L5)):*
> ```JavaScript
> angular.module('bellows.services')
>   .service('activityPageService', ['jsonRpc', 'sessionService', function(jsonRpc, sessionService) {
>     jsonRpc.connect({
> ```

remove dependency on session service and injecting project id (activity list doesn't need projectId

---

*[src/angular-app/bellows/js/bellows.services.user.js, line 7 at r4](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110#-Kl1XfWXDxgTtWGfCvfL:-Kl1XfWXDxgTtWGfCvfM:b-9h9eg3) ([raw file](https://github.com/sillsdev/web-languageforge/blob/4fc0a0ee4dfa12744a593d880d3f94c5564387de/src/angular-app/bellows/js/bellows.services.user.js#L7)):*
> ```JavaScript
>     // Note this doesn't actually 'connect', it simply sets the connection url.
> 
>     var project = ss.session.project;
> ```

let's remove the session dependency here since this user service doesn't depend upon having a projectId (what we've already talked about)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110#-:-Kl1YNrrg55R6YylOpFz:bybacyj)*

Nateowami:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/angular-app/bellows/js/bellows.services.user.js, line 7 at r4](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110#-Kl1XfWXDxgTtWGfCvfL:-Kl1emZjKMxgAt1iRF2T:bkl4s1t) ([raw file](https://github.com/sillsdev/web-languageforge/blob/4fc0a0ee4dfa12744a593d880d3f94c5564387de/src/angular-app/bellows/js/bellows.services.user.js#L7)):*

let's remove the session dependency here since this user service doesn't depend upon having a projectId (what we've already talked about)
</blockquote></details>

For both this user service and the activity service, I initially didn't set the projectId, but that turned out to be problematic. There is only one instance of jsonRpc, so every connect call overwrites any previous connect call (last one wins).  AngularJS services are always singletons, but jsonRpc is a module, so it should be possible to have multiple instances of it, I think (not familiar enough with AngularJS to know whether that's the right way to go or not).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110)*

megahirt:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/angular-app/bellows/js/bellows.services.user.js, line 7 at r4](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110#-Kl1XfWXDxgTtWGfCvfL:-Kl2t6LUtNXkym4-amjv:bklb6d5) ([raw file](https://github.com/sillsdev/web-languageforge/blob/4fc0a0ee4dfa12744a593d880d3f94c5564387de/src/angular-app/bellows/js/bellows.services.user.js#L7)):*

For both this user service and the activity service, I initially didn't set the projectId, but that turned out to be problematic. There is only one instance of jsonRpc, so every connect call overwrites any previous connect call (last one wins).  AngularJS services are always singletons, but jsonRpc is a module, so it should be possible to have multiple instances of it, I think (not familiar enough with AngularJS to know whether that's the right way to go or not).
</blockquote></details>

you're right about the service being a singleton.  The jsonprc.connect() has always been a "fake" connect, and it looks like the module variable is a singleton as well.  I'm inclined to move the connect method into the service method itself.  So that each service method had two lines: jsonRpc.connect() followed by jsonRpc.call().  Another option would be to make a convenience method on jsonRpc, like jsonRpc.callXF() which calls an xforge API method, and internally does the connect()/call() duo for you.  I think I like that better.  We've already modified jsonRpc to include projectId so modifying it further seems OK.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110)*

Nateowami:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/angular-app/bellows/js/bellows.services.user.js, line 7 at r4](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110#-Kl1XfWXDxgTtWGfCvfL:-Kl30gbXw1vHbRTFXCk1:beko1ax) ([raw file](https://github.com/sillsdev/web-languageforge/blob/4fc0a0ee4dfa12744a593d880d3f94c5564387de/src/angular-app/bellows/js/bellows.services.user.js#L7)):*

you're right about the service being a singleton.  The jsonprc.connect() has always been a "fake" connect, and it looks like the module variable is a singleton as well.  I'm inclined to move the connect method into the service method itself.  So that each service method had two lines: jsonRpc.connect() followed by jsonRpc.call().  Another option would be to make a convenience method on jsonRpc, like jsonRpc.callXF() which calls an xforge API method, and internally does the connect()/call() duo for you.  I think I like that better.  We've already modified jsonRpc to include projectId so modifying it further seems OK.
</blockquote></details>

Is there a reason `connect()` and `call()` are separate? `connect()` only sets URL, RPC version, and anything else the caller is interested in. `call()` takes `(remoteFunction, remoteParams, callback, id)`. If everything were passed to jsonRPC in one function call it would be stateless  and we would avoid this issue. And the services can have their own helper functions for calling jsonRpc, providing the projectId only if needed (so they don't have to be so verbose about calling with the URL and projectId).

Actually, I think most of the services could be shortened to something like (untested):
``` js
['read', 'update', 'readProfile' ...].forEach(method => {
  this[method] = function() {
    var args = [].slice.call(arguments) // convert arguments to array
    var callback = args.pop()
    jsonRpc.call('user_'  + method, args, callback)
  }
})
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110)*

megahirt:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/angular-app/bellows/js/bellows.services.user.js, line 7 at r4](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110#-Kl1XfWXDxgTtWGfCvfL:-Kl80fRwIaR0Kc2k0Bjh:b-1knopi) ([raw file](https://github.com/sillsdev/web-languageforge/blob/4fc0a0ee4dfa12744a593d880d3f94c5564387de/src/angular-app/bellows/js/bellows.services.user.js#L7)):*

Is there a reason `connect()` and `call()` are separate? `connect()` only sets URL, RPC version, and anything else the caller is interested in. `call()` takes `(remoteFunction, remoteParams, callback, id)`. If everything were passed to jsonRPC in one function call it would be stateless  and we would avoid this issue. And the services can have their own helper functions for calling jsonRpc, providing the projectId only if needed (so they don't have to be so verbose about calling with the URL and projectId).

Actually, I think most of the services could be shortened to something like (untested):
``` js
['read', 'update', 'readProfile' ...].forEach(method => {
  this[method] = function() {
    var args = [].slice.call(arguments) // convert arguments to array
    var callback = args.pop()
    jsonRpc.call('user_'  + method, args, callback)
  }
})
```
</blockquote></details>

I don't know of a good reason to keep connect() and call() separate.  I like your idea of both condensing the service and making jsonRpc stateless in a single call.  Try it with one service and see if the tests pass.  If you refactor and the tests are still passing, then that's a great way to go.  Thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110)*

Nateowami:
Just fixed some inconsistent E2E tests. Do you want this merged and then PR'd against master?

Nateowami:
Review status: 17 of 20 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/angular-app/bellows/js/bellows.services.user.js, line 7 at r4](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110#-Kl1XfWXDxgTtWGfCvfL:-KlYR-7Q6Dh8dLgR1opL:b-uq7y9c) ([raw file](https://github.com/sillsdev/web-languageforge/blob/4fc0a0ee4dfa12744a593d880d3f94c5564387de/src/angular-app/bellows/js/bellows.services.user.js#L7)):*

I don't know of a good reason to keep connect() and call() separate.  I like your idea of both condensing the service and making jsonRpc stateless in a single call.  Try it with one service and see if the tests pass.  If you refactor and the tests are still passing, then that's a great way to go.  Thanks!
</blockquote></details>

I'm removing the need for a `connect()` call. I've thought of a few downsides to creating services methods so dynamically. For one, it's hard to jump to definitions. It's also going to make TypeScript unhappy if we go that route. I'm trying `this.methodName = api.method('method_name');`. It's more verbose, but easier to read and type check.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110)*

Nateowami:
I've refactored things so each service no longer uses jsonRpc directly. jsonRpc is now stateless, except for its unique id that gets incremented for each request There's now a bellows api service that is consumed by the other services and takes care of setting projectId when applicable, knowing the endpoint to hit, etc.

Nateowami:
The app cache is commented out and the tests are passing when they feel like it (i.e. usually), so we should be good to merge.

megahirt:
I'll take my chances and merge :) :lgtm: 

---

Reviewed 2 of 3 files at r5, 22 of 22 files at r6, 1 of 1 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110#-:-KlbhUokFEoAnWzUHWIc:b-xohb8z)*

Nateowami:
@megahirt Would you like me to resolve the conflicts or are you working on them? It looks to me like for Sf.php, take offline-new, for the e2e tests, take master, and for jsonRpc its a complicated combination.

megahirt:
@Nateowami I tried unsuccessfully to merge and then make tests pass...  I tried for an hour and then had to go home.  If you could please merge master into your branch, make the tests pass, and then I can review it again and merge into master.  Thanks!

Nateowami:
@megahirt I believe I have merged this correctly, but the tests fail because they were broken in 7b618409f577de63e7758e513c1a63ffba6e51e5 (Upgrade to AngularJS 1.6). How would you like me to proceed? I can try to fix them, but given the update to AngularJS 1.6, it may be difficult and take awhile.

megahirt:
@Nateowami can you merge master into this branch (resolving conflicts) so that the PR shows a green merge button?  @irahopkinson do you have any insight into the test failures re: Angular 1.6 ?

Nateowami:
@megahirt Done.

megahirt:
:lgtm: thanks for persevering

---

Reviewed 4 of 4 files at r8, 3 of 3 files at r9.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/110#-:-Km-kndXRgnBBjHylWYA:bke25vz)*

PR #121, opened 2017-05-24
Add Bower install to docs before Sass build

Nateowami:
This was one of the things I ran into during setup. The Sass build has to be able to import Bower dependencies.

There are other issues I ran into that I may add documentation for, but I'm not yet completely clear on yet.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/121)

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,9 @@ Install node_modules used to build Sass files and run E2E tests
 ```
 cd web-languageforge
 npm install
+cd src
+bower install
+cd ..
 gulp sass
 ```
 To watch Sass files for changes, run `gulp sass:watch`. The output will also be in a more readable format (rather than compressed as it is with `gulp sass`). You can also pass the `--debug` flag to enable source comments and source maps in comments in the output files.
@@ -240,7 +243,7 @@ Right-click to select **Options** and set **IDE key**
 PhpStorm PHPSTORM
 ```
 
-When it's time to Debug, check that the bug icon is green for **Debug**.  
+When it's time to Debug, check that the bug icon is green for **Debug**.
 
 Then, from PhpStorm, click the telephone icon near the top right for *Start Listening for PHP Connections*.
 


README.md
```diff
@@ -78,6 +78,9 @@ Install node_modules used to build Sass files and run E2E tests
 ```
 cd web-languageforge
 npm install
+cd src
+bower install
```
irahopkinson: This shouldn't be needed. If you have just done the Ansible task in the instruction prior then it will have done a bower install.

megahirt:
*[README.md, line 82 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/121#-Kkup_w5ckV7Rv5_SAKu-r1-82:-Kl1WaevihAXTDV-JpBL:bywje6p) ([raw file](https://github.com/sillsdev/web-languageforge/blob/51ce4b4f67afe7b1b8c6437c961827236bb3558e/README.md#L82)):*

This shouldn't be needed. If you have just done the Ansible task in the instruction prior then it will have done a bower install.
</blockquote></details>

I'm not so sure about this.  I went in a looked at dependencies.yml which is called from the xenial_playbook.yml and I don't see any reference to `bower install`.  Yes, there is a `npm install -g bower` but no reference to `bower install`.  It makes sense to do the npm install and bower install manually.  And if we did automate the bower install (which I can't find), why did we not also automate the npm install?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/121)*

megahirt:
will wait to see what Ira says.

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/121#-:-Kl1XG1bnvSZ9UlU7KCx:bqporoy)*

Nateowami:
Ira is right, it does a Bower install [in code.yml](https://github.com/sillsdev/web-languageforge/blob/master/deploy/code.yml#L15). As Chris said, it would make sense to do `npm install` as well, unless we want them to both be manual.

The reason I thought it didn't run `bower install` is that when I ran Ansible the end was approximately:
```
localhost                  : ok=<some number>   changed=<some number>   unreachable=0    failed=1
```
The one failed task was symlinking `node` to `nodejs`. However, I already had Node installed, and it was trying to convert a binary into a symlink to a binary, and refusing to do so. Since it said `unreachable=0 failed=1` I understood it to mean this was the only failed task, and all others had succeeded. Then I ran `gulp sass`, which failed because Bower dependencies hadn't been installed. Later I realized that Ansible hadn't actually finished, so I moved the `nodejs` binary out of the way so the task would succeed.

Ideally the Ansbile tasks should play nicely with a pre-existing Node.js install, but what was more concerning was that it seemed to say all other tasks passed, giving a false sense that everything was fine. Perhaps there's something I don't understand about how Ansbile works? Why might it say only one task failed and none were unreachable when (from my perspective) most of the tasks were not attempted due to a failure?

megahirt:
I think the lesson here is that Ansible (both the tool and the scripts) are a work in progress.  It's frustratingly hard to tell from the Ansible output what has actually changed/happened.  Ira just explained to me that the original design of the Ansible scripts was the separate out site dependencies from developer dependencies.  bower dependencies are all front-end site dependencies.  npm dependendencies were developer only dependencies and were not included in the Ansible setup.  But that's no longer the case, npm dependencies are now required for the site (SASS) and likely more stuff should be moved to npm from bower since bower is basically not recommended anymore.  For the short term, let's add the npm install next to bower install in the code.yml.  And make a note about the ansible error with a pre-existing nodejs install.  That can get fixed up later.

megahirt:

---

*[README.md, line 82 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/121#-Kkup_w5ckV7Rv5_SAKu-r1-82:-Klb4NPqQa_Zx20WA3wT:b98wk5n) ([raw file](https://github.com/sillsdev/web-languageforge/blob/51ce4b4f67afe7b1b8c6437c961827236bb3558e/README.md#L82)):*

I'm not so sure about this.  I went in a looked at dependencies.yml which is called from the xenial_playbook.yml and I don't see any reference to `bower install`.  Yes, there is a `npm install -g bower` but no reference to `bower install`.  It makes sense to do the npm install and bower install manually.  And if we did automate the bower install (which I can't find), why did we not also automate the npm install?
</blockquote></details>

I stand corrected.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/121)*

megahirt:
We'll close this one since the bower install isn't actually necessary if you follow the steps, as Ira said.

PR #133, opened 2017-06-28
Remove window.session from the the page template

Nateowami:
Removes window.session from the page template. On the client side it was still being used to determine whether the site is in production or not.

@megahirt We discussed a bit on Flowdock whether to make .org production and all else development, or .local development and all else production. This PR uses .local, but as I started to open this PR it occurred to me that when proxying with host rewriting, the client-side will see a non-.local hostname (for example, I've been accessing Language Forge on my phone as <ip address>:3000). This makes me think it might be better to fall back to development mode when the host ends in neither .org or .local.

Either way it's not a big deal at this time. All the isProduction variable is used for is turning on AngularJS optimizations, though it could be used for more in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/133)

### DIFF
--- a/src/Site/Controller/Base.php
+++ b/src/Site/Controller/Base.php
@@ -100,9 +100,6 @@ class Base
             array_unshift($this->data['cssFiles'], "Site/views/shared/cssBootstrap2/bootstrap.css");
         }
 
-        // Other session data
-        $this->data['jsonSession'] = json_encode(SessionCommands::getSessionData($this->_projectId, $this->_userId, $this->website, $this->_appName), JSON_UNESCAPED_SLASHES);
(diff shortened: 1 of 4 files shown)


megahirt:
.org isn't enough because we still have one .com domain and I can't predict
what the domains might be in the future.  We could do .net, .org and .com
for now and fix it up in the future if that pattern wasn't sufficient.

megahirt:
Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/133)*

PR #143, opened 2017-07-25
New feature: can hide other users' responses

rmunn:
This adds a project setting, "users can see each others' answers", which defaults to ON (true). If turned off, then users will only be able to see their own responses (answers and comments) in the question view and in response counts. Other users' responses will also be hidden in the activity view, since otherwise this could become a "back door" way of seeing what other users have responded.

Also, if a project used to allow users to see (and therefore comment on) each others' answers, and then the setting is turned off, users will no longer be able to see the comments they had previously left on other users' answers.

This feature includes a migration script to add this setting to all projects that don't have it already, setting it to "true" so that existing projects will have their current behavior unchanged.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/143)


irahopkinson:
I haven't reviewed the tests yet but thought I would give you my comments so far. On the whole this is looks good; let me know if you don't agree with any of my comments.

---

Reviewed 10 of 17 files at r1.
Review status: 10 of 17 files reviewed at latest revision, 7 unresolved discussions.

---

*[src/Api/Model/Scriptureforge/Sfchecks/SfchecksProjectModel.php, line 29 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpxfHj0LZm0V62IY9wD:-KpxfHj0LZm0V62IY9wE:b-6rc1j3) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/src/Api/Model/Scriptureforge/Sfchecks/SfchecksProjectModel.php#L29)):*
> ```PHP
> 
>     /** @var boolean Does this project allows users to see each other's answers and comments, or just their own? */
>     public $usersSeeEachOthersResponses;
> ```

If this property was **usersOnlySeeOwnResponses** then I think migration would not be necessary.

---

*[src/Api/Model/Scriptureforge/Sfchecks/Dto/QuestionCommentDto.php, line 74 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpxgTs9zDT481aUkc8K:-KpxgTs9zDT481aUkc8L:bb9i3x8) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/src/Api/Model/Scriptureforge/Sfchecks/Dto/QuestionCommentDto.php#L74)):*
> ```PHP
>             // Can't use array_filter() here because we have MapOf types, so we have to do this by hand
>             $answersToRemove = array();
>             foreach ($question->answers as $akey => $a) {
> ```

rename $a to $answer; the short name is not obvious. rename $akey to $answerKey.

---

*[src/Api/Model/Scriptureforge/Sfchecks/Dto/QuestionCommentDto.php, line 77 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-Kpxghe-zj50LvJc39qD:-Kpxghe-zj50LvJc39qE:b-f3mut) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/src/Api/Model/Scriptureforge/Sfchecks/Dto/QuestionCommentDto.php#L77)):*
> ```PHP
>                 if ($a->userRef->id === $userId) {
>                     $commentsToRemove = array();
>                     foreach ($a->comments as $ckey => $c) {
> ```

rename $c to $comment, rename $ckey to $commentKey. Similarly elsewhere.

---

*[src/Api/Model/Scriptureforge/Sfchecks/Dto/UsxHelper.php, line 61 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpxhyzHquR3dHtP6fw6:-KpxhyzHquR3dHtP6fw7:b-st0wi6) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/src/Api/Model/Scriptureforge/Sfchecks/Dto/UsxHelper.php#L61)):*
> ```PHP
>         // If we can't parse *anything at all*, assume this wasn't actually USX and return it unchanged.
>         if ($this->_out === '')
>             $this->_out = $this->_usx;
> ```

This seems unrelated. Does it belong in a separate commit?

---

*[src/Api/Model/Shared/UserUnreadModel.php, line 93 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-Kpxkb_dy6MHUQEAEfcN:-Kpxkb_dy6MHUQEAEfcO:btflt47) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/src/Api/Model/Shared/UserUnreadModel.php#L93)):*
> ```PHP
>      * @param string[] $itemIds
>      * @param bool $allTypes - whether to remove ALL types of unread items whose IDs are in $itemIds, or just the types this instance was created to handle
>      * @param bool $inputIsAssociativeArray - whether $itemIds is a "dictionary-like" array with item IDs as keys (default is false, meaning it's just a "normal" array with integer keys)
> ```

boolean variables should start with "is" or "has" to make them easy to spot.
Just an order change for $inputIsAssociativeArray to $isInputAssociativeArray.
$allTypes needs some work, maybe $isAllTypesRemoved

---

*[src/Api/Model/Shared/UserUnreadModel.php, line 138 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-Kpxm84qQ2vyz9wLPpFp:-Kpxm84qQ2vyz9wLPpFq:b739dxg) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/src/Api/Model/Shared/UserUnreadModel.php#L138)):*
> ```PHP
> 
>     /**
>      * @return array<string>
> ```

The correct comment here would be **@return string[]**. AFAICT things in <> are ignored.

---

*[src/Api/Model/Shared/Dto/ActivityListDto.php, line 65 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpyAEjX3xKNWfOwUJ_g:-KpyAEjX3xKNWfOwUJ_h:bvznd) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/src/Api/Model/Shared/Dto/ActivityListDto.php#L65)):*
> ```PHP
> 
>     // Helper function for getActivityForUser()
>     public static function filterActivityByUserId($projectModel, $userId, $itemId)
> ```

If your comment is correct (and it appears it is) then this should be a private function (and should therefore be move to below the public functions)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-:-KpyB_sF-cCaIUWUPPb7:b4zwdcr)*

irahopkinson:
Are all the E2E tests still passing?

---

Reviewed 5 of 17 files at r1.
Review status: 15 of 17 files reviewed at latest revision, 16 unresolved discussions.

---

*[scripts/server/migration/addProjectSettingUsersSeeEachOthersResponses.sh, line 16 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpyGCmZgRbalUoGCy3_:-KpyGCmZgRbalUoGCy3a:b-y7lxrp) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/scripts/server/migration/addProjectSettingUsersSeeEachOthersResponses.sh#L16)):*
> ```Shell
> echo After migration, the count of projects with a usersSeeEachOthersResponses setting has become:
> echo
> mongo scriptureforge --eval 'db.projects.find({appName: "sfchecks", usersSeeEachOthersResponses: {$ne: null}}, {projectName: 1, projectCode: 1, usersSeeEachOthersResponses: 1}).count()'
> ```

I like this way of doing migration. I have thought about figuring out how to do this but have not had the time. Well done.

---

*[test/php/model/scriptureforge/Sfchecks/command/QuestionCommandsTest.php, line 212 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpyHvUhrNG_WvXoLwDd:-KpyHvUhrNG_WvXoLwDe:b-ln7fc9) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/test/php/model/scriptureforge/Sfchecks/command/QuestionCommandsTest.php#L212)):*
> ```PHP
>         // setting, then A's vote on B's answer will still be in the DTO, though B's answer is no longer there.
>         $mongoEnviron = new MongoTestEnvironment();
>         $environ = new UserVoteTestEnvironment();
> ```

UserVoteTestEnvironment should inherit from MongoTestEnvironment then you don't need two $environs. I realise the code was already like that.

---

*[test/php/model/scriptureforge/Sfchecks/command/QuestionCommandsTest.php, line 231 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpyJWSBPw8fCsj4XSC5:-KpyJWSBPw8fCsj4XSC6:b-q8ls1s) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/test/php/model/scriptureforge/Sfchecks/command/QuestionCommandsTest.php#L231)):*
> $this->assertNotEmpty($dto['question']['answers']);  // Basically equivalent to the check above
This is not checking what the comments says if the comment below is correct (Here we can't use assertEmpty since a stdClass instance is *not* considered empty in PHP), in which case remove it. If the opposite won't fail its not a valid test.

---

*[test/php/model/scriptureforge/Sfchecks/dto/CommonQuestionsAndAnswersForDto.php, line 100 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpyMM8JWaNXJQ-XRSfv:-KpyMM8JWaNXJQ-XRSfw:bw8qn3m) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/test/php/model/scriptureforge/Sfchecks/dto/CommonQuestionsAndAnswersForDto.php#L100)):*
> ```PHP
>         $comment2Id = QuestionModel::writeComment($project->databaseName(), $question2Id, $answer3Id, $comment2);
> 
>         return array($projectId, $text1Id, $text2Id, $user1Id, $user2Id, $answer1Id, $answer2Id, $answer3Id, $question1Id, $question2Id, $comment0Id, $comment1Id, $comment2Id);
> ```

Use [] notation instead of array(). In all files.

---

*[test/php/model/scriptureforge/Sfchecks/dto/ProjectPageDtoTest.php, line 54 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpyOGtzQ1sqMADn3OPE:-KpyOGtzQ1sqMADn3OPF:b2azta7) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/test/php/model/scriptureforge/Sfchecks/dto/ProjectPageDtoTest.php#L54)):*
> DtoReturnsExpectedData(
This expectation is not informative; tell us the main thing to look for, I can't easily tell what the asserts should do.
This comment applies to the test functions below also.

---

*[test/php/model/scriptureforge/Sfchecks/dto/ProjectPageDtoTest.php, line 103 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpyN9gcHSUSAFboJo-G:-KpyN9gcHSUSAFboJo-H:bblxq5x) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/test/php/model/scriptureforge/Sfchecks/dto/ProjectPageDtoTest.php#L103)):*
> ```PHP
>      * @return array
>      */
>     public function createProjectForTestingResponseCountOnProjectPage(): array
> ```

This should be private (not public)

---

*[test/php/model/scriptureforge/Sfchecks/dto/QuestionCommentDtoTest.php, line 34 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpyPbwVGLRx_onuy4j-:-KpyPbwVGLRx_onuy4j0:bwqgmek) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/test/php/model/scriptureforge/Sfchecks/dto/QuestionCommentDtoTest.php#L34)):*
> ```PHP
>     }
> 
>     public function assertKeyNotPresent($key, $array)
> ```

This function should also be private and therefore declared below the public ones

---

*[test/php/model/scriptureforge/Sfchecks/dto/QuestionCommentDtoTest.php, line 154 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpyPvq_XRfkA8uopPRT:-KpyPvq_XRfkA8uopPRU:b-lskc95) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/test/php/model/scriptureforge/Sfchecks/dto/QuestionCommentDtoTest.php#L154)):*
> ```PHP
>         // John Carter does not see Dejah Thoris's answer, nor any of the comments below it (including his own)
>         $this->assertKeyNotPresent($answer3Id, $dto['question']['answers']);
> //        $this->assertEquals('On the planet we call Barsoom, which you inhabitants of Earth normally call Mars.', $dto['question']['answers'][$answer3Id]['content']);
> ```

Why are there so many commented lines in this file?

---

*[test/php/model/scriptureforge/Sfchecks/dto/QuestionCommentDtoTest.php, line 202 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpyQ4nFcx3vISO3WUmD:-KpyQ4nFcx3vISO3WUmE:b-7jitqw) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/test/php/model/scriptureforge/Sfchecks/dto/QuestionCommentDtoTest.php#L202)):*
> ```PHP
>     }
> 
>     public function createProjectForTestingAnswerVisibility(): array
> ```

This function should be private

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-:-KpyHNUtSzivcvlMl-Xe:b-sp6fo)*

megahirt:
Reviewed 10 of 17 files at r1.
Review status: 15 of 17 files reviewed at latest revision, 16 unresolved discussions.

---

*[src/Api/Model/Scriptureforge/Sfchecks/Dto/QuestionCommentDto.php, line 74 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpxgTs9zDT481aUkc8K:-Kq-5BmRO-jzJXYVrcA3:b-czakek) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/src/Api/Model/Scriptureforge/Sfchecks/Dto/QuestionCommentDto.php#L74)):*

rename $a to $answer; the short name is not obvious. rename $akey to $answerKey.
</blockquote></details>

Is this a coding style thing?  I admit I'm not very aware of the PSR-1/PSR-2 but I personally don't have a problem with $a.  I'm certainly guilty of short variable names so I'd like to know if this is a code guideline thing.  I'd like to stick to the standard coding practices.  It's obvious to me personally that $a represents $answer because of the $question->answers structure over which we're iterating.  If you didn't have that, then absolutely I would require a more descriptive variable name.

---

*[src/Api/Model/Scriptureforge/Sfchecks/Dto/UsxHelper.php, line 61 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpxhyzHquR3dHtP6fw6:-Kq-6GvCHyBE3eDw3oim:b1pnr12) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/src/Api/Model/Scriptureforge/Sfchecks/Dto/UsxHelper.php#L61)):*

This seems unrelated. Does it belong in a separate commit?
</blockquote></details>

looks like a bugfix.

---

*[src/Api/Model/Shared/UserUnreadModel.php, line 138 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-Kpxm84qQ2vyz9wLPpFp:-Kq-8IqEG6XFQULI5_VP:b-c6n98d) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/src/Api/Model/Shared/UserUnreadModel.php#L138)):*

The correct comment here would be **@return string[]**. AFAICT things in <> are ignored.
</blockquote></details>

I looked around on the net but couldn't find a reference to string[] being favorable to array<string> @irahopkinson do you have a reference?  The phpdoc documention is doesn't specify https://phpdoc.org/docs/latest/references/phpdoc/tags/var.html

---

*[src/Api/Model/Shared/Dto/ActivityListDto.php, line 65 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpyAEjX3xKNWfOwUJ_g:-Kq-8t2dH07SbM_G2hbZ:bberozo) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/src/Api/Model/Shared/Dto/ActivityListDto.php#L65)):*

If your comment is correct (and it appears it is) then this should be a private function (and should therefore be move to below the public functions)
</blockquote></details>

`private static function` sounds good to me

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143)*

megahirt:
@rmunn thanks for this good work and I'm looking forward to seeing it in action.  @irahopkinson thanks for your thorough review!

---

Reviewed 7 of 17 files at r1.
Review status: all files reviewed at latest revision, 17 unresolved discussions.

---

*[test/php/model/scriptureforge/Sfchecks/dto/QuestionCommentDtoTest.php, line 37 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-Kq-YK7dF4QLRS4kGTmp:-Kq-YK7dF4QLRS4kGTmq:b-rc75bw) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/test/php/model/scriptureforge/Sfchecks/dto/QuestionCommentDtoTest.php#L37)):*
> ```PHP
>     {
>         if ($array instanceof stdClass) {
>             print "Skipping assert since stdClass, by definition, has no keys";
> ```

a print statement?  Or maybe you meant for this to be a `throw new \Exception("expected array, given StdClass")`.  I believe you can use PHP types to enforce that the parameter is an array.  So something like `public function assertKeyNotPresent($key, array $array)` should work and throw an TypeError exception if it's not an array

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-:-Kq-aLdij7fZF1601TRU:byyt7q)*

irahopkinson:
Reviewed 1 of 17 files at r1.
Review status: all files reviewed at latest revision, 19 unresolved discussions.

---

*[src/Api/Model/Scriptureforge/Sfchecks/Dto/QuestionCommentDto.php, line 69 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-Kq0jAH5yjf9FkzxAzOk:-Kq0jAH5yjf9FkzxAzOl:bezzdei) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/src/Api/Model/Scriptureforge/Sfchecks/Dto/QuestionCommentDto.php#L69)):*
> array()
use [] everywhere

---

*[src/Api/Model/Scriptureforge/Sfchecks/Dto/QuestionCommentDto.php, line 74 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpxgTs9zDT481aUkc8K:-Kq0fIAsl8j64yPVrn8z:b-u438vc) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/src/Api/Model/Scriptureforge/Sfchecks/Dto/QuestionCommentDto.php#L74)):*

Is this a coding style thing?  I admit I'm not very aware of the PSR-1/PSR-2 but I personally don't have a problem with $a.  I'm certainly guilty of short variable names so I'd like to know if this is a code guideline thing.  I'd like to stick to the standard coding practices.  It's obvious to me personally that $a represents $answer because of the $question->answers structure over which we're iterating.  If you didn't have that, then absolutely I would require a more descriptive variable name.
</blockquote></details>

No its not in PSR-2 (or PSR-12). This is my recommendation based on what I have learnt and most recently from Cambell. Since reading code is harder than writing, anything that makes it easier to read is paramount. When you are writing the code $a & $c makes sense; your head is completely in what you are doing. When you are reading the code (code review or even when the author comes back to make changes) although, as you say, it is **discover-able** (particularly in this case) what $a & $c are, non the less, it is not **obvious** as you have to find the *foreach* to discover what $a & $c are. The section of the code is short at the moment. As it grows it becomes harder to read and we are less likely to rename variables as we go to make it more readable. Its a lot easier to make it readable from the beginning.

If you stick with it the way it is, at least use camel case for $akey and $ckey.

---

*[src/Api/Model/Shared/UserUnreadModel.php, line 138 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-Kpxm84qQ2vyz9wLPpFp:-Kq0p3nocaesBADXQFch:bttd5i) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/src/Api/Model/Shared/UserUnreadModel.php#L138)):*

I looked around on the net but couldn't find a reference to string[] being favorable to array<string> @irahopkinson do you have a reference?  The phpdoc documention is doesn't specify https://phpdoc.org/docs/latest/references/phpdoc/tags/var.html
</blockquote></details>

Check the example here https://phpdoc.org/docs/latest/references/phpdoc/tags/param.html
**array** is the same as **mixed[]** and is an array with mixed content. **string[]** specifies it must be an array of strings.

---

*[test/php/model/shared/commands/ActivityCommandsTest.php, line 84 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-Kq0bwBdEpCboHVBP5oY:-Kq0bwBdEpCboHVBP5oZ:b-tvmjwv) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/test/php/model/shared/commands/ActivityCommandsTest.php#L84)):*
> ```PHP
>     }
> 
>     public function createActivityTestEnvironment($canUsersSeeEachOthersResponses = true)
> ```

This function and all below should be private

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143)*

irahopkinson:
Reviewed 1 of 17 files at r1.
Review status: all files reviewed at latest revision, 19 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143)*

rmunn:
Review status: all files reviewed at latest revision, 19 unresolved discussions.

---

*[scripts/server/migration/addProjectSettingUsersSeeEachOthersResponses.sh, line 16 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpyGCmZgRbalUoGCy3_:-Kq0pH_van_H_2W41ff8:b-uqq64p) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/scripts/server/migration/addProjectSettingUsersSeeEachOthersResponses.sh#L16)):*

I like this way of doing migration. I have thought about figuring out how to do this but have not had the time. Well done.
</blockquote></details>

If there's complex logic then it's probably better to write a PHP migration script. But when it's a single Mongo update to all projects, I figure simplest is best.

---

*[src/Api/Model/Scriptureforge/Sfchecks/Dto/QuestionCommentDto.php, line 74 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpxgTs9zDT481aUkc8K:-Kq0pTW34MOML-XT6lzj:bqqzfxa) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/src/Api/Model/Scriptureforge/Sfchecks/Dto/QuestionCommentDto.php#L74)):*
My philosophy for variable names is basically the same as the one expressed by this quote from *Clean Code*:

> The length of a name should be related to the length of the scope. You can use very short variable names for tiny scopes, but for big scopes you should use longer names. Variable names like i and j are just fine if their scope is five lines long.

I especially prefer this in generic functions, where best you can do to pick a descriptive name for a variable might be `item`. If you have more than one such variable, then in my opinion, `a` and `b` would be *better* names than `item1` and `item2`: the human mind sees more distinction between letters than between numbers, so `a` and `b` are more distinct than `item1` and `item2` -- and `itemA` has no readability advantage over a simple `a`.

However, this function is *not* generic. Here, I agree that `$a` should become `$answer`, and so on. Though if the loop was shorter, I would still argue for `$a`, because a one-letter variable name communicates "loop variable" to me. If I saw `$answer`, I might think it was the function parameter, but if I see `$a`, I would be certain that it was a loop variable. But since the loop is long and nested, the disadvantages of names like `$a` begin to outgrow the advantages. So I agree in this case that the short variable names should be replaced.

---

*[src/Api/Model/Scriptureforge/Sfchecks/Dto/UsxHelper.php, line 61 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpxhyzHquR3dHtP6fw6:-Kq0x4Jf5S_2ORVmMa3P:b-aj1jnd) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/src/Api/Model/Scriptureforge/Sfchecks/Dto/UsxHelper.php#L61)):*

looks like a bugfix.
</blockquote></details>

This bug was interfering with my unit testing (which used a passage from the public-domain book *A Princess of Mars* for the test data), because that book passage doesn't contain USX. So that's why it ended up in this PR, because it was relevant to my unit testing.

I can split it out into a separate PR if that's important.

---

*[src/Api/Model/Shared/UserUnreadModel.php, line 93 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-Kpxkb_dy6MHUQEAEfcN:-Kq0xeX23Sk0NeolIm5a:b-9uf99w) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/src/Api/Model/Shared/UserUnreadModel.php#L93)):*

boolean variables should start with "is" or "has" to make them easy to spot.
Just an order change for $inputIsAssociativeArray to $isInputAssociativeArray.
$allTypes needs some work, maybe $isAllTypesRemoved
</blockquote></details>

I tend to feel that variable names should read like correct English, so I don't like `$isAllTypesRemoved`. (The name `$isInputAssociativeArray` reads like a question, so it's good). Maybe `$shouldRemoveAllTypes`.

---

*[src/Api/Model/Shared/UserUnreadModel.php, line 138 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-Kpxm84qQ2vyz9wLPpFp:-Kq0yH-MashdpAAoZHUx:bficicw) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/src/Api/Model/Shared/UserUnreadModel.php#L138)):*

Check the example here https://phpdoc.org/docs/latest/references/phpdoc/tags/param.html
**array** is the same as **mixed[]** and is an array with mixed content. **string[]** specifies it must be an array of strings.
</blockquote></details>

The previous code had the type signature `array<UnreadItem>` (which was actually incorrect), so I thought that's what it should be. I'll change it to PHP syntax rather than the C#-like syntax that was there.

---

*[test/php/model/scriptureforge/Sfchecks/command/QuestionCommandsTest.php, line 231 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpyJWSBPw8fCsj4XSC5:-Kq0zFmr4GgCBqWqk2Ap:b-sfzqgb) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/test/php/model/scriptureforge/Sfchecks/command/QuestionCommandsTest.php#L231)):*

> $this->assertNotEmpty($dto['question']['answers']);  // Basically equivalent to the check above
This is not checking what the comments says if the comment below is correct (Here we can't use assertEmpty since a stdClass instance is *not* considered empty in PHP), in which case remove it. If the opposite won't fail its not a valid test.
</blockquote></details>

It's equivalent because of how our MapOf encoding works: an empty array is *replaced* by a `stdClass` instance. Those two asserts are not equivalent in PHP, but they *are* equivalent in our code because of that feature of MapOf. However, both assertions are necessary. If I just asserted that the `answers` part of the DTO was not a `stdClass`, I could miss a failure where it ended up being empty and yet was not replaced by a `stdClass` instance the way it should have been. But if I just asserted that it was non-empty, then the test would *incorrectly pass* if a `stdClass` instance showed up, because PHP (or PHPUnit) does not consider `stdClass` to be empty.

So what I should do here is rephrase that comment, because while these two assertions are basically testing the *same thing*, they are not "equivalent" because both of them have a failure mode that the other assertion covers.

---

*[test/php/model/scriptureforge/Sfchecks/dto/CommonQuestionsAndAnswersForDto.php, line 100 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpyMM8JWaNXJQ-XRSfv:-Kq1-YMRL8Xiz2N1GmaH:b-vm3nis) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/test/php/model/scriptureforge/Sfchecks/dto/CommonQuestionsAndAnswersForDto.php#L100)):*

Use [] notation instead of array(). In all files.
</blockquote></details>

Will do. (Leftover habit from PHP 5, and copying the existing style -- but if we're going to change the existing code over gradually to `[]` syntax instead of `array()`, then copying the existing style isn't the right thing to do anymore).

---

*[test/php/model/scriptureforge/Sfchecks/dto/ProjectPageDtoTest.php, line 54 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpyOGtzQ1sqMADn3OPE:-Kq1-lFSaiBxFEYOeXkx:b-m2bhn4) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/test/php/model/scriptureforge/Sfchecks/dto/ProjectPageDtoTest.php#L54)):*

> DtoReturnsExpectedData(
This expectation is not informative; tell us the main thing to look for, I can't easily tell what the asserts should do.
This comment applies to the test functions below also.
</blockquote></details>

I was having the hardest time coming up with a good name for this, and I still can't think of a good one. Suggestions welcome. :-)

---

*[test/php/model/scriptureforge/Sfchecks/dto/QuestionCommentDtoTest.php, line 37 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-Kq-YK7dF4QLRS4kGTmp:-Kq0oNXh4pp4XodK8ldk:b16uksr) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/test/php/model/scriptureforge/Sfchecks/dto/QuestionCommentDtoTest.php#L37)):*

a print statement?  Or maybe you meant for this to be a `throw new \Exception("expected array, given StdClass")`.  I believe you can use PHP types to enforce that the parameter is an array.  So something like `public function assertKeyNotPresent($key, array $array)` should work and throw an TypeError exception if it's not an array
</blockquote></details>

I meant for this to be removed before merging the code into production (I found it useful to see the output in the unit tests), but missed this line when I was cleaning the code up before rebasing onto master. I'll turn this into a comment. I won't make it throw an exception, because the whole point of this was that I did *not* want exceptions to be thrown when the "array" I was getting turned out to be a `stdClass` instance instead. (Our `MapOf` class turns empty arrays into `stdClass` instances so that they will be JSON-serialized as an empty JSON object instead of an empty JSON array, but that was causing me a lot of difficulty in the unit tests (since `assertArrayNotHasKey` was throwing exceptions on those `stdClass` instances) until I figured out how to get around that problem).

---

*[test/php/model/scriptureforge/Sfchecks/dto/QuestionCommentDtoTest.php, line 154 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpyPvq_XRfkA8uopPRT:-Kq1-xZFZMO9sQ45U1Ff:b-tb5owz) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/test/php/model/scriptureforge/Sfchecks/dto/QuestionCommentDtoTest.php#L154)):*

Why are there so many commented lines in this file?
</blockquote></details>

I think I wanted the parallels between this comment and the "CanViewEachOthersAnswers" function to be obvious at a glance. So the data that *would* have been in there if the users could see each others' answers, I commented out but didn't delete. I find the test easier to read that way, but that might just be me. If that gets in the way of reading the test for the rest of the team, though, I'll remove it; let me know.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143)*

rmunn:
> ...  let me know if you don't agree with any of my comments.

The only thing I haven't mentioned yet (since it came up multiple times and I figured I should address it once) is whether helper functions should be private. If the helper functions are a vital part of the way the class works and rely on the instance being in a certain state before they're called (e.g., a function that rebalances an AVL tree), then they should be private, because client code should never be calling those functions. But if the function is "standalone" and doesn't require external state, then even though no *current* client code has a need to call that function, that could change. And there could be client code written in the future that would find that helper function useful, and thus not need to reinvent the wheel. Therefore, if a helper function would not be *harmful* to call from outside its class, I believe it should be public.

In other words, most OO designers seem to feel that you should default to private functions, and only write public functions when you have a specific need for them to be public (e.g., they're part of the API you're designing). I feel the opposite: I feel you should write as many potentially-helpful *public* functions as possible, and only make functions private if there's a specific reason for them to be private (e.g., they would be dangerous to call from outside).

As a specific case of this general rule, consider static methods. They cannot depend on any instance variables and thus cannot depend on the instance being in any given state -- they are essentially "stateless" functions. Therefore, IMHO there is almost never a good reason for writing the words `private static` in your code. (The exception might be a function like `private static ResetTheWorld()`, which is dangerous and should only be called from your initializers. But in general, `private static` are two words that IMHO should never go together).

---

Review status: all files reviewed at latest revision, 19 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-:-Kq10VXtXzyTWqMXPVhV:b-hjwqn9)*

rmunn:
P.S. I'll go ahead and make the functions private that you requested. This comment is more to discuss philosophy of code design than to object to those requests.

---

Review status: all files reviewed at latest revision, 19 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-:-Kq13wNuYn7GuuBcDaZr:b-jiv88i)*

rmunn:
Review status: all files reviewed at latest revision, 19 unresolved discussions.

---

*[src/Api/Model/Scriptureforge/Sfchecks/SfchecksProjectModel.php, line 29 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpxfHj0LZm0V62IY9wD:-Kq14CXgUQ-fuvt1F44_:b-1lyi2z) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/src/Api/Model/Scriptureforge/Sfchecks/SfchecksProjectModel.php#L29)):*

If this property was **usersOnlySeeOwnResponses** then I think migration would not be necessary.
</blockquote></details>

Chris and I discussed what the name of the feature should be, and decided that "users can see each others' responses" was a better name for the setting -- and that it would be less confusing overall to have the setting name in Mongo be the same wording as what the project managers see in the Project Settings view.

So for this one review, I'll say "[stet](https://en.wikipedia.org/wiki/Stet)".

---

*[test/php/model/scriptureforge/Sfchecks/command/QuestionCommandsTest.php, line 212 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpyHvUhrNG_WvXoLwDd:-Kq14pPldADIrNthmF2-:b-vxv09w) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/test/php/model/scriptureforge/Sfchecks/command/QuestionCommandsTest.php#L212)):*

UserVoteTestEnvironment should inherit from MongoTestEnvironment then you don't need two $environs. I realise the code was already like that.
</blockquote></details>

I think I'll defer this one change to a future PR, so that I don't get bogged down in OO design. (Does UserVoteTestEnvironment get used elsewhere? What are the implications of changing its inheritance structure? These questions might take me a little while to answer). It's a good suggestion, but I think I'd like to call it out-of-scope for this particular PR and do it later.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143)*

irahopkinson:
I can understand where you are coming from with the private public thing but in this case I think all of them were in unit tests so no chance of them being used outside anyway and making them private makes that clear and further distinguishes them from the test methods.

---

Review status: all files reviewed at latest revision, 19 unresolved discussions.

---

*[src/Api/Model/Scriptureforge/Sfchecks/Dto/UsxHelper.php, line 61 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpxhyzHquR3dHtP6fw6:-Kq1nKa01EmxuOcmGdA4:b-v013va) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/src/Api/Model/Scriptureforge/Sfchecks/Dto/UsxHelper.php#L61)):*

This bug was interfering with my unit testing (which used a passage from the public-domain book *A Princess of Mars* for the test data), because that book passage doesn't contain USX. So that's why it ended up in this PR, because it was relevant to my unit testing.

I can split it out into a separate PR if that's important.
</blockquote></details>

Ordinarily you would commit this fix first stating the bug fix in the commit message. But its 2 lines so leave it as is.

---

*[test/php/model/scriptureforge/Sfchecks/command/QuestionCommandsTest.php, line 212 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpyHvUhrNG_WvXoLwDd:-Kq1mv9cAYltO7R5CbBF:b7n2zrp) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/test/php/model/scriptureforge/Sfchecks/command/QuestionCommandsTest.php#L212)):*

I think I'll defer this one change to a future PR, so that I don't get bogged down in OO design. (Does UserVoteTestEnvironment get used elsewhere? What are the implications of changing its inheritance structure? These questions might take me a little while to answer). It's a good suggestion, but I think I'd like to call it out-of-scope for this particular PR and do it later.
</blockquote></details>

Yep, fair enough.

---

*[test/php/model/scriptureforge/Sfchecks/dto/QuestionCommentDtoTest.php, line 154 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpyPvq_XRfkA8uopPRT:-Kq1m5qRba9eaVqLZuO6:beofmug) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/test/php/model/scriptureforge/Sfchecks/dto/QuestionCommentDtoTest.php#L154)):*

I think I wanted the parallels between this comment and the "CanViewEachOthersAnswers" function to be obvious at a glance. So the data that *would* have been in there if the users could see each others' answers, I commented out but didn't delete. I find the test easier to read that way, but that might just be me. If that gets in the way of reading the test for the rest of the team, though, I'll remove it; let me know.
</blockquote></details>

If you want to leave them in put a comment at the top of the function stating what the purpose of the commented asserts are. Something along the lines of what you described here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-:-Kq1oQV3f04vTDqCVvjv:b-rw7j51)*

megahirt:
Reviewed 5 of 9 files at r2.
Review status: 13 of 17 files reviewed at latest revision, 19 unresolved discussions.

---

*[test/php/model/scriptureforge/Sfchecks/dto/ProjectPageDtoTest.php, line 54 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpyOGtzQ1sqMADn3OPE:-Kq38dnhW53u3sKuD288:bgotby) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/test/php/model/scriptureforge/Sfchecks/dto/ProjectPageDtoTest.php#L54)):*

I was having the hardest time coming up with a good name for this, and I still can't think of a good one. Suggestions welcome. :-)
</blockquote></details>

_DtoHasOthersAnswers()

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143)*

megahirt:
I would like to see at least one e2e test included in this PR.  The PHP tests looks good AFAICT.  Also, please confirm that all e2e tests (SF/LF) are passing before merging.

---

Reviewed 4 of 9 files at r2.
Review status: all files reviewed at latest revision, 18 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-:-Kq39UgCLPItVUEg0yvB:b1zzdxq)*

rmunn:
Rebased onto master after Ira's big changes. Tests closer to passing now, but four failures on E2E tests because they are expecting a condition that's not true when the "users can see each other's answers" setting is turned off. I'm still working out the best way to handle those failures, so this is not yet ready to merge.

rmunn:
Ready to merge now, or to review again if you feel that the E2E tests need reviewing.

megahirt:
Provided that all reviewable comments have been resolved, this is ready to merge.  Thanks!

---

Reviewed 4 of 7 files at r3, 1 of 1 files at r4.
Review status: 22 of 24 files reviewed at latest revision, 19 unresolved discussions.

---

*[test/app/scriptureforge/activity/e2e/activity.spec.js, line 176 at r4](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-Kri7lw0Z2DXMTbva2Ni:-Kri7lw0Z2DXMTbva2Nj:b7hiqq9) ([raw file](https://github.com/sillsdev/web-languageforge/blob/215422205b94aba05f5f05df096464532f146583/test/app/scriptureforge/activity/e2e/activity.spec.js#L176)):*
> ```JavaScript
>   };
> 
>   function verifyResponseVisibility(valueShouldBeTrue) {
> ```

I'm OK with keeping this for now but verifyResponseVisibility is overkill because it just tests that the setting was correctly persisted.  It will add a few seconds to our test run.

Also, setResponseVisibility and verifyResponseVisibility are always called together with the same value it seems, so they could probably just be one function.  Again, I'm thinking verifyResponseVisibility is overkill.

---

*[test/php/model/scriptureforge/Sfchecks/dto/QuestionCommentDtoTest.php, line 34 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpyPbwVGLRx_onuy4j-:-Kri6Y8BAcpn4RLNx6SO:bi6n0b5) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/test/php/model/scriptureforge/Sfchecks/dto/QuestionCommentDtoTest.php#L34)):*

This function should also be private and therefore declared below the public ones
</blockquote></details>

agreed.

---

*[test/php/model/scriptureforge/Sfchecks/dto/QuestionCommentDtoTest.php, line 202 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-KpyQ4nFcx3vISO3WUmD:-Kri6oq3Kfe2qj8Anxve:b-a1v22u) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b16d8257181fb9b90b925f66c30de8cd4607f8b2/test/php/model/scriptureforge/Sfchecks/dto/QuestionCommentDtoTest.php#L202)):*

This function should be private
</blockquote></details>

I don't know what function this comment is referring to anymore.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143#-:-Kri8crss43W2-PRY3El:b-ydk5ws)*

megahirt:
Reviewed 2 of 7 files at r3.
Review status: all files reviewed at latest revision, 19 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/143)*

rmunn:
Finally getting around to merging this. The Reviewable discussions that are "left open" have all been resolved, but the Reviewable system would require Ira or Chris to go through and manually check the "satisfied" status on each one. Rather than making them do that, since we've discussed it verbally and they're satisfied, I'm just going to merge this in via the Github "Squash and merge" button.

Issue #160, opened 2017-09-05
Fields showing up in SBF project for contributor that shouldn't


rmlockwood:
Two custom fields at the entry level (Use Standard Spelling & VParadigm3SgPast), a custom field at the sense level (SimilarPhrase) and the POS field are all turned off for being visible to a contributor yet they are still showing.

rmunn:
I've found the cause of this bug — a mistake in how we filtered out fields that were turned off in the project configuration — and written a bugfix for it. We'll test the fix, then we'll close this issue once the bugfix has been deployed to the Language Forge site.

megahirt:
Thanks again for the bug report. This fix has now been shipped to languageforge.org In some cases I recommend doing a Ctrl-F5 in case the changes are not immediately reflected in your browser.

PR #161, opened 2017-09-07
Fix fields not being properly hidden

rmunn:
The root cause of the bug was the function that removed fields from the config's list of visible fields if they had been hidden for a particular user or role. It was removing fields from the fieldOrder list *while* it was still iterating over that list, which is always fraught with danger. The solution is to filter the list first, then set the config.

This is a fix for #160.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/161)


megahirt:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/161#-:-KtQkme055wIfNeugHMO:bnfp4nl)*

Issue #29, opened 2015-11-26
In Option List config, can delete wrong option


rmunn:
I created a new project, and added one word, and set its Part of Speech to `Verb (v)`. I then went to the Dictionary Configuration settings, on the Option Lists tab. I edited the Part of Speech list by adding a new category called `Danger ahead (danger)` -- the precise name doesn't matter as long as it's a new category. The newly-created category, at the bottom of the list, had a delete button next to it. BEFORE clicking Apply, I dragged that category up to the middle of the list. BUT the delete button stayed where it was, at the bottom of the list, and I was able to click on it to delete the `Verb (v)` category! The word I had entered, whose part of speech had been `Verb (v)`, still showed as `Verb (v)` at first, and seemed to stay that way -- but at some point when I refreshed the page (without having ever edited any part of the word), its part of speech had switched to `Unknown`. (The Mongo database record, though, still contained `"partOfSpeech" : { "value" : "v" }`).

Here's a series of screenshots showing the steps to reproduce this bug. The mouse cursor isn't visible, but it should be obvious what I clicked on: the plus sign to add the new part of speech, then the drag bar to move the new entry, then the delete button to delete the _wrong_ option.

What should have happened, of course, is that the delete button should have moved along with the newly-created option that it was "attached to".

![bug-step-01](https://cloud.githubusercontent.com/assets/90762/11418940/e3f9a898-9456-11e5-94bf-dd4e2e163465.png)
![bug-step-02](https://cloud.githubusercontent.com/assets/90762/11418941/e3ffe6ea-9456-11e5-9729-525df9540242.png)
![bug-step-03](https://cloud.githubusercontent.com/assets/90762/11418942/e40006d4-9456-11e5-9009-c4cffb232f5e.png)
![bug-step-04](https://cloud.githubusercontent.com/assets/90762/11418943/e40018b8-9456-11e5-8366-54185b722209.png)
![bug-step-05](https://cloud.githubusercontent.com/assets/90762/11418944/e4001e94-9456-11e5-952b-09057c2fbca6.png)

ThomasBlok:
Issue moved to https://jira.sil.org/browse/LF-38

PR #163, opened 2017-09-08
Fix wrong items being deleteable from option lists

rmunn:
We now track the index of every item in the list, so that the delete buttons appear to "travel" alongside their corresponding items if items are rearranged.

This is a bugfix for #29.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/163)

### DIFF
--- a/src/angular-app/bellows/directive/bootstrap4/picklist-editor.html
+++ b/src/angular-app/bellows/directive/bootstrap4/picklist-editor.html
@@ -16,7 +16,7 @@
         </div>
     </div>
     <div class="row clearfix">
-        <div sv-root sv-part="items">
+        <div sv-root sv-part="items" sv-on-sort="onSort($indexFrom, $indexTo)">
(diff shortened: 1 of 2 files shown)


megahirt:
This seems somewhat complicated for keeping track of items, but it appears this is necessary given the constraints of the sv feature.  :lgtm: 

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/163#-:-Ktj46Ufbk0lzfvW6g0r:b-molskt)*

rmunn:
Yes, it's a bit complicated. Once we move to using GUIDs (which, once assigned, will never change) for tracking option lists in LF, that will allow me to do this much more simply by keeping a list of GUIDs that are deletable. Then we won't have to track when an index changes, or update the list of existing indices when another one changes (if GUID abc goes from index 4 to 5, the code that looks at the GUID won't need to care about that).

PR #179, opened 2017-12-25
Installer: Fix usage for clean install

alpha0010:
The folder `web-scriptureforge` was created via root.  `node_modules` does not exist on a clean install.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/179)


megahirt:
OK thanks.  This is good.

PR #181, opened 2018-01-28
Comments view integrated into the main entry view

nigel-wells:
Hi Team

This request includes all the code for the new comments view being integrated into the entry view. I tried to my naming conventions for variables and classes to match what was already there but it did seem to change as I went along - nature of multiple developers I guess.

There is a new npm package installed: `angular-inview`. This allows for setting triggers of when a specific element is available in the current viewport (or an offset of the viewport). The original proposal from us was to hide the comment bubbles when not applicable to the current area a user was viewing. However, in practice, I've found it not overly helpful as it makes you wonder some bubbles aren't there and it forces you to have to scroll to see if there is a comment available at the top/bottom of the screen. I could decrease the offset but then all the hide/show logic would appear out of view which then makes it pointless to even have it there in the first place. The initial concern was there would be comment bubbles everywhere. I do feel that with the colour they are initially given they are not overwhelming in anyway so we could probably just leave them in place. This pull requests INCLUDES the logic for the show/hide trigger so you will need to make sure that package is installed and webpack re-built to get them in the dist directory.

You'll see two commits there from apex-nigel - I had my IDE set to this when I pushed them through. Let me know if its easy enough for you to change them at your end or if that's something I need to do.

## Testing

I've not been able to test on iOS devices as I had trouble getting them to connect to localhost on my machine. 

I've tested in the latest versions of:

- Firefox
- Chrome
- Microsoft Edge
- Android Chrome
- Android Samsung Internet

I had tried on the old Android kit-kat browser as Campell mentioned but the app itself didn't work at all let alone being able to test what I've done.

E2E tests have been updated to work with the updated entry view. I've kept the same ones there while also adding in a couple of new ones.

## Issues

One of the main issues, that is partially unresolved, is being able to tell which comment specifically relates to which part of the entry. The old comment view just listed all comments in a column and kept a copy of the original text it related to. With the new system they need to remain in context. Currently there is nothing stored in the DB that relates to a specific part of the entry screen so, when a field name is replicated (such as a meaning or example) it becomes impossible to tell which comment relates to which field. Although a check could be done on the value of a field this isn't ideal as, if the value is changed (i.e. it was spelt wrong) then the comment will disappear as its context has changed. For now I've kept the context based on the field name and the input system. For pictures, semantic domains, and multi-choice options, I was able to use the value as the interface doesn't allow it to change - only delete. Chris said he was going to provide a `senseId` to help provide better context.

## Screenshots

Updated entry view showing comment bubbles with comments and without:
image

Viewing a comment:
image

Smaller viewports (tablets) will hide the words in the dictionary:
image

Mobile viewports will show the comments in an overlaid panel:
image

Let me know if you need clarification on anything else.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/181)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -339,6 +339,12 @@
       "resolved": "https://registry.npmjs.org/angular/-/angular-1.6.4.tgz",
       "integrity": "sha1-A7exXAGggC1+LPWTJA5gQFTcd/s="
     },
+    "angular-inview": {
+      "version": "2.1.0",
+      "resolved": "https://registry.npmjs.org/angular-inview/-/angular-inview-2.1.0.tgz",
+      "integrity": "sha1-noYVuJ8m+AVlqu9wagSpxmfH14U=",
+      "dev": true
+    },
     "angular-mocks": {
       "version": "1.6.6",
       "resolved": "https://registry.npmjs.org/angular-mocks/-/angular-mocks-1.6.6.tgz",
--- a/package-lock.json
+++ b/package-lock.json
@@ -339,12 +339,6 @@
       "resolved": "https://registry.npmjs.org/angular/-/angular-1.6.4.tgz",
       "integrity": "sha1-A7exXAGggC1+LPWTJA5gQFTcd/s="
     },
-    "angular-inview": {
-      "version": "2.1.0",
-      "resolved": "https://registry.npmjs.org/angular-inview/-/angular-inview-2.1.0.tgz",
-      "integrity": "sha1-noYVuJ8m+AVlqu9wagSpxmfH14U=",
-      "dev": true
-    },
     "angular-mocks": {
       "version": "1.6.6",
       "resolved": "https://registry.npmjs.org/angular-mocks/-/angular-mocks-1.6.6.tgz",
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "@types/offline-js": "^0.7.28",
     "@types/quill": "1.3.2",
     "@types/rangy": "0.0.30",
+    "angular-inview": "^2.1.0",
     "angular-mocks": "^1.6.5",
     "awesome-typescript-loader": "^3.2.2",
     "copy-webpack-plugin": "^4.0.1",
--- a/package.json
+++ b/package.json
@@ -69,7 +69,6 @@
     "@types/offline-js": "^0.7.28",
     "@types/quill": "1.3.2",
     "@types/rangy": "0.0.30",
-    "angular-inview": "^2.1.0",
     "angular-mocks": "^1.6.5",
     "awesome-typescript-loader": "^3.2.2",
     "copy-webpack-plugin": "^4.0.1",
--- a/src/Api/Model/Languageforge/Lexicon/LexCommentModel.php
+++ b/src/Api/Model/Languageforge/Lexicon/LexCommentModel.php
@@ -53,6 +53,10 @@ class LexCommentModel extends MapperModel
     /** @var LexCommentFieldReference */
     public $regarding;
 
+    /** @var string
+     *  This is either a sense, example or pronunciation guid that references the location of this comment */
+    public $contextGuid;
+
     /** @var int */
     public $score;
 
--- a/src/angular-app/bellows/core/offline/lexicon-comments.service.ts
+++ b/src/angular-app/bellows/core/offline/lexicon-comments.service.ts
@@ -9,6 +9,7 @@ class Comment {
   id: string;
   entryRef: string;
   regarding: any;
+  contextGuid: string;
   replies: any[];
   score: number;
   status: string;
@@ -79,23 +80,28 @@ export class LexiconCommentService {
         promises.push(this.lexConfig.getFieldConfig(comment.regarding.field).then(fieldConfig => {
           // As the promise runs when its ready the comments can double up if loadEntryComments is run multiple times
           if (this.comments.items.currentEntry.indexOf(comment) === -1) {
-            let contextId = comment.regarding.field + '_' + comment.regarding.inputSystemAbbreviation;
-            if (fieldConfig.type === 'pictures' && comment.regarding.fieldValue.indexOf('/assets/') !== -1) {
-              contextId += '_' + comment.regarding.fieldValue; // Would prefer to use the picture ID
-            } else if (fieldConfig.type === 'multioptionlist') {
-              contextId += '_' + comment.regarding.fieldValue;
+            let contextGuid = '';
+            if (comment.contextGuid !== undefined) {
+              contextGuid = comment.contextGuid;
+            } else {
+              contextGuid = comment.regarding.field +
+                (comment.regarding.inputSystemAbbreviation ? '.' + comment.regarding.inputSystemAbbreviation : '');
+              if (fieldConfig.type === 'multioptionlist') {
+                contextGuid += '#' + comment.regarding.fieldValue;
+              }
+              comment.contextGuid = contextGuid;
             }
 
-            if (contextId && angular.isUndefined(this.comments.counts.currentEntry.fields[contextId])) {
-              this.comments.counts.currentEntry.fields[contextId] = 0;
+            if (contextGuid && angular.isUndefined(this.comments.counts.currentEntry.fields[contextGuid])) {
+              this.comments.counts.currentEntry.fields[contextGuid] = 0;
             }
 
             this.comments.items.currentEntry.push(comment);
 
             // update the appropriate count for this field and update the total count
             if (comment.status !== 'resolved') {
-              if (contextId) {
-                this.comments.counts.currentEntry.fields[contextId]++;
+              if (contextGuid) {
+                this.comments.counts.currentEntry.fields[contextGuid]++;
               }
 
               this.comments.counts.currentEntry.total++;
@@ -123,9 +129,9 @@ export class LexiconCommentService {
     }
   }
 
-  getFieldCommentCount(contextId: string): number {
-    if (angular.isDefined(this.comments.counts.currentEntry.fields[contextId])) {
-      return this.comments.counts.currentEntry.fields[contextId];
+  getFieldCommentCount(contextGuid: string): number {
+    if (angular.isDefined(this.comments.counts.currentEntry.fields[contextGuid])) {
+      return this.comments.counts.currentEntry.fields[contextGuid];
     }
 
     return 0;
--- a/src/angular-app/bellows/core/offline/lexicon-comments.service.ts
+++ b/src/angular-app/bellows/core/offline/lexicon-comments.service.ts
@@ -72,18 +72,18 @@ export class LexiconCommentService {
     this.comments.counts.currentEntry.fields = {};
     this.comments.items.currentEntry.length = 0;
     for (const comment of this.comments.items.all) {
-      const fieldName = comment.regarding.field  + '_' + comment.regarding.inputSystemAbbreviation;
+      const contextId = comment.regarding.field  + '_' + comment.regarding.inputSystemAbbreviation;
       if (comment.entryRef === entryId) {
-        if (fieldName && angular.isUndefined(this.comments.counts.currentEntry.fields[fieldName])) {
-          this.comments.counts.currentEntry.fields[fieldName] = 0;
+        if (contextId && angular.isUndefined(this.comments.counts.currentEntry.fields[contextId])) {
+          this.comments.counts.currentEntry.fields[contextId] = 0;
         }
 
         this.comments.items.currentEntry.push(comment);
 
         // update the appropriate count for this field and update the total count
         if (comment.status !== 'resolved') {
-          if (fieldName) {
-            this.comments.counts.currentEntry.fields[fieldName]++;
+          if (contextId) {
+            this.comments.counts.currentEntry.fields[contextId]++;
           }
 
           this.comments.counts.currentEntry.total++;
--- a/src/angular-app/bellows/core/offline/lexicon-comments.service.ts
+++ b/src/angular-app/bellows/core/offline/lexicon-comments.service.ts
@@ -72,7 +72,7 @@ export class LexiconCommentService {
     this.comments.counts.currentEntry.fields = {};
     this.comments.items.currentEntry.length = 0;
     for (const comment of this.comments.items.all) {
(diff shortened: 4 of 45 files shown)


megahirt:
Hi Nigel,

This is great work.  I'm looking forward to integrating this as soon as we can.  As soon as you reply to my comments and fix the ES6 Promise, then I will merge this into master and begin testing.  At that point I'll someone to get the senseId and exampleId (likely also needed) into the comment object.  You can also submit a separate PR to remove the angular in-view code as I agree with you that it's not really that helpful.

---

Reviewed 42 of 42 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[package.json, line 72 at r2](https://reviewable.io:443/reviews/sillsdev/web-languageforge/181#-L4-KBQ9mJuOxETmY1Wz:-L4-KBQAK5z0b5SPQVeg:bi3lxbd) ([raw file](https://github.com/sillsdev/web-languageforge/blob/c6a7cef6ba33461dda4072aefe36b905ce5af021/package.json#L72)):*
> ```JSON
>     "@types/quill": "1.3.2",
>     "@types/rangy": "0.0.30",
>     "angular-inview": "^2.1.0",
> ```

Based on your analysis of how the feature works in-practice, I am in favor of removing angular-inview and not hiding bubbles out of view.  So basically revert this functionality in a future PR.  I am happy to accept this PR with this dependency for the moment.

---

*[src/angular-app/bellows/core/offline/lexicon-comments.service.ts, line 72 at r2](https://reviewable.io:443/reviews/sillsdev/web-languageforge/181#-L4-LbddCkBN7QIdUmLL:-L4-LbddCkBN7QIdUmLM:b-c4zrin) ([raw file](https://github.com/sillsdev/web-languageforge/blob/c6a7cef6ba33461dda4072aefe36b905ce5af021/src/angular-app/bellows/core/offline/lexicon-comments.service.ts#L72)):*
> ```XML
>    * This should be called whenever the entry context changes (to update the comments and comment counts)
>    */
>   loadEntryComments(entryId: string): angular.IPromise<any> {
> ```

As of Typescript > 2.0 we can use native ES6 Promises.  No need to import an AngularJS implementation.  Our tsconfig.json already includes lib es7 so we should be good to go.  So you should just be able to do
`loadEntryComments(entryId: string): Promise<any>`

references:
https://stackoverflow.com/questions/27573365/how-to-use-typescript-with-native-es6-promises
http://www.datchley.name/es6-promises/

---

*[src/angular-app/bellows/core/offline/lexicon-comments.service.ts, line 76 at r2](https://reviewable.io:443/reviews/sillsdev/web-languageforge/181#-L4-MhTp_ihnsKNQ0fTs:-L4-MhTp_ihnsKNQ0fTt:b-h2k1oi) ([raw file](https://github.com/sillsdev/web-languageforge/blob/c6a7cef6ba33461dda4072aefe36b905ce5af021/src/angular-app/bellows/core/offline/lexicon-comments.service.ts#L76)):*
> ```XML
>     this.comments.counts.currentEntry.fields = {};
>     this.comments.items.currentEntry.length = 0;
>     const defer = this.$q.defer();
> ```

To use ES6 Promises, you'll need to change to the pattern:

return new Promise((resolve, reject) => {
resolve();
...
});

---

*[src/angular-app/languageforge/lexicon/editor/editor.js, line 702 at r2](https://reviewable.io:443/reviews/sillsdev/web-languageforge/181#-L4-P8d3vidqxnZVY1y7:-L4-P8d3vidqxnZVY1y8:b-561b38) ([raw file](https://github.com/sillsdev/web-languageforge/blob/c6a7cef6ba33461dda4072aefe36b905ce5af021/src/angular-app/languageforge/lexicon/editor/editor.js#L702)):*
> ```JavaScript
>         if ($scope.commentPanelVisible !== true) {
>           $scope.commentPanelVisible = true;
>           angular.element('.comments-right-panel-container').addClass('panel-opening');
> ```

So this is the way to implement the animation, correct?  We don't have angular-animate as a dependency right now.  Would adding angular-animate back in be something useful for the UI redesign?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/181#-:-L4-QZCTpXHdtzBXivDJ:b-wwdfd9)*

rmunn:
I've spent a little time looking through the design of the comment context ID, and thinking about how it could incorporate multiple senses and examples. An additional wrinkle in the design is that we need to handle custom fields, which could be named anything, and we need to be able to do Send/Receive between FieldWorks and Language Forge. So here's my suggestion for how we should handle context IDs:

1. The context ID of a field belonging to a sense will be `senses.##.fieldName`, where `##` is a number starting with 1 for the first sense.
1. The context ID of a field belonging to an example will be `senses.##.examples.##.fieldName`, with both the sense number and the example number specified in that context ID.
1. Since custom fields are allowed to be named *anything*, including names with periods in them, it would be remotely possible for a user to create a custom field named "senses.1", which would probably cause unexpected behavior in the context ID code. I propose that the LF code which allows users to create custom fields should become slightly stricter, and forbid any custom field names that begin with `senses.##` and/or `examples.##`, where `##` is any series of digits.

Note regarding that last point: there's no danger of such a custom field name coming from FieldWorks via a Send/Receive, because the FieldWorks code forbids all punctuation in custom field names. We could go farther and forbid all punctuation in custom field names, but I think some users might want custom fields named things like "Notes (Sam)" and "Notes (Janet)", so I'd rather just forbid the names `senses.##` and `examples.##` rather than going whole-hog and forbidding all punctuation.

Changing the context ID to be a hierarchical (dot-separated) list of names like this would have implications for various other parts of the code, like `getFieldConfig` in `src/angular-app/languageforge/lexicon/core/lexicon-config.service.ts`, but @nigel-wells has already realized that and made appropriate changes to those functions. My suggestion would (as far as I can tell) require only minor tweaks to what @nigel-wells has already done, so I think it will be the fastest to implement. It's the end of the day here so I'm not going to start working on this until tomorrow. Any thoughts? Further refinements to this suggestion?

megahirt:
Robin, we can talk more about this tomorrow but the context id is, at least
from Nigel's perspective, the senseId.  We already have one of these.  It's
just not exposed yet (your task).  Sense number isn't a good idea because
senses can be reordered.  We don't have a UI for that but it's something
that users have asked for.  This is already a feature in FLEx.

megahirt:
I don't see how custom fields are a problem in this situation since they
are uniquely encoded in the client side.  One potential issue though is
that the context id could refer to a sense or an example or (in the future)
a pronunciation.  These three things are list types.  If we presume the id
to be globally unique, then the matching code would need to traverse all
three list types to find a matching context.

nigel-wells:
Review status: 34 of 39 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/angular-app/bellows/core/offline/lexicon-comments.service.ts, line 72 at r2](https://reviewable.io:443/reviews/sillsdev/web-languageforge/181#-L4-LbddCkBN7QIdUmLL:-L430kcLPUGzohHLAHcw:b-wntwbp) ([raw file](https://github.com/sillsdev/web-languageforge/blob/c0728fb36ec89c17be1629795a3c5ac00c8e4b48/src/angular-app/bellows/core/offline/lexicon-comments.service.ts#L72)):*

As of Typescript > 2.0 we can use native ES6 Promises.  No need to import an AngularJS implementation.  Our tsconfig.json already includes lib es7 so we should be good to go.  So you should just be able to do
`loadEntryComments(entryId: string): Promise<any>`

references:
https://stackoverflow.com/questions/27573365/how-to-use-typescript-with-native-es6-promises
http://www.datchley.name/es6-promises/

</blockquote></details>

I've updated this to the ES6 version

---

*[src/angular-app/bellows/core/offline/lexicon-comments.service.ts, line 76 at r2](https://reviewable.io:443/reviews/sillsdev/web-languageforge/181#-L4-MhTp_ihnsKNQ0fTs:-L4311WZR9-Z48CdkIe1:b-3jzjvq) ([raw file](https://github.com/sillsdev/web-languageforge/blob/c0728fb36ec89c17be1629795a3c5ac00c8e4b48/src/angular-app/bellows/core/offline/lexicon-comments.service.ts#L76)):*

To use ES6 Promises, you'll need to change to the pattern:

return new Promise((resolve, reject) => {
resolve();
...
});
</blockquote></details>

Have updated though it was slightly different as the  Promise is returned based on an array of promises:

return Promise.all(promises);

---

*[src/angular-app/languageforge/lexicon/editor/editor.js, line 702 at r2](https://reviewable.io:443/reviews/sillsdev/web-languageforge/181#-L4-P8d3vidqxnZVY1y7:-L43191A5N_-KNH35zTD:bgel05d) ([raw file](https://github.com/sillsdev/web-languageforge/blob/c0728fb36ec89c17be1629795a3c5ac00c8e4b48/src/angular-app/languageforge/lexicon/editor/editor.js#L702)):*

So this is the way to implement the animation, correct?  We don't have angular-animate as a dependency right now.  Would adding angular-animate back in be something useful for the UI redesign?
</blockquote></details>

Due to how I decided to do the opening and closing animation I needed a new class that only got applied for the closing of the panel. All the animation is done via CSS which even angular-animate relies on unless you're using a JavaScript animation. I'm happy with the below though it is technically using jQuery to add/remove the class. We could make use of another scope variable if you wanted to and then trigger the class that way?

---

*[package.json, line 72 at r2](https://reviewable.io:443/reviews/sillsdev/web-languageforge/181#-L4-KBQ9mJuOxETmY1Wz:-L431aUfCxwrBGcNRawU:b-fge7rq) ([raw file](https://github.com/sillsdev/web-languageforge/blob/c0728fb36ec89c17be1629795a3c5ac00c8e4b48/package.json#L72)):*

Based on your analysis of how the feature works in-practice, I am in favor of removing angular-inview and not hiding bubbles out of view.  So basically revert this functionality in a future PR.  I am happy to accept this PR with this dependency for the moment.
</blockquote></details>

As I was pushing other code changes through I've just gone ahead and removed this from all the related files.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/181)*

nigel-wells:
Good discussion on the context id. @megahirt can you confirm for me, as I've not seen it yet, if the senseId is a unique key of some kind that doesn't change regardless of how the entry view is created? Just thinking ahead to when you adapt the UI to re-order senses and making sure this doesn't break comments i.e. if comments were made on the first sense which has now been moved to be sense number 3. If the senseId is unique, and would never change, then we'll be fine. My ideal would be for every piece of possible information on the entry form to have its own unique key (I'm used to working in MySQL and having unique and/or primary keys) and this way it would never matter how the entry view is built. It would also make it easier to have comments displaying in other areas of the app as it would no longer have to make any assumptions of context based on the make up of the entry view.

megahirt:
yes, the assumption is that the senseId and exampleId are globally unique ids in the system

---

Reviewed 6 of 8 files at r3.

---

*[src/angular-app/languageforge/lexicon/editor/editor.js, line 702 at r2](https://reviewable.io:443/reviews/sillsdev/web-languageforge/181#-L4-P8d3vidqxnZVY1y7:-L43ZIYyV5cww0z8z3mt:b-lbmwdq) ([raw file](https://github.com/sillsdev/web-languageforge/blob/c0728fb36ec89c17be1629795a3c5ac00c8e4b48/src/angular-app/languageforge/lexicon/editor/editor.js#L702)):*
<details><summary><i>Previously, nigel-wells (Nigel Wells) wrote…</i></summary><blockquote>

Due to how I decided to do the opening and closing animation I needed a new class that only got applied for the closing of the panel. All the animation is done via CSS which even angular-animate relies on unless you're using a JavaScript animation. I'm happy with the below though it is technically using jQuery to add/remove the class. We could make use of another scope variable if you wanted to and then trigger the class that way?
</blockquote></details>

We have been trying to remove dependency on jQuery in hopes of removing it.  AngularJS ships with a jQuery lite implementation that we will naturally have available to us as long as we're using AngularJS.  Does the element().addClass() method work with the built-in jQuery lite if we were to remove the full jQuery down the road?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/181#-:-L43ZpHAi8Co4jMhSL_t:babqwcp)*

nigel-wells:

---

*[src/angular-app/languageforge/lexicon/editor/editor.js, line 702 at r2](https://reviewable.io:443/reviews/sillsdev/web-languageforge/181#-L4-P8d3vidqxnZVY1y7:-L43v1O12VM41Xno5wR8:b-7ev3jy) ([raw file](https://github.com/sillsdev/web-languageforge/blob/c0728fb36ec89c17be1629795a3c5ac00c8e4b48/src/angular-app/languageforge/lexicon/editor/editor.js#L702)):*

We have been trying to remove dependency on jQuery in hopes of removing it.  AngularJS ships with a jQuery lite implementation that we will naturally have available to us as long as we're using AngularJS.  Does the element().addClass() method work with the built-in jQuery lite if we were to remove the full jQuery down the road?
</blockquote></details>

Yes it does - I made sure to check it there. Docs are here for what's included: https://docs.angularjs.org/partials/api/ng/function/angular.element.html

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/181)*

nigel-wells:
PR now has logic for contextGuid as well as a few other bug fixes.

I've squashed the 3 initial comments and changed the author to all be my nigel-wells account.

Existing comments without a contextGuid have a fallback for the Entry card. The senses and example comments can be accessed by clicking the comments button in the header as normal. It may be worth writing an upgrade script that goes through the database and tries to create a contextGuid for as many comments as possible. For projects that only have a single sense and a single example could easily be updated with the correct contextGuid - just a thought anyway.

As discussed Chris, there is some kind of delay between when a sense or an example are added and their guid being applied from the server. For now I've set it to disable the ability to comment until the contextGuid is resolved. This will work fine for now and may not even be an issue as, be the time the user has completed filling out the fields in the first place, the contextGuid will most likely be available.

Let me know if you require any further changes.

megahirt:
Great!  Can't wait to test this.

---

Reviewed 1 of 2 files at r4, 22 of 22 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/181#-:-L4KlIzJXO-bUFTUlZ9N:bq41afs)*

PR #182, opened 2018-02-01
Better handling of OAuth errors including timeouts

rmunn:
Now if we get a connection timed out on getting user details, we'll retry once. If that fails, or if anything else goes wrong, we'll redirect to the login page (with an error message that basically says "Try again, but if the problem persists, try username & password").

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/182)

### DIFF
--- a/src/Site/OAuth/OAuthBase.php
+++ b/src/Site/OAuth/OAuthBase.php
@@ -217,7 +217,17 @@ abstract class OAuthBase extends Base
     protected function loginWithOAuthToken(Application $app, AbstractProvider $provider, OAuthAccessToken $token)
     {
         try {
-            $userDetails = $provider->getResourceOwner($token);
+            try {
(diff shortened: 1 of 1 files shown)


rmunn:
I can't verify that this catches connection timeouts properly as I don't know how to force a connection timeout to Google's servers (I only know how to force a connection timeout to a server I control). But I was able to test that normal OAuth login still works, and that any exceptions thrown during the process will now redirect the user to the login page with a friendly error message at the top:

image

So this is about as tested as I can make it, and I believe it's ready to merge.

megahirt:
:lgtm: Nice work!

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/182#-:-L4KT5_ERDKlspUkDDkc:bcmw7ss)*

PR #183, opened 2018-02-01
Fix comments export to ParaText 7 and/or 8

rmunn:
PR originally on [web-scriptureforge repo](https://github.com/sillsdev/web-scriptureforge/pull/128), transferred over to this repo instead. Will probably squash-and-merge rather than merge the branch.

This fixes comment exporting for both Paratext 7 and 8. I've changed the "Prepare Export" button to be two buttons, one for Paratext 7 and one for Paratext 8 instead, and also changed the two-step process into a single-step process. (You used to click "Prepare", then wait a short while, then click "Download", but it makes more sense to download the file as soon as it's available. Thanks to @megahirt for the suggestion.) Screenshot:

screenshot

One feature NOT included in this PR is the ability to export all comments, not just those in a single text. The code is in place, we just have to implement it and add the UI to the project settings screen. But this branch will at least get the feature working first; then we can add convenience.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/183)

### DIFF
--- a/src/Api/Library/Scriptureforge/Sfchecks/ParatextExport.php
+++ b/src/Api/Library/Scriptureforge/Sfchecks/ParatextExport.php
@@ -28,18 +28,26 @@ class ParatextExport
             'xml' => '<?xml version="1.0" encoding="utf-8"?>' . "\n<CommentList>\n"
         );
         $commentFormatter = "formatForPT7";
+        $filenamePrefix = "Comments";
         if (isset($params['commentFormat'])) {
             switch ($params['commentFormat'])
             {
             case "PT7":
                 $commentFormatter = "formatForPT7";
+                $filenamePrefix = "Comments";
                 break;
             case "PT8":
                 $commentFormatter = "formatForPT8";
+                $filenamePrefix = "Notes";
                 break;
             // No need for a default case since we've already set the default above
             }
         }
+
+        $dateForFilename = date('Ymd_Gi');
+        $dateTimeForFilename = \DateTime::createFromFormat('Ymd_Gi', $dateForFilename);
+        $dl['xml'] .= self::makeDummyComment($commentFormatter, $dateTimeForFilename, $dateForFilename);
+
         foreach ($questionlist->entries as $question) {
             if (! array_key_exists('isArchived', $question) || ! $question['isArchived']) {
                 foreach ($question['answers'] as $answerId => $answer) {
@@ -59,7 +67,7 @@ class ParatextExport
         $dl['totalCount'] = $dl['answerCount'] + $dl['commentCount'];
         $dl['xml'] .= "</CommentList>";
 
-        $dl['filename'] = 'Comments_sf_' . date('Ymd_Gi') . '.xml';
+        $dl['filename'] = $filenamePrefix . '_SF-' . $dateForFilename . '.xml';
         //$dl['filename'] = preg_replace("([^\w\d\-]|[\.]{2,})", '_', $filename) . '.xml';
         return $dl;
     }
@@ -112,9 +120,14 @@ class ParatextExport
         return call_user_func(["\Api\Library\Scriptureforge\Sfchecks\ParatextExport", $commentFormatter], $threadId, $username, $textInfo, "", $comment['dateEdited']->toDateTime(), $content);
     }
 
+    private static function makeDummyComment($commentFormatter, \DateTime $dateTime, string $dummyCommenterName)
+    {
+        return call_user_func(["\Api\Library\Scriptureforge\Sfchecks\ParatextExport", $commentFormatter], $dummyCommenterName, $dummyCommenterName, [], "", $dateTime, "");
+    }
+
     private static function formatVerseRef($textInfo) : string
     {
-        return $textInfo['bookCode'] . " " . $textInfo['startChapter'] . ":" . $textInfo['startVerse'];
+        return empty($textInfo) ? "" : ($textInfo['bookCode'] . " " . $textInfo['startChapter'] . ":" . $textInfo['startVerse']);
     }
 
     public static function commentHeaderForPT7(string $threadId, string $username, string $verseRef, string $language, \DateTime $dateTime) : string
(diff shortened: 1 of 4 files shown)


megahirt:
I have a few comments, but everything else looks good.  Nice work Robin!

---

Reviewed 4 of 4 files at r1.

---

*[src/angular-app/scriptureforge/sfchecks/partials/questions-settings.html, line 154 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/183#-L4KOdPyvKh_paA6SiNZ:-L4KOdPyvKh_paA6SiN_:b-gisgt0) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2258e41110949b11bc8cb58cd21f8d08b80ce300/src/angular-app/scriptureforge/sfchecks/partials/questions-settings.html#L154)):*
> ```HTML
>                 <div class="col-sm-6" data-ng-show="download.totalCount">
>                     <br class="d-sm-none">
>                     <p data-ng-show="download.totalCount > 0" class="text-success">
> ```

I think it would still be nice to see the number of comments downloaded, even after the user has saved the file.  Can we still keep the answerCount and commentCount in the display somewhere?

---

*[src/Api/Library/Scriptureforge/Sfchecks/ParatextExport.php, line 120 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/183#-L4KPk-Jn4m2ilE9mPNX:-L4KPk-Jn4m2ilE9mPNY:b-okl9x8) ([raw file](https://github.com/sillsdev/web-languageforge/blob/071570172a7c1f9a11ea8fd505f5a66f53eb54e3/src/Api/Library/Scriptureforge/Sfchecks/ParatextExport.php#L120)):*
> ```PHP
> 
>         // Is there a better way than specifying the function name *and* class name as "magic strings"? I'm unfamiliar with how to do this in PHP. - 2018-01 RM
>         return call_user_func(["\Api\Library\Scriptureforge\Sfchecks\ParatextExport", $commentFormatter], $threadId, $username, $textInfo, "", $comment['dateEdited']->toDateTime(), $content);
> ```

I don't think self::$commentFormatter actually works, but how about:

```
if ($commentFormatter == "formatForPT7") {
    return self::formatForPT7($threadId, $username, $textInfo, "", $comment['dateEdited']->toDateTime(), $content);
} else {
    return self::formatForPT8($threadId, $username, $textInfo, "", $comment['dateEdited']->toDateTime(), $content);
}
```

But seeing it now it looks like maybe you really only need a self::formatComment() which takes a $format parameter, and inside you can switch formats based on that.  The methods seem really similar to me at this point (although they could in theory diverge more later)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/183#-:-L4KSElDpX5Mrpo-vivC:b-olxkpo)*

rmunn:

---

*[src/angular-app/scriptureforge/sfchecks/partials/questions-settings.html, line 154 at r1](https://beta.reviewable.io/reviews/sillsdev/web-languageforge/183#-L4KOdPyvKh_paA6SiNZ:-L4Kb9id2ttOQfsM8O0_:bjq9g8u) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2258e41110949b11bc8cb58cd21f8d08b80ce300/src/angular-app/scriptureforge/sfchecks/partials/questions-settings.html#L154)):*

I think it would still be nice to see the number of comments downloaded, even after the user has saved the file.  Can we still keep the answerCount and commentCount in the display somewhere?
</blockquote></details>

Yes, I can make it display as the file starts downloading.

---

*[src/Api/Library/Scriptureforge/Sfchecks/ParatextExport.php, line 120 at r1](https://beta.reviewable.io/reviews/sillsdev/web-languageforge/183#-L4KPk-Jn4m2ilE9mPNX:-L4KbI4NEybk92nZkJoE:b-d430km) ([raw file](https://github.com/sillsdev/web-languageforge/blob/071570172a7c1f9a11ea8fd505f5a66f53eb54e3/src/Api/Library/Scriptureforge/Sfchecks/ParatextExport.php#L120)):*

I don't think self::$commentFormatter actually works, but how about:

```
if ($commentFormatter == "formatForPT7") {
    return self::formatForPT7($threadId, $username, $textInfo, "", $comment['dateEdited']->toDateTime(), $content);
} else {
    return self::formatForPT8($threadId, $username, $textInfo, "", $comment['dateEdited']->toDateTime(), $content);
}
```

But seeing it now it looks like maybe you really only need a self::formatComment() which takes a $format parameter, and inside you can switch formats based on that.  The methods seem really similar to me at this point (although they could in theory diverge more later)
</blockquote></details>

They are very similar; in fact, they're basically the same method except for which header method they call. I decided not to combine them into a single method, though, because I felt that the extra mental step to understand that combined method wouldn't be justified by the savings. After all, as you point out, they might diverge later, and this isn't a case where I'll want to change both methods at once in the future. So in this one case, factoring out the repeated code would (IMO) have added more complexity than it removed (though factoring out repeated code is *usually* what you want to do).

Yes, I think moving that `switch` statement into its own function (one that takes a `$format` parameter) probably makes the most sense. I'll do that before I merge the branch.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/sillsdev/web-languageforge/183)*

megahirt:

---

*[src/angular-app/scriptureforge/sfchecks/partials/questions-settings.html, line 154 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/183#-L4KOdPyvKh_paA6SiNZ:-L4Km7ZMuIJzmwmJ0uFg:bp7vbga) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2258e41110949b11bc8cb58cd21f8d08b80ce300/src/angular-app/scriptureforge/sfchecks/partials/questions-settings.html#L154)):*

Yes, I can make it display as the file starts downloading.
</blockquote></details>

sounds good.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/183)*

megahirt:
@rmunn merge this one yourself when you're ready

PR #184, opened 2018-02-08
Apex/comments fixes

nigel-wells:
Latest round of updates to the new comments panel integration into the entry view.

Updates include:
- Submit/reply buttons are now disabled when clicking submit button to avoid double clicks
- When viewing all comments you can now click the main comments button again to close the panel
- Changing entries now also forces a change of the context and entry ID for any new comments
- View original meaning will now only display if there is an original value (for historic comments)
- Can now filter by resolved comments
- Improved layout for adding a new comment under and existing thread so as not to confuse it with replying
- Deleting comments now recalculates the total comments on the entry correctly
- Option list field types now display the label instead of the value stored with the comment

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/184)

### DIFF
--- a/src/angular-app/bellows/core/offline/lexicon-comments.service.ts
+++ b/src/angular-app/bellows/core/offline/lexicon-comments.service.ts
@@ -117,6 +117,7 @@ export class LexiconCommentService {
    * this should be called whenever new data is received
    */
   updateGlobalCommentCounts(): void {
+    this.comments.counts.byEntry = {};
     for (const comment of this.comments.items.all) {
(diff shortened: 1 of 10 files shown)


megahirt:
:lgtm: thanks!

---

Reviewed 10 of 10 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/184#-:-L4oHGGkgGOoB4-ItdBB:brfqlux)*

PR #185, opened 2018-02-08
Fixed comment context changing when editing other fields

nigel-wells:
Fixed the latest bug in Trello that was introduced from the last PR.

- Fixed dynamic updating of comment context only when using an optionlist field type
- Only clear an existing comment after posting a comment and not on a refresh

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/185)


rmunn:
I can still trigger the bug by starting a comment on the Word field, then changing the Part of Speech, so the fix related to optionlists needs to be a bit more detailed. This is good enough to cover most cases, though.

PR #186, opened 2018-02-08
Fix when adding a comment and then changing an optionlist field befor…

nigel-wells:
Latest PR to address remaining issue with the context changing before submitting a comment:

- Fix when adding a comment and then changing an optionlist field before submitting the comment
- Updated E2E tests to test to ensure comments made appear within the intended context

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/186)


rmunn:
Tested locally, looks good.

PR #188, opened 2018-02-12
Latest fixes to the new comments panel

nigel-wells:
Latest fixes to the new comments panel

- Setup watch for to wait for a promise to resolve on the input system
- Force an new entry to save prior to adding any comments

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/188)

### DIFF
--- a/src/angular-app/bellows/directive/palaso.ui.comments.comment-bubble.js
+++ b/src/angular-app/bellows/directive/palaso.ui.comments.comment-bubble.js
@@ -14,8 +14,8 @@ angular.module('palaso.ui.comments')
         multiOptionValue: '<',
         picture: '<'
       },
-      controller: ['$scope', 'lexCommentService', 'sessionService', '$element', 'lexConfigService',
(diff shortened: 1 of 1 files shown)


megahirt:
:lgtm: Thanks Nigel.  Let me just make a comment here that this behavior is caused by a systemic design flaw where critical data required to render the view comes down in two separate network calls.  This is something we wish to refactor.  There is a "getSessionData" method which returns data needed for the view, as well as the view's DTO.   This needs to be refactored into a single get[Viewname]DTO which will contain the session data in it.  All of the promises related to getting config information would go away, as that would come down at the same time as the DTO itself.  So right now we have some nasty behavior which you are fixing with a watch.

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/188#-:-L56s5qWIDairlsJEpjk:b307ebg)*

PR #189, opened 2018-02-15
Fix operation of slider while playing and position of audio player

Nateowami:
This PR does two things:

- Previously when the audio was playing and the user tried to seek, the slider would keep jerking forward to the current playback position. Now the slider stops updating when the user is using it.
- Previously the audio player on SF was to the left of the title and expanded to show playback only when playing. Now it is below the title and the playback control is always visible.

I have not succeeded in running the e2e tests for this branch. As it stands I am unable to get any of the tests to run properly on master, so either they're broken on master already or (more likely) there's a problem with my setup.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/189)

### DIFF
--- a/src/Site/views/scriptureforge/theme/default/sass/_global.scss
+++ b/src/Site/views/scriptureforge/theme/default/sass/_global.scss
@@ -363,10 +363,6 @@ $vote-width: 2em;
   display: block;
 }
 
-.audio-buttons {
-  margin-right: 0.5em;
-}
-
 .fieldset {
   legend {
     width: auto;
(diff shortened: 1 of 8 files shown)


rmunn:
@Nateowami I've run the E2E tests for both LF and SF on your PR, and they all pass. If you're having a problem running them, it might be because our `refreshDeps.sh` script is currently pulling in an outdated version of `chromedriver`, which doesn't work with the latest Chrome. If the behavior you're seeing is that the E2E tests just sit forever at the login page and appear to do nothing, that could be the cause.

To fix this issue, go to https://chromedriver.storage.googleapis.com/index.html and pull down the latest version of ChromeDriver (as of this writing, that's 2.35). Then unzip it and move the `chromedriver` file to `node_modules/protractor/node_modules/webdriver-manager/selenium/chromedriver_2.26` (yes, with the version number `2.26` since the E2E code looks for that specific version number). This is a temporary workaround until we figure out why NPM is giving us outdated versions of chromedriver instead of the current version; once we figure that out, we'll fix our `gulp test-e2e-webdriver_update` target (which `refreshDeps.sh` calls) and then it will be easier again.

megahirt:
Thank you @Nateowami

PR #190, opened 2018-02-16
Activity feed API for Language Forge

rmunn:
The activity list API is missing a couple of unit tests for the new entry or project API calls, but I've tested them by hand and they do return the right DTO data structure. I also think that some thought could be given to how the "old" API could be improved, at least with a view to making the names more consistent (e.g., `getActivityForUser` doesn't actually get the activity of a single user; rather, it gets *all* available activity, and the "ForUser" part of the name just refers to which user should be considered as having *seen* all that activity. Whereas `getActivityForProject` is actually getting the activity for a single project. The naming is inconsistent at the moment, which may cause future confusion).

This is ready for review, and could be merged into master so that work on the frontend can proceed while I write unit tests. I wouldn't want to deploy it without unit tests, but if merging this will allow frontend work to proceed faster, then we can merge it now and I'll create another PR for unit tests.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/190)

### DIFF
--- a/src/Api/Model/Languageforge/Lexicon/Command/LexEntryCommands.php
+++ b/src/Api/Model/Languageforge/Lexicon/Command/LexEntryCommands.php
@@ -55,6 +55,7 @@ class LexEntryCommands
         $project->lastEntryModifiedDate = $now;
         if (array_key_exists('id', $params) && $params['id'] != '') {
             $entry = new LexEntryModel($project, $params['id']);
+            $oldEntry = new LexEntryModel($project, $params['id']); // NOT $entry = $oldEntry
             $action = 'update';
         } else {
             $entry = new LexEntryModel($project);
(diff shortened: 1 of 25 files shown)


rmunn:
Also just realized I forgot to implement the limit/skip parameters in the Typescript API. I'll make that change on Monday.

rmunn:
Talked to Chris about this PR today, and he asked me to merge it without waiting for a review.

megahirt:
:lgtm:  Thanks Robin

---

Reviewed 21 of 24 files at r1, 4 of 4 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/190#-:-L5rF8l5lIs3Y5ZIfdB9:b-clyb18)*

src/Api/Model/Languageforge/Lexicon/Command/LexEntryCommands.php
```diff
@@ -55,6 +55,7 @@ public static function updateEntry($projectId, $params, $userId, $mergeQueuePath
         $project->lastEntryModifiedDate = $now;
         if (array_key_exists('id', $params) && $params['id'] != '') {
             $entry = new LexEntryModel($project, $params['id']);
+            $oldEntry = new LexEntryModel($project, $params['id']); // NOT $entry = $oldEntry
```
irahopkinson: @rmunn do you know about PHP object cloning? e.g. `$entry = clone $oldEntry;`
See http://php.net/manual/en/language.oop5.cloning.php

PR #193, opened 2018-02-21
Commenting disabled for observers and anyone with no rights

nigel-wells:
Applies to the following Trello cards:

https://trello.com/c/QqQ8Q8Od/1084-comment-bubbles-are-visible-for-observer-role-but-this-should-not-be-the-comment-bubbles-should-only-be-visible-for-the-permissi

https://trello.com/c/6q97M4FK/905-during-s-r-users-should-be-locked-out-of-comments

As I don't have the ability to test a S/R I am unable to completely test that it works. The logic is the same as the observer so in theory it will be fine. I tested it to some extent with the observer role so I'm pretty sure it will be ok. For the send receive in progress I decided not to hide the entire commenting panel if it was open but rather just hide the textarea and any buttons that would allow edit/delete functionality. If we decided to hide it then I suspect the user would get frustrated with it opening and closing after every little change to the entry. As I couldn't test a S/R I can't be 100% sure though.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/193)

### DIFF
--- a/src/angular-app/bellows/directive/palaso.ui.comments.comment-bubble.js
+++ b/src/angular-app/bellows/directive/palaso.ui.comments.comment-bubble.js
@@ -111,7 +111,7 @@ angular.module('palaso.ui.comments')
         $scope.setContextGuid();
 
         $scope.isCommentingAvailable = function isCommentingAvailable() {
-          return ($scope.control.currentEntry.id.indexOf('_new_') !== -1);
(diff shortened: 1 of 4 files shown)


megahirt:
:lgtm:

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/sillsdev/web-languageforge/193#-:-L6Eu-Rt1ad4L4Agia9Z:bnfp4nl)*

PR #194, opened 2018-02-22
Mongo mapper trims empty fields

rmunn:
There are a few fields that will cause test failures if they're trimmed when empty, because of some assumptions in our code (like "a question will always have an answers array", for example). The `shouldKeepKey` function ensures that those fields will still be persisted even if they're empty. This is not a great long-term solution; it would be better to find and correct all those assumptions. But in the short term, this works and makes all PHP unit tests pass.

All PHP unit tests passing with this PR, and all E2E tests for Language Forge. (I haven't yet run E2E tests for Scripture Forge due to a misconfiguration on my laptop).

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/194)

### DIFF
--- a/src/Api/Model/Languageforge/Lexicon/Config/LexConfiguration.php
+++ b/src/Api/Model/Languageforge/Lexicon/Config/LexConfiguration.php
@@ -30,7 +30,7 @@ class LexConfiguration
     {
         $this->tasks = new MapOf(
             function ($data) {
-                switch ($data['type']) {
+                switch ($data['type'] ?? "") {
                     case LexTask::DASHBOARD:
                         return new LexTaskDashboard();
                     case LexTask::SEMDOM:
--- a/src/Api/Model/Languageforge/Semdomtrans/Command/SemDomTransItemCommands.php
+++ b/src/Api/Model/Languageforge/Semdomtrans/Command/SemDomTransItemCommands.php
@@ -16,10 +16,7 @@ class SemDomTransItemCommands
      */
     public static function update($data, $projectId) {
         $projectModel = new SemDomTransProjectModel($projectId);
-        $previousItemModel = new SemDomTransItemModel($projectModel);
-        if ($data["id"] != '') {
-            $previousItemModel->read($data['id']);
-        }
+        $previousItemModel = new SemDomTransItemModel($projectModel, $data["id"] ?? "");
 
         $guid = $data["xmlGuid"];
         $s = new SemDomTransItemModel($projectModel);
--- a/src/Api/Model/Scriptureforge/Sfchecks/Command/QuestionCommands.php
+++ b/src/Api/Model/Scriptureforge/Sfchecks/Command/QuestionCommands.php
@@ -141,7 +141,7 @@ class QuestionCommands
         $answerId = $question->writeAnswer($answer);
 
         // Re-read question model to pick up new answer
-        $question->read($questionId);
+        $question = new QuestionModel($project, $questionId);
         $newAnswer = $question->readAnswer($answerId);
         if ($answerJson['id'] != '') {
             // TODO log the activity after we confirm that the comment was successfully updated ; cjh 2013-08
@@ -235,7 +235,7 @@ class QuestionCommands
         JsonDecoder::decode($commentModel, $comment);
         $commentModel->userRef->id = $authorId;
         $commentId = QuestionModel::writeComment($projectModel->databaseName(), $questionId, $answerId, $commentModel);
-        $questionModel->read($questionId);
+        $questionModel = new QuestionModel($projectModel, $questionId);
         $newComment = $questionModel->readComment($answerId, $commentId);
         $commentDTO = QuestionCommentDto::encodeComment($newComment);
 
--- a/src/Api/Model/Scriptureforge/Sfchecks/Command/QuestionCommands.php
+++ b/src/Api/Model/Scriptureforge/Sfchecks/Command/QuestionCommands.php
@@ -21,11 +21,9 @@ class QuestionCommands
     {
         $projectModel = new ProjectModel($projectId);
         ProjectCommands::checkIfArchivedAndThrow($projectModel);
-        $questionModel = new QuestionModel($projectModel);
-        $isNewQuestion = ($object['id'] == '');
-        if (!$isNewQuestion) {
-            $questionModel->read($object['id']);
-        }
+        $questionId = $object['id'] ?? '';
+        $isNewQuestion = ($questionId == '');
+        $questionModel = new QuestionModel($projectModel, $questionId);
         JsonDecoder::decode($questionModel, $object);
         $questionId = $questionModel->write();
         if ($isNewQuestion) {
--- a/src/Api/Model/Scriptureforge/Sfchecks/Command/TextCommands.php
+++ b/src/Api/Model/Scriptureforge/Sfchecks/Command/TextCommands.php
@@ -54,11 +54,9 @@ class TextCommands
     {
         $projectModel = new ProjectModel($projectId);
         ProjectCommands::checkIfArchivedAndThrow($projectModel);
-        $textModel = new TextModel($projectModel);
-        $isNewText = ($object['id'] == '');
-        if (!$isNewText) {
-            $textModel->read($object['id']);
-        }
+        $textId = $object['id'] ?? '';
+        $isNewText = ($textId == '');
+        $textModel = new TextModel($projectModel, $textId);
         JsonDecoder::decode($textModel, $object);
         TextCommands::makeValidRange($object);
         if (TextCommands::hasRange($object)) {
--- a/src/Api/Model/Shared/Command/UserCommands.php
+++ b/src/Api/Model/Shared/Command/UserCommands.php
@@ -569,8 +569,7 @@ class UserCommands
     public static function sendJoinRequest($projectId, $userId, $website, DeliveryInterface $delivery = null)
     {
         $newUser = new UserModel($userId);
-        $project = new ProjectModel();
-        $project->read($projectId);
+        $project = new ProjectModel($projectId);
 
         // Make sure the user exists on the site
         if (!$newUser->hasRoleOnSite($website)) {
@@ -606,8 +605,7 @@ class UserCommands
     public static function acceptJoinRequest($projectId, $userId, $website, $role, DeliveryInterface $delivery = null)
     {
         $newUser = new UserModel($userId);
-        $project = new ProjectModel();
-        $project->read($projectId);
+        $project = new ProjectModel($projectId);
 
         ProjectCommands::updateUserRole($projectId, $userId, $role);
 
--- a/src/Api/Model/Shared/Mapper/MapperModel.php
+++ b/src/Api/Model/Shared/Mapper/MapperModel.php
@@ -40,7 +40,7 @@ class MapperModel extends ObjectForEncoding
      * @see MongoMapper::read()
      * @throws \Exception
      */
-    public function read($id)
+    protected function read($id)
     {
         if ($this->_mapper->exists($id)) {
             $this->_mapper->read($this, $id);
--- a/src/Api/Model/Shared/Mapper/MongoMapper.php
+++ b/src/Api/Model/Shared/Mapper/MongoMapper.php
@@ -370,6 +370,66 @@ class MongoMapper
     }
 
     /**
+     * Since MongoEncoder::encode returns new \stdClass() instead of an empty array, but empty(new \stdClass()) returns false, we need a different way to detect if an encoded object is empty.
+     * See https://stackoverflow.com/questions/9412126/how-to-check-that-an-object-is-empty-in-php
+     */
+    public static function objectIsEmpty($object) {
+        if (is_null($object)) return true;
+        foreach ($object as $key) {
+            return false;
+        }
+        return true;
+    }
+
+    public static function shouldPersist($value) {
+        if (is_null($value))
+            return false;
+        if (is_bool($value))
+            return true;
+        if (empty($value))
+            return false;
+        if (is_object($value) && self::objectIsEmpty($value))
+            return false;
+        return true;
+    }
+
+    public static function shouldKeepKey(string $key) {
+        // Some keys shouldn't be removed even if empty, at least as long as our code is still making assumptions that these keys exist
+        return ($key === "guid" || $key === "translation" || $key === "description" || $key === "answers");
+    }
+
+    public static function removeEmptyItems(array $array) {
+        foreach ($array as $key => &$value) {
+            if (is_array($value)) {
+                $value = self::removeEmptyItems($value);
+            }
+            if (self::shouldPersist($value) || self::shouldKeepKey($key)) {
+                $array[$key] = $value;
+            } else {
+                unset($array[$key]);
+            }
+        }
+        return $array;
+    }
+
+    protected function prepareUpdateCommand($data) {
+        // Returns two arrays: keysToSet and keysToUnset (both with key => value)
+        $keysToSet = [];
+        $keysToUnset = [];
+        foreach ($data as $key => $value) {
+            if (is_array($value)) {
+                $value = self::removeEmptyItems($value);
+            }
+            if (self::shouldPersist($value) || self::shouldKeepKey($key)) {
+                $keysToSet[$key] = $value;
+            } else {
+                $keysToUnset[$key] = null;
+            }
+        }
+        return [$keysToSet, $keysToUnset];
+    }
+
+    /**
      * @param array $data
      * @param string $id
      * @param int $keyType
@@ -385,7 +445,11 @@ class MongoMapper
             if (empty($rootId)) {
                 $mongoid = self::mongoID($id);
                 $filter = array('_id' => $mongoid);
-                $updateCommand = array('$set' => $data);
+                list($keysToSet, $keysToUnset) = $this->prepareUpdateCommand($data);
+                $updateCommand = array('$set' => $keysToSet);
+                if (! empty($keysToUnset)) {
+                    $updateCommand['$unset'] = $keysToUnset;
+                }
                 $options = array('upsert' => true);
                 $this->_collection->updateOne($filter, $updateCommand, $options);
                 $id = $mongoid->__toString();
--- a/test/php/model/languageforge/lexicon/command/LexCommentCommandsTest.php
+++ b/test/php/model/languageforge/lexicon/command/LexCommentCommandsTest.php
@@ -31,14 +31,13 @@ class LexCommentCommandsTest extends TestCase
             'fieldNameForDisplay' => 'Word',
             'inputSystemAbbreviation' => 'th',
             'inputSystem' => 'th',
-            'word' => '',
-            'meaning' => ''
         );
 
         $data = array(
             'id' => '',
             'content' => $commentContent,
-            'regarding' => $regarding
+            'regarding' => $regarding,
+            'contextGuid' => 'lexeme.th'
         );
 
         $commentList = new LexCommentListModel($project);
@@ -52,7 +51,7 @@ class LexCommentCommandsTest extends TestCase
         $commentArray = $commentList->entries[0];
         $this->assertEquals($commentContent, $commentArray['content']);
         $this->assertEquals($regarding, $commentArray['regarding']);
-        $this->assertEquals(0, $commentArray['score']);
+        $this->assertEquals(0, $commentArray['score'] ?? 0);
         $this->assertEquals('open', $commentArray['status']);
     }
 
@@ -74,7 +73,8 @@ class LexCommentCommandsTest extends TestCase
         $data = array(
             'id' => '',
             'content' => 'hi there!',
-            'regarding' => $regarding
+            'regarding' => $regarding,
+            'contextGuid' => 'lexeme.th'
         );
         $commentId = LexCommentCommands::updateComment($project->id->asString(), $userId, $environ->website, $data);
 
@@ -111,7 +111,8 @@ class LexCommentCommandsTest extends TestCase
         $data = array(
             'id' => '',
             'content' => 'hi there!',
-            'regarding' => $regarding
+            'regarding' => $regarding,
+            'contextGuid' => 'lexeme.th'
         );
         $commentId = LexCommentCommands::updateComment($project->id->asString(), $userId, $environ->website, $data);
         $comment = new LexCommentModel($project, $commentId);
@@ -149,7 +150,8 @@ class LexCommentCommandsTest extends TestCase
         $data = array(
             'id' => '',
             'content' => 'hi there!',
-            'regarding' => $regarding
+            'regarding' => $regarding,
+            'contextGuid' => 'lexeme.th'
         );
         $commentId = LexCommentCommands::updateComment($project->id->asString(), $userId, $environ->website, $data);
         $comment = new LexCommentModel($project, $commentId);
@@ -198,7 +200,8 @@ class LexCommentCommandsTest extends TestCase
         $data = array(
             'id' => '',
             'content' => 'hi there!',
-            'regarding' => $regarding
+            'regarding' => $regarding,
+            'contextGuid' => 'lexeme.th'
         );
         $commentId = LexCommentCommands::updateComment($project->id->asString(), $userId, $environ->website, $data);
         $comment = new LexCommentModel($project, $commentId);
@@ -235,7 +238,8 @@ class LexCommentCommandsTest extends TestCase
         $data = array(
             'id' => '',
             'content' => 'hi there!',
-            'regarding' => $regarding
+            'regarding' => $regarding,
+            'contextGuid' => 'lexeme.th'
         );
         $commentId = LexCommentCommands::updateComment($project->id->asString(), $userId, $environ->website, $data);
         $comment = new LexCommentModel($project, $commentId);
@@ -276,7 +280,8 @@ class LexCommentCommandsTest extends TestCase
         $data = array(
             'id' => '',
             'content' => 'hi there!',
-            'regarding' => $regarding
+            'regarding' => $regarding,
+            'contextGuid' => 'lexeme.th'
         );
         $commentId = LexCommentCommands::updateComment($project->id->asString(), $userId, $environ->website, $data);
         $comment = new LexCommentModel($project, $commentId);
@@ -310,7 +315,8 @@ class LexCommentCommandsTest extends TestCase
         $data = array(
             'id' => '',
             'content' => 'hi there!',
-            'regarding' => $regarding
+            'regarding' => $regarding,
+            'contextGuid' => 'lexeme.th'
         );
         $commentId = LexCommentCommands::updateComment($project->id->asString(), $userId, $environ->website, $data);
         $comment = new LexCommentModel($project, $commentId);
@@ -355,7 +361,8 @@ class LexCommentCommandsTest extends TestCase
         $data = array(
             'id' => '',
             'content' => 'hi there!',
-            'regarding' => $regarding
+            'regarding' => $regarding,
+            'contextGuid' => 'lexeme.th'
         );
         $commentId = LexCommentCommands::updateComment($project->id->asString(), $user1Id, $environ->website, $data);
         $comment = new LexCommentModel($project, $commentId);
@@ -386,7 +393,8 @@ class LexCommentCommandsTest extends TestCase
         $data = array(
             'id' => '',
             'content' => 'hi there!',
-            'regarding' => $regarding
+            'regarding' => $regarding,
+            'contextGuid' => 'lexeme.th'
         );
         $commentId = LexCommentCommands::updateComment($project->id->asString(), $user1Id, $environ->website, $data);
         $comment = new LexCommentModel($project, $commentId);
--- a/test/php/model/languageforge/lexicon/command/LexCommentCommandsTest.php
+++ b/test/php/model/languageforge/lexicon/command/LexCommentCommandsTest.php
@@ -124,7 +124,7 @@ class LexCommentCommandsTest extends TestCase
         $this->assertCount(0, $comment->replies);
 
         $replyId = LexCommentCommands::updateReply($project->id->asString(), $userId, $environ->website, $commentId, $replyData);
-        $comment->read($commentId);
+        $comment = new LexCommentModel($project, $commentId);
 
         $reply = $comment->getReply($replyId);
 
@@ -164,7 +164,7 @@ class LexCommentCommandsTest extends TestCase
         LexCommentCommands::updateReply($project->id->asString(), $userId, $environ->website, $commentId, $replyData);
         $replyId = LexCommentCommands::updateReply($project->id->asString(), $userId, $environ->website, $commentId, $replyData);
 
-        $comment->read($commentId);
+        $comment = new LexCommentModel($project, $commentId);
         $reply = $comment->getReply($replyId);
 
         $this->assertEquals($replyData['content'], $reply->content);
@@ -175,7 +175,7 @@ class LexCommentCommandsTest extends TestCase
         );
 
         LexCommentCommands::updateReply($project->id->asString(), $userId, $environ->website, $commentId, $replyData);
-        $comment->read($commentId);
+        $comment = new LexCommentModel($project, $commentId);
         $reply = $comment->getReply($replyId);
 
         $this->assertCount(2, $comment->replies);
@@ -214,7 +214,7 @@ class LexCommentCommandsTest extends TestCase
         LexCommentCommands::deleteComment($project->id->asString(), $userId, $environ->website, $commentId);
 
         $commentList->read();
-        $comment->read($commentId);
+        $comment = new LexCommentModel($project, $commentId);
 
         $this->assertEquals(0, $commentList->count);
         $this->assertTrue($comment->isDeleted);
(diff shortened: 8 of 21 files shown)


megahirt:
Robin, this is excellent work.  As you know, any changes made to MongoMapper should be done with fear and trepidation.  Your modification is bold and will surely result in lots of empty data being deleted from our mongoDB project instances.  I like it :)  Let's do lots of testing with this before shipping to ensure that we don't actually create unintentional data loss for existing projects.

My only corrections are regarding what I perceive to be confusion around `$model = new Model($id)` and `$model->read($id)` are identical. The constructor inside MapperModel calls $this->read($id) so you don't have to do it manually.

Looking forward to getting this into master soon.

---

Reviewed 3 of 7 files at r1.
Review status: 3 of 7 files reviewed at latest revision, 9 unresolved discussions.

---

*[src/Api/Model/Languageforge/Lexicon/Config/LexConfiguration.php, line 33 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-L6Eo6qiuhGY-68JVdYG:-L6Eo6qk0N6dpICilFcp:bi7ibvz) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/src/Api/Model/Languageforge/Lexicon/Config/LexConfiguration.php#L33)):*
> ```PHP
>         $this->tasks = new MapOf(
>             function ($data) {
>                 switch ($data['type'] ?? "") {
> ```

I learned something new today.  Thanks Robin.

---

*[src/Api/Model/Shared/Mapper/MongoMapper.php, line 396 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-L6EoL-t8Rf9C-8B2net:-L6EoL-t8Rf9C-8B2neu:bebnuxx) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/src/Api/Model/Shared/Mapper/MongoMapper.php#L396)):*
> ```PHP
>     }
> 
>     public static function shouldKeepKey(string $key) {
> ```

OK, I agree that this is only necessary because of other assumptions in our code.  Where are the other places that have these assumptions and can they be fixed up in an additional PR so that this is is not necessary?

---

*[test/php/model/languageforge/lexicon/command/LexCommentCommandsTest.php, line 54 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-L6Ep9qWeaxNLAf5QnDv:-L6Ep9qWeaxNLAf5QnDw:bwnysdz) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/test/php/model/languageforge/lexicon/command/LexCommentCommandsTest.php#L54)):*
> ```PHP
>         $this->assertEquals($commentContent, $commentArray['content']);
>         $this->assertEquals($regarding, $commentArray['regarding']);
>         $this->assertEquals(0, $commentArray['score'] ?? 0);
> ```

Is score somehow not being set initially?  The value zero should actually be initialized somewhere, but your modification makes the statement that it's not.  Best to initialize the score to zero somewhere in the actual code if it's not.

---

*[test/php/model/languageforge/lexicon/command/LexEntryCommandsTest.php, line 189 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-L6EpWpm47w3m6gsp8Hg:-L6EpWpm47w3m6gsp8Hh:b-3jjx0i) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/test/php/model/languageforge/lexicon/command/LexEntryCommandsTest.php#L189)):*
> ```PHP
> 
>         $updatedEntry = LexEntryCommands::readEntry($projectId, $entryId);
>         $this->assertEquals('', $updatedEntry['lexeme']['th']['value'] ?? "");
> ```

Why is the `??` necessary here?  If $entry['lexeme']['th'] is supposed to be empty, and the new feature you've added means that this key is unset, then the assertion should be changed to check that the key/value is unset or doesn't exist.  But let's not use null coalescence in a test.

---

*[test/php/model/shared/ProjectModelTest.php, line 238 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-L6EqHuFRguSlOW8yei9:-L6EqHuFRguSlOW8yeiA:b-ma8wei) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/test/php/model/shared/ProjectModelTest.php#L238)):*
> ```PHP
> 
>         // re-read the user
>         $user = new UserModel($userId);
> ```

this is unnecessary here.  You can choose either:
`$user = new UserModel($id)` or
`$user->read($id)` if you already have an UserModel instance.  They are equivalent.  Putting them sequentially is redundant and results in redundant DB calls.

If you find $user->read($id) to be confusing, I'm fine with you replacing it with the constructor, which is how we do it most of the time.  Rarely do we need to re-read the model inside the same method.  Pretty much only in tests do we do this.

---

*[test/php/model/shared/UserModelTest.php, line 378 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-L6EqzGoP56ciY8xs0GW:-L6EqzGoP56ciY8xs0GX:bqz47wj) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/test/php/model/shared/UserModelTest.php#L378)):*
> ```PHP
>         // re-read the project
>         $projectId = $project->id->asString();
>         $project = new \Api\Model\Shared\ProjectModel();
> ```

These lines, while perhaps more clear to the reader, are functionally equivalent to the original line:
`$project->read($project->id->asString());`

---

*[test/php/model/shared/UserUnreadModelTest.php, line 89 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-L6ErLhX26HqY-V3NeO-:-L6ErLhX26HqY-V3NeO0:b-2aox78) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/test/php/model/shared/UserUnreadModelTest.php#L89)):*
> ```PHP
> 
>         // Re-read the unread activity
>         $unreadModel = new UnreadActivityModel($userId1, $project->id->asString());
> ```

this is likely redundant.  See comments above.

---

*[test/php/model/shared/UserUnreadModelTest.php, line 115 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-L6ErOjr9NQPZjuPx4QO:-L6ErOjr9NQPZjuPx4QP:b-2aox78) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/test/php/model/shared/UserUnreadModelTest.php#L115)):*
> ```PHP
> 
>         // Re-read the unread activity
>         $unreadModel = new UnreadActivityModel($userId1, $project->id->asString());
> ```

this is likely redundant.  See comments above.

---

*[test/php/model/shared/UserUnreadModelTest.php, line 171 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-L6ErPlhViPOclZtxnYk:-L6ErPlhViPOclZtxnYl:b-2aox78) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/test/php/model/shared/UserUnreadModelTest.php#L171)):*
> ```PHP
> 
>         // Re-read the unread question model
>         $unreadModel = new UnreadQuestionModel($userId1, $project->id->asString());
> ```

this is likely redundant.  See comments above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-:-L6ErRlTwDIwLsLBm-WL:bmkuspk)*

megahirt:
Reviewed 4 of 7 files at r1.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194)*

rmunn:
I've replied to your comments individually below, but I'll mention (again, since I'm writing this summary last but you'll probably read it first) that due to the way I've changed MongoMapper, `$model->read($id)` is no longer quite identical to `$model = new Model($id)` if the model had old values while a field was deleted in Mongo. The former would leave the deleted field still in the old model and wouldn't remove it (it used to be removed when the mapper wrote null, or the empty string, or an empty array over it). The latter will produce a model *without* the field that was recently deleted from Mongo.

But yes, in many cases after I do `$model = new Model($id)`, the subsequent `$model->read($id)` is completely redundant now, and I'll delete those. I just want to point out (I think I mention it three times between all these comments) that `$model->read($id)` is no longer reliable as a way of *completely* refreshing an existing model instance, and we should transition any code that does this to use `$model = new Model($id)` instead.

---

Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[test/php/model/languageforge/lexicon/command/LexEntryCommandsTest.php, line 189 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-L6EpWpm47w3m6gsp8Hg:-L6GopJI6iVucbuxqOhl:bg0mg07) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/test/php/model/languageforge/lexicon/command/LexEntryCommandsTest.php#L189)):*

Why is the `??` necessary here?  If $entry['lexeme']['th'] is supposed to be empty, and the new feature you've added means that this key is unset, then the assertion should be changed to check that the key/value is unset or doesn't exist.  But let's not use null coalescence in a test.
</blockquote></details>

Here, `$updatedEntry['lexeme']` doesn't (shouldn't) exist. But PHP null-coalescing is similar to C# 6's new `?.` operator, in that it will allow any number of chained key lookups that didn't exist. To perform the same check *without* null-coalescing here, I would have had two options:

1. Do a simple `$this->assertFalse(array_has_key('lexeme', $updatedEntry))`. Except that this is brittle: what if we later add an English field to the lexeme in this test? Then the array would have the key `lexeme`, it just wouldn't have the subkey `lexeme.th` (or `$updatedEntry['lexeme']['th']` in PHP syntax). So the simple test doesn't allow modifying the test, or the test data, very much.

2. Make the test non-brittle by checking every key in turn: 
```
if (!array_has_key('lexeme', $updatedEntry)) {
    $success = true;
} else if (array_has_key('lexeme', $updatedEntry) && !array_has_key('th', $updatedEntry['lexeme'])) {
    $success = true;
} else if (array_has_key('lexeme', $updatedEntry) && array_has_key('th', $updatedEntry['lexeme']) && !array_has_key('value', $updatedEntry['lexeme']['th'])) {
    $success = true;
} else {
    $success = $updatedEntry['lexeme']['th']['value'] === "";
}
$this->assertTrue($success);
```

This avoids using a null-coalescing operator and yet remains non-brittle (this logic will still produce the right result if there's a `['lexeme']['en']['value']` but no `['lexeme']['th']`)... but yikes! Talk about ugly, hard-to-maintain code!

So I looked through the PHP documentation, learned that the null-coalescing operator would do exactly what I wanted (allow me to chain a series of keys, and treat *any* missing key in that chain as a null), and decided that that was the perfect way to write this test in a comprehensible way. In C#, I would have written it as `Assert.Equals("", updatedEntry.Lexeme?.Th?.Value ?? "")` for similar reasons: if the Lexeme or Th keys weren't there, the `?.` operator would just produce a `null` without complaining about the missing middle key.

If you really want me to remove the null-coalescing operator, I can replace it with option 2 above. (I'd rather avoid option 1 for the reason I mentioned earlier: brittle test if we change our test data later). But I'd like to get a confirmation from you that yes, it really is important to not coalesce nulls here. (Why, BTW? I don't quite follow the code-design philosophy that underpins that design choice, at least in this situation where I'm trying to write non-brittle, yet maintainable, test code.)

---

*[test/php/model/shared/ProjectModelTest.php, line 238 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-L6EqHuFRguSlOW8yei9:-L6GmmbOj_ehg3TNmCkM:b8kmjqa) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/test/php/model/shared/ProjectModelTest.php#L238)):*

this is unnecessary here.  You can choose either:
`$user = new UserModel($id)` or
`$user->read($id)` if you already have an UserModel instance.  They are equivalent.  Putting them sequentially is redundant and results in redundant DB calls.

If you find $user->read($id) to be confusing, I'm fine with you replacing it with the constructor, which is how we do it most of the time.  Rarely do we need to re-read the model inside the same method.  Pretty much only in tests do we do this.
</blockquote></details>

Due to the changes I've made to MongoMapper, and how fields with nulls are not written to the database, `$user = new UserModel($id)` and `$user->read($id)` are no longer equivalent: if your `$user` field contains old data (as this one does: the list of "projects I'm a member of") which is no longer contained in the MongoDB, a mere `$user->read($userId)` will not delete the old data. Proof: delete the `$user = new UserModel($userId)` line and leave the `$user->read($userId)` line, and watch the test fail. Because the projects list still exists in the old user model, whereas creating a new user model means it now doesn't have that projects list in it.

OTOH, it's quite true that `$user->read($userId)` has become redundant now in this code. But with MongoMapper only storing fields with actual data, and the implementation of our `->read()` methods (which get data from Mongo and set properties on the object, but don't do anything to existing properties), it's important to realize that if you have an old `$model` instance and you believe some data may have been deleted in the database, then `$model->read($modelId)` isn't going to delete the fields that were deleted in Mongo.

---

*[test/php/model/shared/UserModelTest.php, line 378 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-L6EqzGoP56ciY8xs0GW:-L6GnvwWNoqojrcT8hHC:bl429ws) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/test/php/model/shared/UserModelTest.php#L378)):*

These lines, while perhaps more clear to the reader, are functionally equivalent to the original line:
`$project->read($project->id->asString());`

</blockquote></details>

Except that as I mentioned just above, they're no longer functionally equivalent: `$project->read($project->id->asString())` without the `new ProjectModel()` line will cause the test to fail, because the test is verifying that data was deleted, and that data still lingers in the old model. You have to create a new model from scratch now.

I'm quite willing to consider whether the `->read()` method should be rethought, so that it always flushes all old properties from the instance and gets a completely fresh model from Mongo... but the `new` call here is necessary now. (Though of course, the `->read($projectId)` is redundant now and I could have combined it with the next line to make it `new ProjectModel($projectId)`).

---

*[test/php/model/shared/UserUnreadModelTest.php, line 89 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-L6ErLhX26HqY-V3NeO-:-L6GoTr47uqesaP3mUFD:bi2jcp2) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/test/php/model/shared/UserUnreadModelTest.php#L89)):*

this is likely redundant.  See comments above.
</blockquote></details>

Yes, the `$unreadModel->read()` is indeed redundant with the `new`.

---

*[test/php/model/shared/UserUnreadModelTest.php, line 115 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-L6ErOjr9NQPZjuPx4QO:-L6GojCaiSKbrTDWWq2L:bi2jcp2) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/test/php/model/shared/UserUnreadModelTest.php#L115)):*

this is likely redundant.  See comments above.
</blockquote></details>

Yes, the `$unreadModel->read()` is indeed redundant with the `new`.

---

*[test/php/model/shared/UserUnreadModelTest.php, line 171 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-L6ErPlhViPOclZtxnYk:-L6GokWeLT1ncXJ4yPoO:bi2jcp2) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/test/php/model/shared/UserUnreadModelTest.php#L171)):*

this is likely redundant.  See comments above.
</blockquote></details>

Yes, the `$unreadModel->read()` is indeed redundant with the `new`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-:-L6GskwjNATrdpw-6tyV:bb5tkwk)*

rmunn:
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[src/Api/Model/Shared/Mapper/MongoMapper.php, line 396 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-L6EoL-t8Rf9C-8B2net:-L6GtfHbFs-1Iscw0fAe:b9wjkq6) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/src/Api/Model/Shared/Mapper/MongoMapper.php#L396)):*

OK, I agree that this is only necessary because of other assumptions in our code.  Where are the other places that have these assumptions and can they be fixed up in an additional PR so that this is is not necessary?
</blockquote></details>

I don't remember off the top of my head where each assumption is found, but you can find them easily by removing each of these in turn and seeing what breaks. For example, if you change the logic of this function to not include `$key === "answers"`, there's code somewhere in QuestionCommands that will break because it does a `foreach ($question['answers'] as $answer)` without ever checking whether the `$question` object actually has an `answers` property. Now that empty arrays are removed, it's possible that the `answers` property might not exist on that particular question model.

So it's pretty easy to find out what breaks when you remove each of these keys from this function. I decided not to do that in this PR because its scope was already large enough, but I can pretty easily submit a new PR (on top of this one) to remove those assumptions.

---

*[test/php/model/languageforge/lexicon/command/LexCommentCommandsTest.php, line 54 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-L6Ep9qWeaxNLAf5QnDv:-L6Gu_TolpLQ_QwqcD6s:by1axyd) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/test/php/model/languageforge/lexicon/command/LexCommentCommandsTest.php#L54)):*

Is score somehow not being set initially?  The value zero should actually be initialized somewhere, but your modification makes the statement that it's not.  Best to initialize the score to zero somewhere in the actual code if it's not.
</blockquote></details>

What's happening is that a zero score is being removed from Mongo, because the logic that checks that is PHP's `empty()` function, which considers all falsey values (including the number 0) as empty. I could make an exception for integers, but I decided that it's actually a good idea to have a missing integer value in Mongo be equivalent to 0. But yes, the code, rather than the test, is ultimately the right place to make that change.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194)*

megahirt:
Robin, this is good conversation.  Let's keep working on this.

---

Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[test/php/model/languageforge/lexicon/command/LexCommentCommandsTest.php, line 54 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-L6Ep9qWeaxNLAf5QnDv:-L6HKK01fr6fi-TaWLW7:b3kpodo) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/test/php/model/languageforge/lexicon/command/LexCommentCommandsTest.php#L54)):*

What's happening is that a zero score is being removed from Mongo, because the logic that checks that is PHP's `empty()` function, which considers all falsey values (including the number 0) as empty. I could make an exception for integers, but I decided that it's actually a good idea to have a missing integer value in Mongo be equivalent to 0. But yes, the code, rather than the test, is ultimately the right place to make that change.
</blockquote></details>

This is a problem with empty() in my opinion, and storing a score of zero is perfectly valid and different from a value that is unset.  Could we not use `!isset()` instead of `empty()` ?

---

*[test/php/model/languageforge/lexicon/command/LexEntryCommandsTest.php, line 189 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-L6EpWpm47w3m6gsp8Hg:-L6HKpUeBz839eeR2I0n:bgh70j) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/test/php/model/languageforge/lexicon/command/LexEntryCommandsTest.php#L189)):*
> $this->assertEquals('', $updatedEntry['lexeme']['th']['value'] ?? "");
When I read this I think, "This is a fancy way to test that `$updatedEntry['lexeme']['th']['value']` is unset.  Null coalescence seems like the less clear way to make this assertion.  Instead I would do:
`$this->assertFalse(isset($updatedEntry['lexeme']['th']['value']))`
isset() is the cannonical way in PHP to test if a value is set, even multidimensional array values where the key may or may not exist.  The only "gotcha" is if the value is 'null' in which case isset() returns false.  But this is our case anyway so it's acceptable.

Granted, PHP has some weird language functions like isset() that you may not have known about.  I like the null coalescence operator that you are using, but I think it makes the test less readable, and that's what I'm going for.

---

*[test/php/model/shared/ProjectModelTest.php, line 238 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-L6EqHuFRguSlOW8yei9:-L6HM0FNhZ69_VM3iMUr:b149wh2) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/test/php/model/shared/ProjectModelTest.php#L238)):*

Due to the changes I've made to MongoMapper, and how fields with nulls are not written to the database, `$user = new UserModel($id)` and `$user->read($id)` are no longer equivalent: if your `$user` field contains old data (as this one does: the list of "projects I'm a member of") which is no longer contained in the MongoDB, a mere `$user->read($userId)` will not delete the old data. Proof: delete the `$user = new UserModel($userId)` line and leave the `$user->read($userId)` line, and watch the test fail. Because the projects list still exists in the old user model, whereas creating a new user model means it now doesn't have that projects list in it.

OTOH, it's quite true that `$user->read($userId)` has become redundant now in this code. But with MongoMapper only storing fields with actual data, and the implementation of our `->read()` methods (which get data from Mongo and set properties on the object, but don't do anything to existing properties), it's important to realize that if you have an old `$model` instance and you believe some data may have been deleted in the database, then `$model->read($modelId)` isn't going to delete the fields that were deleted in Mongo.
</blockquote></details>

I actually think that $user->read($id) and $user = new UserModel($id) should remain equivalent, so if the new MongoMapper code makes that no longer true, then we need to modify the MapperModel->read() method to make sure that they are equivalent if at all possible.  I'm willing to accept that it's impossible, if you say it is, but I'd first like you to look into it.

---

*[test/php/model/shared/UserModelTest.php, line 378 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-L6EqzGoP56ciY8xs0GW:-L6HMNHe0mQ1bNsPvGCN:b-h51bw2) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/test/php/model/shared/UserModelTest.php#L378)):*

Except that as I mentioned just above, they're no longer functionally equivalent: `$project->read($project->id->asString())` without the `new ProjectModel()` line will cause the test to fail, because the test is verifying that data was deleted, and that data still lingers in the old model. You have to create a new model from scratch now.

I'm quite willing to consider whether the `->read()` method should be rethought, so that it always flushes all old properties from the instance and gets a completely fresh model from Mongo... but the `new` call here is necessary now. (Though of course, the `->read($projectId)` is redundant now and I could have combined it with the next line to make it `new ProjectModel($projectId)`).
</blockquote></details>

Yes, please look into ->read() doing a refresh from the DB, including clearing out properties that are unset.

---

*[test/php/model/shared/UserUnreadModelTest.php, line 171 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-L6ErPlhViPOclZtxnYk:-L6HMcAv_PFRQYqplXIj:b-soyq8p) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/test/php/model/shared/UserUnreadModelTest.php#L171)):*

Yes, the `$unreadModel->read()` is indeed redundant with the `new`.
</blockquote></details>

ok, so I assume you will remove it with a commit?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-:-L6HMhUdwANe-bPFhyvS:bh7jymj)*

megahirt:
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[src/Api/Model/Shared/Mapper/MongoMapper.php, line 396 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-L6EoL-t8Rf9C-8B2net:-L6HMpQ2TT9DfatUINxQ:b-jl9ky6) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/src/Api/Model/Shared/Mapper/MongoMapper.php#L396)):*

I don't remember off the top of my head where each assumption is found, but you can find them easily by removing each of these in turn and seeing what breaks. For example, if you change the logic of this function to not include `$key === "answers"`, there's code somewhere in QuestionCommands that will break because it does a `foreach ($question['answers'] as $answer)` without ever checking whether the `$question` object actually has an `answers` property. Now that empty arrays are removed, it's possible that the `answers` property might not exist on that particular question model.

So it's pretty easy to find out what breaks when you remove each of these keys from this function. I decided not to do that in this PR because its scope was already large enough, but I can pretty easily submit a new PR (on top of this one) to remove those assumptions.
</blockquote></details>

I would like these to be resolved down the road...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194)*

megahirt:
@rmunn If we are going to make this change (which you already have), and if model->read() is no longer equivalent to new Model(), then model-read() should be made a protected function since it shouldn't be used outside the class.

Also, still waiting for you to address my other comments.

---

Reviewed 14 of 14 files at r2.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/sillsdev/web-languageforge/194#-:-L6jJvnq9lkocewRpExT:bkvdfeu)*

rmunn:
I've made read() a protected function, and discussed the reasons why in detail in one of my comments below.

---

Review status: 16 of 20 files reviewed at latest revision, 6 unresolved discussions.

---

*[test/php/model/languageforge/lexicon/command/LexCommentCommandsTest.php, line 54 at r1](https://beta.reviewable.io/reviews/sillsdev/web-languageforge/194#-L6Ep9qWeaxNLAf5QnDv:-L6ujsYd0kOq8uIOskuE:b672g3b) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/test/php/model/languageforge/lexicon/command/LexCommentCommandsTest.php#L54)):*

This is a problem with empty() in my opinion, and storing a score of zero is perfectly valid and different from a value that is unset.  Could we not use `!isset()` instead of `empty()` ?
</blockquote></details>

Using `empty()` detects empty strings and empty arrays, both of which we want to strip. Using `isset()`, I'd still have to check for empty arrays and strings, which is actually harder than special-casing for integers (and also for Boolean false values, which is something I already do).

Also, while there is a difference in theory between a score that's unset and a score that has been set to 0, I'm actually having a hard time coming up with a situation where that would be meaningful to us in practice. And I suspect that if I add an exception for integers, it will result in unset integer properties being stored as zeroes in the Mongo database anyway (which would then make that distinction moot). But I'll add that special-case exception for integers to ensure that ints set to 0 are not stripped, and we'll see if that prediction is right.

---

*[test/php/model/languageforge/lexicon/command/LexEntryCommandsTest.php, line 189 at r1](https://beta.reviewable.io/reviews/sillsdev/web-languageforge/194#-L6EpWpm47w3m6gsp8Hg:-L6upQIm2FlGome0rZd-:b-896fix) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/test/php/model/languageforge/lexicon/command/LexEntryCommandsTest.php#L189)):*

> $this->assertEquals('', $updatedEntry['lexeme']['th']['value'] ?? "");
When I read this I think, "This is a fancy way to test that `$updatedEntry['lexeme']['th']['value']` is unset.  Null coalescence seems like the less clear way to make this assertion.  Instead I would do:
`$this->assertFalse(isset($updatedEntry['lexeme']['th']['value']))`
isset() is the cannonical way in PHP to test if a value is set, even multidimensional array values where the key may or may not exist.  The only "gotcha" is if the value is 'null' in which case isset() returns false.  But this is our case anyway so it's acceptable.

Granted, PHP has some weird language functions like isset() that you may not have known about.  I like the null coalescence operator that you are using, but I think it makes the test less readable, and that's what I'm going for.
</blockquote></details>

Done.

---

*[test/php/model/shared/ProjectModelTest.php, line 238 at r1](https://beta.reviewable.io/reviews/sillsdev/web-languageforge/194#-L6EqHuFRguSlOW8yei9:-L6uln-Z62BsEG68yFnp:b-8uq065) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/test/php/model/shared/ProjectModelTest.php#L238)):*

I actually think that $user->read($id) and $user = new UserModel($id) should remain equivalent, so if the new MongoMapper code makes that no longer true, then we need to modify the MapperModel->read() method to make sure that they are equivalent if at all possible.  I'm willing to accept that it's impossible, if you say it is, but I'd first like you to look into it.
</blockquote></details>

I've looked into it and found that it's just not possible to make `MapperModel->read()` throw away "stale" properties in any safe, clean, maintanable way. PHP doesn't (AFAICT) distinguish between so-called "dynamic" properties (ones not declared in the class declaration, but set later at runtime: most of our lazy properties are this type) and properties set at class declaration. So solutions like `foreach ($this as $key => $value) { unset $this->$key; }` don't work because they unset things like `$id`, `$entryRef`, `$answers` and so on. We would have to have each derived class implement a `reset()` function on their own, which would basically throw away the benefit that MapperModel gives us (each derived class being able to "magically" have their properties set). So as we discussed at standup today, I'm just going to make read() a protected function and change all our API to call `new FooModel($id)` instead of `$fooModel->read($id)` when we want to re-read a model.

---

*[test/php/model/shared/UserModelTest.php, line 378 at r1](https://beta.reviewable.io/reviews/sillsdev/web-languageforge/194#-L6EqzGoP56ciY8xs0GW:-L6uohDR46Uuy3OTiTxD:b77nykg) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/test/php/model/shared/UserModelTest.php#L378)):*

Yes, please look into ->read() doing a refresh from the DB, including clearing out properties that are unset.
</blockquote></details>

As mentioned in my previous comment, I've looked into it and found that it's not feasible.

---

*[test/php/model/shared/UserUnreadModelTest.php, line 171 at r1](https://beta.reviewable.io/reviews/sillsdev/web-languageforge/194#-L6ErPlhViPOclZtxnYk:-L6uoozGCv4NEnja75nX:b-896fix) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/test/php/model/shared/UserUnreadModelTest.php#L171)):*

ok, so I assume you will remove it with a commit?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/sillsdev/web-languageforge/194#-:-L6uoti5-IXOI77Lw_Uq:b-9800ag)*

megahirt:
:lgtm: Thanks @rmunn for persevering.

---

Reviewed 6 of 6 files at r3.

---

*[test/php/model/shared/ProjectModelTest.php, line 238 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-L6EqHuFRguSlOW8yei9:-L6z5cotVnRBv5fmM_KI:b-4d2za2) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2dd59b76aaebe97cba416cce0b0e6b98e03a9871/test/php/model/shared/ProjectModelTest.php#L238)):*

I've looked into it and found that it's just not possible to make `MapperModel->read()` throw away "stale" properties in any safe, clean, maintanable way. PHP doesn't (AFAICT) distinguish between so-called "dynamic" properties (ones not declared in the class declaration, but set later at runtime: most of our lazy properties are this type) and properties set at class declaration. So solutions like `foreach ($this as $key => $value) { unset $this->$key; }` don't work because they unset things like `$id`, `$entryRef`, `$answers` and so on. We would have to have each derived class implement a `reset()` function on their own, which would basically throw away the benefit that MapperModel gives us (each derived class being able to "magically" have their properties set). So as we discussed at standup today, I'm just going to make read() a protected function and change all our API to call `new FooModel($id)` instead of `$fooModel->read($id)` when we want to re-read a model.
</blockquote></details>

Thank you for looking into it!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/194#-:-L6z5WOtRvLbzDhOG4HM:bg5yn9x)*

PR #196, opened 2018-03-03
Port dc-formattedtext to a plain text input

Nateowami:
This does not actually remove textAngular as the branch name says. Rather, it is no longer used except for multiparagraph. Multiparagraph is always disabled, so textAngular isn't really needed there either, unless there are plans to change that.

Everything appears to work the same as before, and the e2e tests pass.

EDIT: I should have opened this PR from `sillsdev:feature/remove-textangular` rather than `Nateowami:feature/remove-textangular`. The same commits are pushed to both, and maintainers on sillsdev/web-languageforge can push to my fork's branch.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/196)


megahirt:
Thanks Nathaniel.  Can you confirm that the multi-paragraph code still works using textAngular?  If you can figure out a way to do this, either by changing the field config in the DB on a field directly or by importing data as Ira suggested that has a multi-paragraph, then we can merge this.  The other option, which would be even better, would be to implement the multi-paragraph using a textarea.  I don't think it would be too complicated to handle both textarea for multiline and input for single line.  But this might push out your PR by a week or so, I'm not sure how much it would take to finish that.

---

Reviewed 6 of 6 files at r1.

---

*[src/angular-app/languageforge/lexicon/editor/field/dc-formattedtext.js, line 271 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/196#-L6jMjUpdC3uRuttkApQ:-L6jMjUpdC3uRuttkApR:b-7eff5e) ([raw file](https://github.com/sillsdev/web-languageforge/blob/e5eb1ced8c8a956a18f2ca4121908ebcd8cc899f/src/angular-app/languageforge/lexicon/editor/field/dc-formattedtext.js#L271)):*
> ```JavaScript
>       var ctrl = this;
>       ctrl.fte = {};
>       ctrl.disabledMsg = 'This field cannot be edited because it contains metadata that would '
> ```

This is good.  Glad there is a message on disabling the text field.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/196#-:-L6jNMfcR0-_838--ZhL:bdxmbp7)*

Nateowami:
Yes, it works using textAngular (I imported TestLangProj, found a couple issues, and got it working).

A textarea would be nice, and since it doesn't appear that multitext has been used much, the HTML issue won't cause much of a problem (textAngular is set to wrap paragraphs with a <p> tag, which won't work well with a plain textarea. However, if it hasn't been used much, we can just disable those that have HTML in them).

What do you think about changing it to a textarea, either before or after merging? The way it's currently handled is not terribly different. It uses either a textbox or textAngular, and this would use textbox or textarea. It will take some time though.

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/196#-:-L6jTNeEHBh7W_Cay2xH:b3ismx9)*

Nateowami:
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[src/angular-app/languageforge/lexicon/editor/field/dc-formattedtext.js, line 271 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/196#-L6jMjUpdC3uRuttkApQ:-L6jUYqQIUw1xRsixpnR:b-hqelq) ([raw file](https://github.com/sillsdev/web-languageforge/blob/e5eb1ced8c8a956a18f2ca4121908ebcd8cc899f/src/angular-app/languageforge/lexicon/editor/field/dc-formattedtext.js#L271)):*

This is good.  Glad there is a message on disabling the text field.
</blockquote></details>

I just realized that it's only ever displayed on a disabled textbox--not on a disabled multitext using textAngular. That would be a simple change to make.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/196)*

megahirt:
Yes, change it to use a textarea.  The idea behind multi-paragraph is that the UI is an editable multi-line box (text area).  It's fine with me if we don't support line breaks (<p> tag) while editing with textarea.  As long as the text can flow across multiple lines we're good.

---

Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[src/angular-app/languageforge/lexicon/editor/field/dc-formattedtext.js, line 271 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/196#-L6jMjUpdC3uRuttkApQ:-L6jiTEhh0X2cZCuut6q:b-dispc2) ([raw file](https://github.com/sillsdev/web-languageforge/blob/e5eb1ced8c8a956a18f2ca4121908ebcd8cc899f/src/angular-app/languageforge/lexicon/editor/field/dc-formattedtext.js#L271)):*

I just realized that it's only ever displayed on a disabled textbox--not on a disabled multitext using textAngular. That would be a simple change to make.
</blockquote></details>

Yes, let's make it work as intended.  Field is not-editable, so a message is displayed.  Make sure the message is visible for the case it talks about but is NOT displayed when the field is not editable for another reason, e.g. Observer role.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/196#-:-L6jj7uPrB52IKjxvqyi:bpn66v)*

megahirt:
To summarize our conversation today:

- rename dc-formattedtext to dc-text
- defer change to textarea to a separate PR

Nateowami:
All tests passing, rebased on master, and dc-formattedtext has been renamed to dc-text. This should be ready to merge.

---

Review status: 1 of 11 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/196#-:-L7860JT0se2zibYW7AN:bgxlwvp)*

megahirt:
:lgtm:

---

Reviewed 13 of 13 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/196#-:-L7C7-nhtzTzZlP3oC0G:bnfp4nl)*

PR #200, opened 2018-03-05
Apex/comments migration

nigel-wells:
Created migration script for comments that have no contextGuid available so that they now appear in the context they were likely to have been made in.

Trello Card:
https://trello.com/c/yGJuYzVD/1038-migration-script-comments-that-were-created-before-this-new-feature-got-merged-dont-increase-the-comment-counter-next-to-the-fie

Part of the fix also includes changing the `comment.regarding.inputSystem` to use the `tag` as opposed to the `languageName` that was recently fixed in `569d222`

There is also a fix to display a thumbnail of any pictures when viewing comments from the global button in the header:
image

### DIFF
--- a/src/Api/Model/Shared/DbIntegrityHelper.php
+++ b/src/Api/Model/Shared/DbIntegrityHelper.php
@@ -3,6 +3,11 @@
 namespace Api\Model\Shared;
 
 use Api\Library\Shared\Palaso\DbScriptLogger;
+use Api\Model\Languageforge\Lexicon\LexCommentListModel;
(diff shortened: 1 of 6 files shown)


nigel-wells:
Forgot to say (I think?) it can take around 30 seconds for the local storage cache to clear after the update script has run. This confused me to start with until I realized what was going on. Just leave it on the entry view and then refresh the page.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/200#-:-L7S0x4g_fTcJ53kAVxC:b5l64mf)*

PR #201, opened 2018-03-09
Fix highlight background color of typeahead

Nateowami:
The typeahead had a bright yellowish background for the item being hovered over. This made it illegible because the text goes to white on hover. There is already a dark blue background in place on hover. With these lines removed it looks how one would think it would.

Not sure how this became a problem, but before it just looked wrong and was illegible, and now it looks right.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/201)

### DIFF
--- a/src/angular-app/bellows/directive/palaso-ui.scss
+++ b/src/angular-app/bellows/directive/palaso-ui.scss
@@ -27,10 +27,6 @@ div.typeahead ul {
 
   li {
     padding: 0.25rem;
-
-    &:hover {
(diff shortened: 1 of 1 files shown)


megahirt:
:lgtm: Thanks Nathaniel

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/201#-:-L7C7Zy3cx7bVTdZATph:bj6e1jo)*

PR #205, opened 2018-03-19
Feature/unified config

irahopkinson:
This adds a **Configuration > Unified** tab but leaves the **Configuration > Fields** tab and **View Settings** intact so that the E2E still passes. Before we go to **QA** I have a commit on another branch that removes them but will also need some E2E changes first. I'll get on to that next now that the E2E changes are in **master**.

Known issues:

 * When dragging items to reorder in the **field-specific Input Systems**. Reordering normal fields works as expected but in the **field-specific Input Systems** region (these are draggable sub-regions to the normal field draggable region) also works as expected unless you drag outside the region into the normal fields region. That results in a normal field being reordered rather than dragged item from the sub-region.
 * There is currently no way to change **Pictures** > **Caption Hidden If Empty**. Use **Fields** tab instead.
 * The **Fields** tab has collateral damage from the **Unified** tab implementation:

    * **New Custom Field** button can only add items to the entry. No work-around.
    * You can re-order **Input Systems** but changing which items are selected is not saved. Use the **Unified** tab instead.

 * **View Settings** has an existing issue - when changing role Tabs the right-hand panel is not updated. The work-around is to select another field then back to the original field.


megahirt:
Reviewed 21 of 25 files at r1.
Review status: 21 of 25 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/angular-app/languageforge/lexicon/settings/configuration/new-custom-field.html, line 28 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/205#-L80KMu_6yHiuXcVMTQQ:-L80KMu_6yHiuXcVMTQR:bwxjer2) ([raw file](https://github.com/sillsdev/web-languageforge/blob/a0ce65975c1c456e573a316c9f81cee2ce1cc145/src/angular-app/languageforge/lexicon/settings/configuration/new-custom-field.html#L28)):*
> ```HTML
>         </div>
>         <div class="form-group">
>             <label class="col-form-label" for="level">Field Level</label>
> ```

I'm guessing that this was removed because there is a "new custom field" button at every level, so you don't need to pick the level explicitly.  Correct.

---

*[test/app/languageforge/lexicon/configuration/e2e/configFields.spec.ts, line 170 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/205#-L80M6GZBt5MYs8fhKIa:-L80M6G_7_NRBegp6ON8:b-d36tjc) ([raw file](https://github.com/sillsdev/web-languageforge/blob/8b574178c5f51b7ee7bad82c028fa0b139301b92/test/app/languageforge/lexicon/configuration/e2e/configFields.spec.ts#L170)):*
> ```XML
>     });
> 
>     xit('can add a duplicate field name at a different level', () => {
> ```

I'm generally not a fan of committing xit() since it leaves the e2e tests with a PENDING state which is confusing to outsiders.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/205)*

megahirt:
Reviewed 4 of 25 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/205)*

megahirt:
This is a great contribution and I'm looking forward to getting this into master.  Ira, please merge when you are ready.  I understand that there are no e2e tests for the unified config view.  I also understand that you don't want to ship until the old fields tab is gone.  Is that correct?  We may be able to get some help writing e2e tests.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/205#-:-L80N_2AsQDyqMAE1okn:b9hkjsi)*

irahopkinson:

---

*[src/angular-app/languageforge/lexicon/settings/configuration/new-custom-field.html, line 28 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/205#-L80KMu_6yHiuXcVMTQQ:-L80N089-EJyYDyp6Ybz:b-khw0t) ([raw file](https://github.com/sillsdev/web-languageforge/blob/a0ce65975c1c456e573a316c9f81cee2ce1cc145/src/angular-app/languageforge/lexicon/settings/configuration/new-custom-field.html#L28)):*

I'm guessing that this was removed because there is a "new custom field" button at every level, so you don't need to pick the level explicitly.  Correct.
</blockquote></details>

Yes, its also why you can only add a custom entry in the **Fields** tab. I could put it back in and make it work again for the **Fields** tab but there are other things broken there as well.

---

*[test/app/languageforge/lexicon/configuration/e2e/configFields.spec.ts, line 170 at r1](https://reviewable.io:443/reviews/sillsdev/web-languageforge/205#-L80M6GZBt5MYs8fhKIa:-L80Ng26WdsLpkGWPf_D:b-ppv6d3) ([raw file](https://github.com/sillsdev/web-languageforge/blob/8b574178c5f51b7ee7bad82c028fa0b139301b92/test/app/languageforge/lexicon/configuration/e2e/configFields.spec.ts#L170)):*

I'm generally not a fan of committing xit() since it leaves the e2e tests with a PENDING state which is confusing to outsiders.
</blockquote></details>

I agree but in this case its only there until this file is removed once I get the rest of the **Unified** tab E2E sorted.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/205)*

irahopkinson:
Yes.

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/web-languageforge/205#-:-L80OJHtodMjq0Q66Drb:bb77ufg)*

PR #209, opened 2018-03-29
Reordering senses and examples

Nateowami:
Reordering senses and examples in the UI is functional and the e2e tests pass. The PHP test, however, is failing, and there appear to be issues with GUIDs changing due to reordering.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/209)

### DIFF
--- a/src/angular-app/languageforge/lexicon/editor/field/dc-entry.component.html
+++ b/src/angular-app/languageforge/lexicon/editor/field/dc-entry.component.html
@@ -27,6 +27,7 @@
         <!--suppress JSUnusedLocalSymbols -->
         <div data-ng-repeat="sense in $ctrl.model.senses">
             <dc-sense model="sense" config="$ctrl.config.fields.senses" control="$ctrl.control" index="$index"
+                      move="$ctrl.moveSense" num-senses="$ctrl.numSenses"
                       parent-context-guid="$ctrl.contextGuid" remove="$ctrl.deleteSense" ></dc-sense>
         </div>
     </div>
--- a/src/angular-app/languageforge/lexicon/editor/field/dc-entry.component.ts
+++ b/src/angular-app/languageforge/lexicon/editor/field/dc-entry.component.ts
@@ -59,6 +59,17 @@ export class FieldEntryController implements angular.IController {
       }, () => {});
   }
 
+  numSenses = () => this.model.senses.length;
+
+  moveSense = (index: number, distance: number) => {
+    const senses = this.model.senses;
+    const sense = senses[index];
+    const newPosition = index + distance;
+    if (newPosition < 0 || newPosition >= senses.length) throw new Error();
+    senses.splice(index, 1); // remove 1 element starting from index
+    senses.splice(newPosition, 0, sense); // insert sense, overwriting 0 elements
+  }
+
 }
 
 export const FieldEntryComponent: angular.IComponentOptions = {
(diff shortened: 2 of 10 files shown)


megahirt:
This all looks fine to me and I'm very glad that PHP test is in there, otherwise we might not have gotten down to the root cause of this issue!  We'll wait until @rmunn is able to make an appropriate fix to support this option.

---

Reviewed 10 of 10 files at r1.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/sillsdev/web-languageforge/209#-:-L8kl53r88wpf7jVDRla:b-3fu8qp)*

megahirt:
@rmunn Can you please make a short comment on this PR describing the minimal necessary change needed for this feature to be accepted?  The changes to MongoMapper, etc should be made on this branch.  Thank you.

Nateowami:
As it currently stands, changing the order of senses or examples causes their GUIDs to change. The PHP unit test on this branch shows this behavior. MongoMapper needs to be changed to take reordering into account so the GUIDs are not changed. Presumably FieldWorks already allows reordering, so it might be helpful to verify that FieldWorks does the reordering in the way we expect (i.e. by changing the order of the elements of an array without changing the GUIDs).

Reordering currently works from the user's perspective but is not implemented correctly on the back end.

Nateowami:
@megahirt In case you may not have seen it, @rmunn has looked into this and [suggests updating to Mongo 3.6 first.](https://trello.com/c/bwo4rj1J/1154-design-for-rearranging-senses-in-mongo)

megahirt:
Yes, this PR is blocked because we first need Mongo 3.6.

rmunn:
FYI, I have written a Bash script that does some sanity checks, then upgrades to Mongo 3.6. A mongodump/mongorestore is required because Mongo's default storage format has changed (old format was MMAPI, and takes 17 GiB to store the data we have on live; the new format, WiredTiger, takes 500 MiB to store that same data). We plan to run it on developer machines first, then upgrade dev, QA, and live (in that order). We'll upgrade live on a Monday since our backups run over the weekend. I'll send you the script once I've had a chance to make sure that it runs on more than just my machine (e.g., once Chris Hirt has had a chance to run it on his machine).

Nateowami:
@rmunn With Mongo now at 3.6, where are we at with MongoMapper?

megahirt:
@rmunn We've got to bring dev, qa and live up to mongo 3.6 before we merge
this.

Nateowami:
@megahirt I believe upgrading Mongo was in preparation for making changes to MongoMapper that this PR needs. So I think it's a three-step process: upgrade Mongo, @rmunn makes changes to MongoMapper, then merge this PR. I'm not positive, but I don't think the changes have been made to MongoMapper yet.

megahirt:
That sounds right.

rmunn:
I'm about to rebase this PR onto current master. If anyone has the `feature/lexeme-action-menu` branch currently checked out, you'll need to check it out again. If you need the original commits for anything, they'll still be in the Git repo for thirty days, so you can check them out again. GitHub will hide the commit hashes of the original commits after I do the rebase, so here they are for future reference in case anyone needs them:

e87fca5 - Add options to move senses up and down
96f34cd - Add options to move examples up and down
3e0ba68 - Add e2e tests for reordering senses
827b0e2 - Add failing PHP unit test for reordering senses and examples

rmunn:
@irahopkinson - Since @megahirt is unavailable right now, do you want to give this a review?

rmunn:
The TC build fails because the PHP test that @Nateowami added doesn't pass. (Quite correctly, I might add, since this PR relies on PR #333 being merged in first). I've verified that this unit test *does* pass if PR #333 is merged in first, so don't let the failing build stop you from reviewing this PR. Once #333 is merged in, this PR should build successfully and we can then merge it in.

PR #213, opened 2018-04-04
JWT tokens for Android app OAuth

rmunn:
This should be what's needed for the Android app, as far as I can tell.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/213)

### DIFF
--- a/src/Api/Library/Shared/JWTToken.php
+++ b/src/Api/Library/Shared/JWTToken.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace Api\Library\Shared;
+
+use Firebase\JWT\JWT;
+
+class JWTToken
+{
+
+    /**
+     * @param int     $hoursValid Hours Valid
+     * @param string $userId
+     * @param Website $website
+     * @return string - Access token
+     */
+    static function getAccessToken(int $hoursValid, string $userId, Website $website) {
+        $issuedAt = time();
+        $expiration = $issuedAt + ($hoursValid * 3600);
+        $token = array(
+            "iss" => $website->domain,
+            "aud" => $website->domain,
+            "iat" => $issuedAt,
+            "exp" => $expiration,
+            "sub" => (string) $userId
+        );
+        return JWT::encode($token, JWT_KEY);
+    }
+}
--- a/src/Api/Library/Shared/JWTToken.php
+++ b/src/Api/Library/Shared/JWTToken.php
@@ -26,7 +26,13 @@ class JWTToken
         return JWT::encode($token, JWT_KEY, 'HS256');
     }
 
-    static function verifyAccessToken(object $tokenData, Website $website) {
+    static function verifyAccessToken(string $token, Website $website)
+    {
+        $tokenData = JWT::decode($token, JWT_KEY, ['HS256']);
+        return static::verifyAccessTokenData($tokenData, $website);
+    }
+
+    static function verifyAccessTokenData($tokenData, Website $website) {
         // JWT::decode takes care of checking signature for us, so we just need to check that the token is for our site
         if (empty($tokenData->iss) || empty($tokenData->aud)) {
             return false;
@@ -40,7 +46,7 @@ class JWTToken
 
     static function getUsernameFromToken(string $token, Website $website) {
         $tokenData = JWT::decode($token, JWT_KEY, ['HS256']);
-        if (! self::verifyAccessToken($tokenData, $website)) {
+        if (! static::verifyAccessTokenData($tokenData, $website)) {
             return null;
         }
         if (! isset($tokenData->sub)) {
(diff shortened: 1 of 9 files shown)


megahirt:
Robin, this is good work.  This is also more than I was expecting.  And the function that I was asking for doesn't appear to actually do the login.  From what you've written, it looks like you intended the JWT token to be submitted for a silex login as a separate request.  I think what you've written is an asset, and I'm happy to accept it, but I need the oauth function to be written to spec (see comment below).  Thanks!

---

Reviewed 1 of 3 files at r1, 8 of 8 files at r2.

---

*[src/Site/OAuth/OAuthJWTToken.php, line 66 at r2](https://beta.reviewable.io/reviews/sillsdev/web-languageforge/213#-L9JlyHO6PZuFhzzLiGw:-L9JlyHO6PZuFhzzLiGx:b-etgknw) ([raw file](https://github.com/sillsdev/web-languageforge/blob/6da03d08db393e75c4f8e044e0d2b5c4fb303e36/src/Site/OAuth/OAuthJWTToken.php#L66)):*
> ```PHP
>         // return valid JWT token
> 
>         // login
> ```

This function turned out different than I was expecting.  I was expecting that, given an OAuth token we:

1) validate it
2) get the email from the Oauth token, check that we have an user with that email address (or existing oauth login)
3) Login
4) create a JWT session based upon a successful login

This function returns
on success: a valid JWT token
on failure: an error object with an error message stating where it failed

Currently, I don't see that this function does the login at all.  It appears to just encode a JWT token based upon a matching user.  But it doesn't actually do a Silex Login.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/sillsdev/web-languageforge/213#-:-L9JnmZDCyv6KJ6y2k5N:b-usit9k)*

rmunn:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/Site/Listener/JWTListener.php, line 52 at r2](https://beta.reviewable.io/reviews/sillsdev/web-languageforge/213#-L9Nd-yE-nhR3cULXWW2:-L9Nd-yE-nhR3cULXWW3:b-83eu56) ([raw file](https://github.com/sillsdev/web-languageforge/blob/6da03d08db393e75c4f8e044e0d2b5c4fb303e36/src/Site/Listener/JWTListener.php#L52)):*
> ```PHP
>             $token = new JWTSilexToken('anon.', $token, 'site');  // TODO: Figure out how Silex expects us to handle provider keys
>             $authToken = $this->authenticationManager->authenticate($token);  // This should end up being the JWTAuthenticationProvider->authenticate() function.
>             $this->tokenStorage->setToken($authToken);
> ```

This is where the login happens: setting an authenticated token in Silex token storage is how you log in in Silex.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/sillsdev/web-languageforge/213)*

rmunn:
As far as I can tell, I did write this to spec. The spec asked for a "JWTToken representing logged in user session" in the success case, and that's just what this does. If this JWT token is submitted, it will be taken as authentication as if the user is logged in.

I did wonder about the spec, because it seemed to be asking for two different things. One thing was "when OAuth token is submitted, log the user in", but that's what the existing code did; if that was all that's needed, returning the JWT token would be pointless: the user is already logged in, and the session cookie is all that's needed now. (Unless the Android app doesn't allow for cookies...? But I doubt that's the case: *everything* on the Internet uses cookies these days, so the Android HTTP system surely allows for handling cookies). So in the former case, what's the JWT token for?  And in the latter case, where submitting the JWT token is supposed to mean "Hey, I'm logged in and here's the proof"... see my code comment on one specific line: that's where the logging in happens.

---

Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/sillsdev/web-languageforge/213#-:-L9NdObkB4HhoCdDSFwq:b1l4i8f)*

megahirt:
I think we should clear this up in person, but essentially I'm asking for a function that does the steps outlined below.  The purpose of this method is not intuitive, because the Android app that will consume this API will only use the C# REST API, which accepts a JWT token for session state.  The Android app does not use our Silex JSON RPC / (Cookie-based auth) at all.

---

Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/sillsdev/web-languageforge/213#-:-L9U_EUb0Dxv860jLc-G:begt2ry)*

rmunn:
@megahirt I've updated this PR with the "set login token in session" code you asked for.

megahirt:
Thank you @rmunn :lgtm: We'll give this a go.

---

Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/sillsdev/web-languageforge/213#-:-L9m7Eji0RrTQum_LRqX:booiqkj)*

PR #224, opened 2018-04-23
Feature/activity log filters

rmunn:
To address the needs of the frontend code implementing activity log for Language Forge.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/224)

### DIFF
--- a/src/Api/Model/Shared/ActivityModel.php
+++ b/src/Api/Model/Shared/ActivityModel.php
@@ -2,6 +2,7 @@
 
 namespace Api\Model\Shared;
 
+use Api\Library\Shared\Website;
 use Api\Model\Shared\Mapper\Id;
 use Api\Model\Shared\Mapper\IdReference;
 use Api\Model\Shared\Mapper\MapperModel;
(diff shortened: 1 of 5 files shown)


megahirt:
:lgtm:

---

Reviewed 5 of 5 files at r1.

---

*[src/angular-app/bellows/core/api/activity.service.ts, line 14 at r1](https://beta.reviewable.io/reviews/sillsdev/web-languageforge/224#-LAkjWI90tPb8gsZNG3a:-LAkjWI90tPb8gsZNG3b:bs06bmk) ([raw file](https://github.com/sillsdev/web-languageforge/blob/dbc42d29b6d1aec5a32d18dba02f9893571e107d/src/angular-app/bellows/core/api/activity.service.ts#L14)):*
> ```XML
>   constructor(private api: ApiService) {}
> 
>   validActivityTypes(callback?: JsonRpcCallback) {
> ```

This all looks good.  Let me cast a long-term vision for our server-side DTO architecture.  In order to achieve an offline-first first application.  I have determined that we should have one DTO for the entire application.  This DTO will contain everything needed for the application to run.  It can be a synthesis of lots of different data, including settings, activity, user information, etc.  All of this should be stored off-line.  Adding another DTO like this is OK in the short-term but we want to move away from adding additional network calls that our code is dependent on.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/sillsdev/web-languageforge/224#-:-LAkk2BpBQJy5AdLWNfu:bnfp4nl)*

PR #244, opened 2018-05-04
Allow comments to appear directly on the main entry

nigel-wells:
Relates to https://trello.com/c/5nmjsPhK

3 E2E tests are failing - same as `master`

New comments can not be posted directly against the entry - view only. Replying is allowed for any comments that do appear on the entry.

If no comments are set on the entry then the comment bubble is hidden. No point showing it as clicking it would effectively do nothing as you can't comment.

Have tested with a comment on the entry that has no `contextGuid` set. I've not been able to test with a real sync though - hopefully it is fine though. So long as the `contextGuid` is undefined OR an empty string then it will be fine. Otherwise, if the sync wants to be updated, the `contextGuid` can be set to `"entry"`

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/244)

### DIFF
--- a/src/angular-app/bellows/core/offline/lexicon-comments.service.ts
+++ b/src/angular-app/bellows/core/offline/lexicon-comments.service.ts
@@ -81,7 +81,7 @@ export class LexiconCommentService {
           // As the promise runs when its ready the comments can double up if loadEntryComments is run multiple times
           if (this.comments.items.currentEntry.indexOf(comment) === -1) {
             let contextGuid = '';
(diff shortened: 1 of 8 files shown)


megahirt:
:lgtm:

---

Reviewed 8 of 8 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/244#-:-LBdDPOHDsIZ1kAEcLAI:bnfp4nl)*

Issue #155, opened 2017-09-05
Feature request: allow manager to reorder fields in entry and sense


PaulNel:
When adding a meaning, it is most common to add a "gloss" field and then a "definition" field. Sometimes only one or the other may be used.

Suggest that when a new sense is created, the "gloss" and "definition" fields be included by default, with the gloss showing up first. This would reflect the order in FieldWorks.

There is no option to show the "gloss" fields unless all fields are shown.
![lf meaning field](https://user-images.githubusercontent.com/17054082/30041566-a82e7f38-91b0-11e7-8224-19b5034f29a4.png)

rmunn:
We don't yet have a way to configure the order of fields in Language Forge. The way we've written the code, it would be theoretically possible, but we don't yet have a way for the user to configure that. We do, however, have a way to make the Gloss field (or any other field) visible by default without having to click on the Show All Fields button. I've written user documentation on how to do that here:

https://community.software.sil.org/t/how-to-show-more-fields-gloss-citation-form-etc/299

ThomasBlok:
Issue moved to https://jira.sil.org/browse/LF-20

PR #257, opened 2018-05-15
Fix GH-155: Meaning fields need to be in different order

Umasengoden:
* Added code to reorder the fields


megahirt:
GH-155 was originally requested by Paul Nelson to match the default field order in FLEx.  I believe that with the new configuration view, users will be able to re-order fields at will, which makes this a less important change.  However, we should still match what FLEx's default order is unless we have a good reason not to.  So I approve this PR.  @umasengoden Nice work figuring out what needed to be changed.  I appreciate the PR.

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/257#-:-LC_SyAH5WKky9o2BOlt:b-glhxc7)*

Issue #165, opened 2017-09-13
Left side word list scrolls away for no reason


rmlockwood:
1. Go to my Swedish project
2. Search for 'bakom' and click on it to go to the edit entry view
3. Go the bottom of the first meaning and add an example.

A couple seconds after the example is added, the left word list scrolls up so that 'bakom' is no longer in view. Very annoying!

The same thing seems to happen if you simply start typing data in the definition field. 

I'm using Chrome.

rmunn:
Verified on my local installation. The trigger for this to happen is the auto-save kicking in. When the auto-save saves the entry, it tries to scroll the left side word list to make sure that the entry is in view. (This is so that when you create a new entry and fill it its lexeme form, it will be visible in the correct position in the list). But that is not working well with the "infinite scroll" feature of the left word list: the word list only loads a certain number of words at once (I think it's 50), and if the word it's trying to scroll to is too far down the list, it doesn't load enough words to reach the right point.

So if the word you're editing is lower than the top 50, this bug will cause it to move out of view. The solution to the bug is going to be to tweak the "scroll the current word into view" feature so that it knows how to load the correct "chunk" of 50 words in the list view.

ThomasBlok:
Issue moved to https://jira.sil.org/browse/LF-16

PR #259, opened 2018-05-17
Fix: GH #165 Left side word list scrolls away for no reason

Umasengoden:
* Adjusted the code based on the Scrollbar

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/259)

### DIFF
--- a/src/angular-app/languageforge/lexicon/editor/editor.js
+++ b/src/angular-app/languageforge/lexicon/editor/editor.js
@@ -37,12 +37,12 @@ angular.module('lexicon.editor', [
     '$state', '$window', '$interval', '$filter', 'lexLinkService', 'lexUtils', 'lexRightsService',
     'silNoticeService', '$rootScope', '$location', 'lexConfigService', 'lexCommentService',
     'lexEditorDataService', 'lexProjectService', 'lexSendReceive', 'modalService', '$timeout',
-    'activityService', 'applicationHeaderService',
+    'activityService', 'applicationHeaderService','$anchorScroll',
   function ($scope, userService, sessionService, lexService, $q,
             $state, $window, $interval, $filter, linkService, utils, rightsService,
             notice, $rootScope, $location, lexConfig, commentService,
             editorService, lexProjectService, sendReceive, modal, $timeout,
-            activityService, applicationHeaderService) {
+            activityService, applicationHeaderService,$anchorScroll) {
 
     var pristineEntry = {};
     var warnOfUnsavedEditsId;
@@ -404,7 +404,8 @@ angular.module('lexicon.editor', [
 
       $scope.editEntryAndScroll = function editEntryAndScroll(id) {
         $scope.editEntry(id);
-        scrollListToEntry(id, 'middle');
+        var sId = 'entryId_' + id;
+        $anchorScroll(sId);
       };
 
       function setCurrentEntry(entry) {


megahirt:
@Umasengoden We don't have E2E tests running automatically for PRs at this point, so you'll need to run the E2E tests manually and confirm that all tests pass on your machine.  Please state this in your PR summary as well.

Umasengoden:
My comments to Chris

*  With my changes for scroll to the last item of the list issue, run the test cases and all got passed.

*  We did not remove the method  scrollListToEntry in all the places. In this method only (editEntryAndScroll),
 we removed the method scrollListToEntry(id, 'middle')  and added the changes var sId = 'entryId_' + id;
 $anchorScroll(sId); in order to move the scroll to the selected item particulary the last item which could not be shown 
in the middle of the list.

*  I can reproduce the issue -- if I am entering the word which is found on the last item of the list, scroll is not moving
 to the last since we used the method scrollListToEntry(id, 'middle') which does not make sense of bringing the last word 
in the middel. For this issue only we made the changes.

I have sent the document which is having issue information with screenshots.

Umasengoden:
[scroll issue on the list 18052018.docx](https://github.com/sillsdev/web-languageforge/files/2016889/scroll.issue.on.the.list.18052018.docx)
Pl find the document in which the specific issue information is given along with the screenshots.

jasonleenaylor:
@megahirt, @irahopkinson or @Nateowami could you give this a second look after Uma's answers?

Nateowami:
@megahirt @jasonleenaylor 
The bug as reported is reproducible. In fact, on my tests locally and on live, the automatic scrolling is consistently broken (perhaps it has gotten worse since the bug was reported). When you search for an entry it scrolls so that the entry is one or two entries below the scrolling element's viewport.

The scrollListToEntry function is used two other places, both within editor.js.

I can't run the e2e tests on my office machine due to file permissions being messed up (despite using the install script), but I'll try running them my laptop this evening.

Nateowami:
Just ran the e2e tests. There are two failures in the user profile app, presumably completely unrelated to this PR (they were both "document unloaded while waiting for result"). I also got a third failure once, but it didn't fail the second time through, and didn't appear to be related to this change either. So e2e tests are good to go.

megahirt:
@umasengoden I see now that the scrollListToEntry is completely broken, and was probably broken with the new design.  The offset is hardcoded, which is likely not mobile-friendly which is one of the reasons I like using the anchor scroll.  It may be that it's just too difficult to calculate "middle" which is what the 487 offset was designed to do.  In any case, you should be able to change these two offset numbers (see comment below) and make the item be visible in the list again.

A more ambitious refactor would rewrite scrollListToEntry() using $anchorScroll and utilize the location.hash and  (bonus) account for a display offset that puts the highlighted entry somewhere in the middle of the list instead of the top.

I did ask for a test to accompany this PR so will wait for that and also see what you come up with in response.  Thanks for your work on this.

---

---

*[src/angular-app/languageforge/lexicon/editor/editor.js, line 368 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/259#-LDA2SwyBBfrh1gXjrGo:-LDA1KKzEUz2ggJ1-Y-X:b-ghibx) ([raw file](https://github.com/sillsdev/web-languageforge/blob/63dea785d65b6121bb9f3080a32e5191b99b9f7a/src/angular-app/languageforge/lexicon/editor/editor.js#L368)):*
> ```JavaScript
> 
>       function scrollListToEntry(id, position) {
>         var posOffset = (position === 'top') ? 274 : 487;
> ```

@umasengoden You could fiddle with the 274 and 487 hard-coded numbers to adjust the offset to fix the current problem.  See my summary reply for a more info.

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/259#-:-LDA2gvbBng8AuSDQ-mK:b-3pmsoj)*

Umasengoden:

PR #260, opened 2018-05-18
Refactor comments to Angular components

irahopkinson:
This removes API calls and takes data from component inputs instead. Since Angular components have an isolated scopes inputs were added (esp. `dc-comment` which entirely relied on the parent scope).

 - also simplify multi option lists and semantic domain fields

LF E2E is passing.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/260)

### DIFF
--- a/src/angular-app/languageforge/lexicon/core/lexicon-rights.service.ts
+++ b/src/angular-app/languageforge/lexicon/core/lexicon-rights.service.ts
@@ -5,9 +5,9 @@ import {
   RightsFunction, Session,
   SessionService
 } from '../../../bellows/core/session.service';
-import { LexiconSendReceiveService } from './lexicon-send-receive.service';
+import {LexiconSendReceiveService} from './lexicon-send-receive.service';
 
-export type ConditionFunction = () => boolean;
+export type ConditionFunction = (userId?: string) => boolean;
 
 export class Rights {
   readonly canRemoveUsers: ConditionFunction;
@@ -19,18 +19,24 @@ export class Rights {
   readonly canEditEntry: ConditionFunction;
   readonly canDeleteEntry: ConditionFunction;
   readonly canComment: ConditionFunction;
+  readonly canDeleteComment: ConditionFunction;
+  readonly canEditComment: ConditionFunction;
+  readonly canUpdateCommentStatus: ConditionFunction;
 
   constructor(private domain: Domains, private operation: Operations,
               private sendReceive: LexiconSendReceiveService, public session?: Session) {
-    this.canRemoveUsers = this.isPermitted(true, domain.USERS, operation.DELETE);
-    this.canCreateUsers = this.isPermitted(true, domain.USERS, operation.CREATE);
-    this.canEditUsers = this.isPermitted(false, domain.USERS, operation.EDIT);
-    this.canArchiveProject = this.isPermitted(true, domain.PROJECTS, operation.ARCHIVE, true);
-    this.canDeleteProject = this.isPermitted(true, domain.PROJECTS, operation.DELETE, true);
-    this.canEditProject = this.isPermitted(false, domain.PROJECTS, operation.EDIT);
-    this.canEditEntry = this.isPermitted(false, domain.ENTRIES, operation.EDIT);
-    this.canDeleteEntry = this.isPermitted(false, domain.ENTRIES, operation.DELETE);
-    this.canComment = this.isPermitted(false, domain.COMMENTS, operation.CREATE);
+    this.canRemoveUsers = this.isPermitted(true, false, domain.USERS, operation.DELETE);
+    this.canCreateUsers = this.isPermitted(true, false, domain.USERS, operation.CREATE);
+    this.canEditUsers = this.isPermitted(false, false, domain.USERS, operation.EDIT);
+    this.canArchiveProject = this.isPermitted(true, true, domain.PROJECTS, operation.ARCHIVE);
+    this.canDeleteProject = this.isPermitted(true, true, domain.PROJECTS, operation.DELETE);
+    this.canEditProject = this.isPermitted(false, false, domain.PROJECTS, operation.EDIT);
+    this.canEditEntry = this.isPermitted(false, false, domain.ENTRIES, operation.EDIT);
+    this.canDeleteEntry = this.isPermitted(false, false, domain.ENTRIES, operation.DELETE);
+    this.canComment = this.isPermitted(false, false, domain.COMMENTS, operation.CREATE);
+    this.canDeleteComment = this.isPermitted(false, false, domain.COMMENTS, operation.DELETE_OWN, operation.DELETE);
+    this.canEditComment = this.isPermitted(false, false, domain.COMMENTS, operation.EDIT_OWN, false);
+    this.canUpdateCommentStatus = this.isPermitted(false, false, domain.COMMENTS, operation.EDIT);
   }
 
   /**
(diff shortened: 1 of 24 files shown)


rmunn:
:lgtm:

This is definitely the right direction to be moving. I've made some comments on various parts of the code, but nothing looks like it should block the PR; rather, there are a few places where some future work might be good.

---

Reviewed 24 of 24 files at r1.

---

*[src/angular-app/languageforge/lexicon/core/lexicon-rights.service.ts, line 68 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/260#-LD0RczBDWQGUAg4QL-L:-LD0RczBDWQGUAg4QL-M:b-ssfx35) ([raw file](https://github.com/sillsdev/web-languageforge/blob/1e45bb0ba15f3e454dba48590c40b689787d6247/src/angular-app/languageforge/lexicon/core/lexicon-rights.service.ts#L68)):*
> ```XML
>         if (otherUserOperation != null && userId != null && this.session.userId() !== userId) {
>           if (typeof otherUserOperation === 'boolean') {
>             hasRight = otherUserOperation;
> ```

It looks like this code branch is only used in a single case, `canEditComment`. Instead of making `otherUserOperation` a dual-typed parameter just for one case where you always want its value to be `false`, why not make it always be a `RightsFunction` and, for the `canEditComment` case, instead of passing the Boolean value `false`, pass a function that always returns `false`. E.g.:

```
this.canEditComment = this.isPermitted(false, false, domain.COMMENTS, operation.EDIT_OWN, () => false);
```

Then the type signature for `isPermitted` becomes a lot simpler and there's no need for this type check every time it's called.

---

*[src/angular-app/languageforge/lexicon/editor/editor.js, line 718 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/260#-LD0ezJAA9rte7XkKUVE:-LD0ezJAA9rte7XkKUVF:b6zrrk4) ([raw file](https://github.com/sillsdev/web-languageforge/blob/1e45bb0ba15f3e454dba48590c40b689787d6247/src/angular-app/languageforge/lexicon/editor/editor.js#L718)):*
> ```JavaScript
>       };
> 
>       $scope.showActivityFeed = function showActivityFeed() {
> ```

Good catch!

---

*[src/angular-app/languageforge/lexicon/editor/comment/comments-right-panel.component.ts, line 24 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/260#-LD0iTfq06NtjOSD6_u4:-LD0iTfq06NtjOSD6_u5:b-3ko5ra) ([raw file](https://github.com/sillsdev/web-languageforge/blob/1e45bb0ba15f3e454dba48590c40b689787d6247/src/angular-app/languageforge/lexicon/editor/comment/comments-right-panel.component.ts#L24)):*
> ```XML
>     visible: false
>   };
>   commentFilter: any;
> ```

For future work, I'd like to see us create a CommentFilter class so that this is strongly typed, because multiple functions in this module are relying on this object to have a specific shape. Not urgent just yet because all the code that relies on this object is pretty close together, so I don't feel that this should block acceptance of the PR.

---

*[src/angular-app/languageforge/lexicon/editor/comment/comments-right-panel.component.ts, line 36 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/260#-LD0h9APDTXfGf0OVSi5:-LD0h9APDTXfGf0OVSi6:bqhayho) ([raw file](https://github.com/sillsdev/web-languageforge/blob/1e45bb0ba15f3e454dba48590c40b689787d6247/src/angular-app/languageforge/lexicon/editor/comment/comments-right-panel.component.ts#L36)):*
> ```XML
>       status: 'all',
>       contextGuid: '',
>       byText: (comment: LexComment): boolean => {
> ```

It may be worth optimizing this function to check if `this.commentFilter.text` is the empty string, and just returning true immediately if that's the case. Otherwise every time `refreshFilteredComments` is called, every comment will be converted to JSON every time, which is relatively expensive, and is completely unnecessary if the text filter was the empty string. I expect the most common filter to be by status, and when someone changes to filter by status of "Todo" vs. status of "all", it's totally unnecessary to JSONify the comments. Doing less unnecessary work will mean putting less demand on the user's CPU, which might be important if they're loading the page on a smartphone (where more CPU work = more battery consumption).

---

*[src/angular-app/languageforge/lexicon/editor/comment/comments-right-panel.component.ts, line 199 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/260#-LD0kpZ28dxCDLSjsLNM:-LD0kpZ28dxCDLSjsLNN:blq9g9s) ([raw file](https://github.com/sillsdev/web-languageforge/blob/1e45bb0ba15f3e454dba48590c40b689787d6247/src/angular-app/languageforge/lexicon/editor/comment/comments-right-panel.component.ts#L199)):*
> ```XML
>     } else if (this.currentEntryCommentsFiltered.length > 0) {
>       if (this.newComment != null && this.newComment.regarding != null) {
>         label = 'Start a new conversation relating to the ' + this.newComment.regarding.fieldNameForDisplay;
> ```

This string is going to cause issues with Transifex, because each field name will be seen as a separate string in Transifex's translation. (E.g., "relating to the meaning" and "relating to the lexeme" will end up being marked as separate strings for translators to localize; and it gets worse if fieldNameForDisplay is localized, because then English strings are going to leak through: you translated "relating to the meaning", but if the French UI has the "meaning" field as being "sens", then Transifex will see the string "relating to the sens" which nobody has localized).

The simplest solution might be to remove the "relating to the _____" text, but that's not ideal either as we lose features that way (it's no longer as clear what you're commenting on). I don't have a good solution for this issue yet, so I'm mentioning it for future work rather than as something that needs to be solved before the PR is accepted. But it's going to be an issue that we'll need to address.

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/260#-:-LD0oub75LQTWDOTUioM:bcc63yg)*

irahopkinson:
Review status: 23 of 24 files reviewed at latest revision, all discussions resolved.

---

*[src/angular-app/languageforge/lexicon/core/lexicon-rights.service.ts, line 68 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/260#-LD0RczBDWQGUAg4QL-L:-LD128Es1A00Y8aAbbXa:b-y1szji) ([raw file](https://github.com/sillsdev/web-languageforge/blob/1e45bb0ba15f3e454dba48590c40b689787d6247/src/angular-app/languageforge/lexicon/core/lexicon-rights.service.ts#L68)):*

It looks like this code branch is only used in a single case, `canEditComment`. Instead of making `otherUserOperation` a dual-typed parameter just for one case where you always want its value to be `false`, why not make it always be a `RightsFunction` and, for the `canEditComment` case, instead of passing the Boolean value `false`, pass a function that always returns `false`. E.g.:

```
this.canEditComment = this.isPermitted(false, false, domain.COMMENTS, operation.EDIT_OWN, () => false);
```

Then the type signature for `isPermitted` becomes a lot simpler and there's no need for this type check every time it's called.
</blockquote></details>

I don't quite see how that would work. If we pass in `() => false` and remove the type check it would return `this.session.hasProjectRight(domain, () => false)` when it should return `false`.

---

*[src/angular-app/languageforge/lexicon/editor/comment/comments-right-panel.component.ts, line 24 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/260#-LD0iTfq06NtjOSD6_u4:-LD15SkBCQ_a2R_1o0Ho:bb6un1j) ([raw file](https://github.com/sillsdev/web-languageforge/blob/1e45bb0ba15f3e454dba48590c40b689787d6247/src/angular-app/languageforge/lexicon/editor/comment/comments-right-panel.component.ts#L24)):*

For future work, I'd like to see us create a CommentFilter class so that this is strongly typed, because multiple functions in this module are relying on this object to have a specific shape. Not urgent just yet because all the code that relies on this object is pretty close together, so I don't feel that this should block acceptance of the PR.
</blockquote></details>

Done

---

*[src/angular-app/languageforge/lexicon/editor/comment/comments-right-panel.component.ts, line 36 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/260#-LD0h9APDTXfGf0OVSi5:-LD15TqnAPwAK7MM8nBQ:bb6un1j) ([raw file](https://github.com/sillsdev/web-languageforge/blob/1e45bb0ba15f3e454dba48590c40b689787d6247/src/angular-app/languageforge/lexicon/editor/comment/comments-right-panel.component.ts#L36)):*

It may be worth optimizing this function to check if `this.commentFilter.text` is the empty string, and just returning true immediately if that's the case. Otherwise every time `refreshFilteredComments` is called, every comment will be converted to JSON every time, which is relatively expensive, and is completely unnecessary if the text filter was the empty string. I expect the most common filter to be by status, and when someone changes to filter by status of "Todo" vs. status of "all", it's totally unnecessary to JSONify the comments. Doing less unnecessary work will mean putting less demand on the user's CPU, which might be important if they're loading the page on a smartphone (where more CPU work = more battery consumption).
</blockquote></details>

Done

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/260)*

irahopkinson:
Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/260)*

rmunn:
I had forgotten the type of `RightsFunction` when I made that suggestion; if `RightsFunction` was of type `() => boolean` then what I suggested would make sense, but it's not. It's of type `() => number` (defined in session.service.ts) and so the type check is necessary after all. So my suggestion won't work after all given the current shape of our API.

PR #261, opened 2018-05-22
Convert/lex new project ts

irahopkinson:
Convert **Lexicon New Project** and all dependencies to TypeScript.

This is in preparation for refactoring to Angular components and better use of the State pattern.

In **Reviewable** use the **Review each commit separately** feature (top-right) so you don't see deletion and addition of renamed and changed files.

LF E2E is passing.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/261)


rmunn:
:lgtm:

---

Reviewed 10 of 12 files at r1, 29 of 34 files at r2.

---

*[src/angular-app/languageforge/lexicon/new-project/lexicon-new-project.controller.ts, line 1 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/261#-LD58mQ04gZx-ZgBdEqr:-LD58mQ1D6rlpzY-pgOs:b-nl31if) ([raw file](https://github.com/sillsdev/web-languageforge/blob/b5945f6fc34f2907f29501a86390510cf498b4e5/src/angular-app/languageforge/lexicon/new-project/lexicon-new-project.controller.ts#L1)):*
> ```XML
> import * as angular from 'angular';
> ```

I'm commenting on the top line because this comment applies to the whole file.

This refactoring is FAR nicer than the ugliness we had before (which I think I'm responsible for: yikes, that was bad. Steps assigned numbers instead of names, and two *different* "step 2" and "step 3" steps? What was I thinking?).

One concern I had while reading this is that the makeFormValid/Neutral/Invalid functions return promises, but those promises aren't always dealt with. (E.g., sometimes you see `this.makeFormValid();` with nothing being done with the function's return value). I've looked through it carefully and I haven't found a place where this would trigger a bug, but I'm not 100% certain that I've covered all the possibilities. For example, what happens in `gotoNextState()` in the `else` branch that runs the following two lines?

```
this.createProject().then(this.getProject);
this.makeFormNeutral();
```

If `getProject()` fails, it calls `gotoEditor()`, which calls `this.makeFormValid()`. When this is immediately followed by `this.makeFormNeutral()`, what's going to happen? Especially since `gotoEditor()` changes `this.$window.location.href`...

I *think* that this particular combination won't cause anything bad to happen, but I can't prove it. So I'm overall happy with this, but anytime a function returns a promise but that promise is ignored, I get a little bit worried. I think this is something we could look at in the future: whether all promises are actually being used (or, if they're deliberately ignored, whether we should mark that in some way, like `this.makeFormNeutral(); // Ignored return value`).

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/261#-:-LD5EHwHDzQiW8p17Vm7:bnfp4nl)*

PR #262, opened 2018-05-22
Properly delete books during PT S/R

ddaspit:
- update user document order if documents are inserted or deleted
- delete ShareDB document when book deleted
- throw exceptions in `ParatextService`
- update NuGet packages

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/262)

### DIFF
--- a/src/angular-app/bellows/apps/translate/core/translate-send-receive.service.ts
+++ b/src/angular-app/bellows/apps/translate/core/translate-send-receive.service.ts
@@ -115,7 +115,7 @@ export class TranslateSendReceiveService {
     this.translateSendReceiveApiService.getJob(this.job.id).then(job => {
       if (!clone) {
         if (job.percentCompleted > 0) {
-          this.notice.setPercentComplete(job.percentCompleted);
+          this.notice.setPercentComplete(Math.round(job.percentCompleted * 100));
         } else {
           this.notice.cancelProgressBar();
         }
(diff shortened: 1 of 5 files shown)


rmunn:
:lgtm:

---

Reviewed 5 of 5 files at r1.

---

*[src/netcore-api/SIL.XForge.WebApi.Server/Services/ParatextSendReceiveRunner.cs, line 183 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/262#-LD5t41g0I6EcNRtaddJ:-LD5t41g0I6EcNRtaddK:b-vk77iu) ([raw file](https://github.com/sillsdev/web-languageforge/blob/d9115aff4c14dff15097f776a2fbfed717d2be4a/src/netcore-api/SIL.XForge.WebApi.Server/Services/ParatextSendReceiveRunner.cs#L183)):*
> ```Smalltalk
>                                     u => u.SetOnInsert(ds => ds.Name, name)
>                                           .SetOnInsert(ds => ds.BookId, bookId)
>                                           .Set(ds => ds.IsDeleted, false), true);
> ```

Minor note: Alignment like this is why I prefer spaces for indentation. I find this alignment style much nicer than something like the following would have been:

```
u => u
    .SetOnInsert(ds => ds.Name, name)
    .SetOnInsert(ds => ds.BookId, bookId)
    .Set(ds => ds.IsDeleted, false), true);
```

But with tab indentation, that's what you would have been forced to write since you can't count on the tab size. (Yes, tabs-for-indentation-and-spaces-for-alignment would have also worked here, but the only editor I know of that can handle that properly is emacs, which most of us don't use).

So thank for for aligning those Set and SetOnInsert calls the way you did; it made this code significantly easier to read and understand the first time through, which makes a significant (and positive) difference to code quality in the long run.

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/262#-:-LD5vO0cBgaPo0KwrHNm:bnfp4nl)*

irahopkinson:
Reviewed 5 of 5 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/262)*

PR #267, opened 2018-05-25
Convert/lex editor ts

irahopkinson:
This is the first stage of converting the **Lexicon Editor** to TypeScript. It's in preparation for refactoring it to an Angular component. Note there are a few hacks to allow this to work before it becomes a component dealing with items that are no longer on the `$scope` from the parent.

I had a good review of dependencies and removed unused ones. Also unused methods are removed.

In **Reviewable** use the **Review each commit separately** feature (top-right) so you don't see deletion and addition of renamed and changed files. The second commit is the main one that needs review; the others are just rename and re-organisation. I separated them to make review easier.

LF E2E passing.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/267)


rmunn:
:lgtm:

I feel I have a much better grasp of the scope of variables and functions in the editor in the refactored version. Thank you Ira for making this important piece of LF quite a bit clearer to read & understand.

---

---

*[src/angular-app/bellows/core/modal/modal.service.ts, line 48 at r4](https://reviewable.io/reviews/sillsdev/web-languageforge/267#-LD_LgC116NTdQ7TlKY9:-LD_LgC28cm17sEHnbSI:bjeqiye) ([raw file](https://github.com/sillsdev/web-languageforge/blob/6345e3f4ed27b74c54b51bdbe1395e30557f7bc9/src/angular-app/bellows/core/modal/modal.service.ts#L48)):*
> ```XML
>                                                   actionButtonText?: string): angular.IPromise<any> {
>     const options: ModalOptions = {
>       headerText,
> ```

I assume that there's a TypeScript feature that says "If the class you're initializing has a property of the same name as the name of the variable you're using to initialize it, you can skip the property name". E.g., it figures out that the `headerText` variable is intended to initialize the `headerText` property of `ModalOptions` here. But I've been unable so far to find where that's documented; if it's mentioned in https://www.typescriptlang.org/docs/handbook/classes.html then I missed it. Where is the documentation for this handy little feature?

---

*[src/angular-app/languageforge/lexicon/editor/field/field-control.model.ts, line 19 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/267#-LDZmKiIBVx_RapuPYWg:-LDZmKiIBVx_RapuPYWh:b-b6efhh) ([raw file](https://github.com/sillsdev/web-languageforge/blob/46dfdef90588a20735f4a4b582009bce030ac4bc/src/angular-app/languageforge/lexicon/editor/field/field-control.model.ts#L19)):*
> ```XML
>   deleteEntry: (currentEntry: LexEntry) => void;
>   getContextParts: (contextGuid: string) => any;
>   getNewComment?: () => LexComment;
> ```

I was initially worried about this change (making getNewComment optional), since comment-bubble.component.ts calls getNewComment() on the FieldControl without ever checking whether that function is defined. But then I traced the code flow and found that lex-comments-view.component.ts sets the getNewComment and selectFieldForComment properties on the FieldControl that's handed to the component, so the comment bubble component will have a valid function to call. This seems like a bit of an encapsulation violation, but I'm okay with it since it's a temporary situation.

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/267#-:-LD_XRp1DfXZUSzfpJ-4:bheeu1z)*

irahopkinson:

---

*[src/angular-app/bellows/core/modal/modal.service.ts, line 48 at r4](https://reviewable.io/reviews/sillsdev/web-languageforge/267#-LD_LgC116NTdQ7TlKY9:-LDaUX5y8MHCFj-JeSb7:b-7cw137) ([raw file](https://github.com/sillsdev/web-languageforge/blob/6345e3f4ed27b74c54b51bdbe1395e30557f7bc9/src/angular-app/bellows/core/modal/modal.service.ts#L48)):*

I assume that there's a TypeScript feature that says "If the class you're initializing has a property of the same name as the name of the variable you're using to initialize it, you can skip the property name". E.g., it figures out that the `headerText` variable is intended to initialize the `headerText` property of `ModalOptions` here. But I've been unable so far to find where that's documented; if it's mentioned in https://www.typescriptlang.org/docs/handbook/classes.html then I missed it. Where is the documentation for this handy little feature?
</blockquote></details>

Nah, that's a standard JavaScript feature. But I didn't know about it until my linter suggested it.

---

*[src/angular-app/languageforge/lexicon/editor/field/field-control.model.ts, line 19 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/267#-LDZmKiIBVx_RapuPYWg:-LDaUuIA23jpAS21wMD7:b5kngxw) ([raw file](https://github.com/sillsdev/web-languageforge/blob/46dfdef90588a20735f4a4b582009bce030ac4bc/src/angular-app/languageforge/lexicon/editor/field/field-control.model.ts#L19)):*

I was initially worried about this change (making getNewComment optional), since comment-bubble.component.ts calls getNewComment() on the FieldControl without ever checking whether that function is defined. But then I traced the code flow and found that lex-comments-view.component.ts sets the getNewComment and selectFieldForComment properties on the FieldControl that's handed to the component, so the comment bubble component will have a valid function to call. This seems like a bit of an encapsulation violation, but I'm okay with it since it's a temporary situation.
</blockquote></details>

Yes I would like that to be fixed also but that's for another time.

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/267)*

PR #268, opened 2018-05-25
Upgrade node-sass to 4.9.0

Nateowami:
Currently we are using node-sass@4.5.0 on [our fork of node-sass](https://github.com/sillsdev/node-sass). 

Node-sass is a wrapper around libsass, which is written in C. The bindings between Node and libsass have to be recompiled for every API change in Node (which is generally quite frequent). The node-sass installer attempts to download a binary for the current platform, Node version, and node-sass version (i.e. it has to find a binary that matches all three). Lately in xForge it has had to fall back to building the bindings from source. I noticed this because on Uma's machine the build fails. The error message isn't very clear about why, but ideally building from source should not be necessary; it should just download the binaries.

The reason it can't download the binaries is because we are using an older, forked version of node-sass (the latest version is 4.9.0; we're using 4.5.0 on [our fork](https://github.com/sillsdev/node-sass/tree/bug/css-build-fix). The node-sass project supplies binaries for 4.5.0 that work with older versions of Node, but they don't support old versions of node-sass on newer versions of Node. 

The original reason we forked node-sass was due to infinite recursion when watching CSS files. The issue still exists but is being addressed in the upcoming release of node-sass 5.0, which upgrades to either libsass 3.5 or 3.6. See https://github.com/sass/node-sass/issues/2111 and https://github.com/sass/node-sass/pull/2312. I can explain more of the reason for the fork if anyone is interested, but for this particular issue it's mostly not relevant.

What I've done is update our fork with version 4.9.0, reapplied the fixes I had made for 4.5.0, and updated xForge's package.json to point to the new branch on our fork.

Everything appears to be working properly. `npm install` now no longer requires compiling libsass from source, produces 11 times less output lines, and runs nearly three times faster (c. 38s rather than 115s).

E2e tests are more or less passing, with a couple failures that look familiar because they've been failing lately. None appear related to updating node-sass.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/268)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5351,15 +5351,13 @@
       "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/generate-function/-/generate-function-2.0.0.tgz",
       "integrity": "sha1-aFj+fAlpt9TpCTM3ZHrHn2DfvnQ=",
-      "dev": true,
-      "optional": true
+      "dev": true
     },
     "generate-object-property": {
       "version": "1.2.0",
       "resolved": "https://registry.npmjs.org/generate-object-property/-/generate-object-property-1.2.0.tgz",
       "integrity": "sha1-nA4cQDCM6AT0eDYYuTf6iPmdUNA=",
       "dev": true,
-      "optional": true,
       "requires": {
         "is-property": "1.0.2"
       }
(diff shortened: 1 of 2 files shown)


irahopkinson:
:lgtm: All Sf and LF E2E passing for me.

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/268#-:-LD_GeLN-taut0R7f-Nd:b4wmmcq)*

PR #274, opened 2018-05-31
Fix TS unit tests

ddaspit:
- use karma-webpack plugin to package unit tests
- update `karma.conf.js` to use webpack config
- update `webpack.config.js` to support packaging for unit tests
- add gulp tasks for running SF and LF unit tests

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/274)


irahopkinson:
:lgtm: Thank you sooo much for getting these working again.
The watch doesn't work quite like I would expect, i.e. it seems to recompile everything and takes a while. But we can work on that later.

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/274#-:-LDspGuXD5RJIhRVGyE1:b-84m562)*

PR #277, opened 2018-06-04
Fix permissions. Misc.

marksvc:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/277)


Nateowami:
Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---

*[deploy/xenial/Vagrantfile, line 78 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/277#-LEAzE2O3RdShEL-zLEN:-LEAzE2O3RdShEL-zLEO:b-e4o5js) ([raw file](https://github.com/sillsdev/web-languageforge/blob/f7fba6570569423ef6c74ffbe8fdd518e6faf446/deploy/xenial/Vagrantfile#L78)):*
> ```Ruby
>     owner: "vagrant",
>     group: "www-data",
>     mount_options: ["dmode=775,fmode=664"]
> ```

I assume this was the offending line? Sorry; I should have thought to look at the options being set in the Vagrantfile. Now I wonder how we were able to use this Vagrantfile in the past.

---

*[deploy/xenial/Vagrantfile, line 88 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/277#-LEB-Fxh9JQd71OUmBnX:-LEB-Fxh9JQd71OUmBnY:b-6nq84z) ([raw file](https://github.com/sillsdev/web-languageforge/blob/dce2f2f35081dabb4fb29b19e683a250971cec85/deploy/xenial/Vagrantfile#L88)):*
> ```Ruby
>   config.vm.provider :virtualbox do |vb|
>     # Use VBoxManage to customize the VM. For example to change memory:
>     vb.customize ["modifyvm", :id, "--memory", "1024"]
> ```

It's possible we'll have to increase it to 2048, as I have had build issues with xForge running out of RAM on a 1GB box. But I'll wait until that happens to make that change.

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/277)*

Nateowami:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[deploy/xenial/Vagrantfile, line 78 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/277#-LEAzE2O3RdShEL-zLEN:-LEB24JsDk57ll4IrpqG:bgbbu1l) ([raw file](https://github.com/sillsdev/web-languageforge/blob/f7fba6570569423ef6c74ffbe8fdd518e6faf446/deploy/xenial/Vagrantfile#L78)):*
<details><summary><i>Previously, marksvc wrote…</i></summary><blockquote>

I removed the owner and group setting here and let it remount. It still runs the PHP tests as well as a couple other scripts. Maybe it's not important to have files group-owned by www-data? I did not try destroy + up but can test with that.
</blockquote></details>

At least some file have to be group-owned by www-data (those in src/assets, because that's written to by the server). I don't know whether others need to be though. I doubt it. 
Am I correct in thinking the mounting change does not require re-cloning the repo, just swapping out the files and rebooting the box?

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/277)*

marksvc:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[deploy/xenial/Vagrantfile, line 78 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/277#-LEAzE2O3RdShEL-zLEN:-LEBU6qo-1hxW_GOJXFU:b-aokdjy) ([raw file](https://github.com/sillsdev/web-languageforge/blob/f7fba6570569423ef6c74ffbe8fdd518e6faf446/deploy/xenial/Vagrantfile#L78)):*

At least some file have to be group-owned by www-data (those in src/assets, because that's written to by the server). I don't know whether others need to be though. I doubt it. 
Am I correct in thinking the mounting change does not require re-cloning the repo, just swapping out the files and rebooting the box?
</blockquote></details>

Right, you can just halt and up the vagrant. You don't need to move the files around, they will just be mounted again with different options. However, any correct-working on incorrect-working from the original provisioning may be masked by this. 

Maybe the Vagrantfile script (or some ansible script) should add group www-data to src/assets , as a better way to do this, rather than mount the whole thing as www-data group?

---

*[deploy/xenial/Vagrantfile, line 88 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/277#-LEB-Fxh9JQd71OUmBnX:-LEBUb309xDdOhXiL2LY:b35atgg) ([raw file](https://github.com/sillsdev/web-languageforge/blob/dce2f2f35081dabb4fb29b19e683a250971cec85/deploy/xenial/Vagrantfile#L88)):*

It's possible we'll have to increase it to 2048, as I have had build issues with xForge running out of RAM on a 1GB box. But I'll wait until that happens to make that change.
</blockquote></details>

I'll plan to update this commit with that.

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/277)*

Nateowami:
Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---

*[deploy/xenial/Vagrantfile, line 78 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/277#-LEAzE2O3RdShEL-zLEN:-LEBWLYZ1pgCIwb78rDY:b-j6llnj) ([raw file](https://github.com/sillsdev/web-languageforge/blob/f7fba6570569423ef6c74ffbe8fdd518e6faf446/deploy/xenial/Vagrantfile#L78)):*
<details><summary><i>Previously, marksvc wrote…</i></summary><blockquote>

Right, you can just halt and up the vagrant. You don't need to move the files around, they will just be mounted again with different options. However, any correct-working on incorrect-working from the original provisioning may be masked by this. 

Maybe the Vagrantfile script (or some ansible script) should add group www-data to src/assets , as a better way to do this, rather than mount the whole thing as www-data group?
</blockquote></details>

There actually is something in the Ansible script that's supposed to set permissions. It must not be working correctly or we wouldn't be having some of these issues.

---

*[deploy/xenial/Vagrantfile, line 88 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/277#-LEB-Fxh9JQd71OUmBnX:-LEBVvDo0O6OD9QY5J4M:b-il5bg6) ([raw file](https://github.com/sillsdev/web-languageforge/blob/dce2f2f35081dabb4fb29b19e683a250971cec85/deploy/xenial/Vagrantfile#L88)):*
<details><summary><i>Previously, marksvc wrote…</i></summary><blockquote>

I'll plan to update this commit with that.
</blockquote></details>

OK. In my tests today during builds there was as little as 60 MB of free RAM. That's cutting it too close to be future-proof, so probably best to increase it now.

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/277)*

marksvc:
Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---

*[deploy/xenial/Vagrantfile, line 78 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/277#-LEAzE2O3RdShEL-zLEN:-LEBXppQ0mXr2YbyA2bY:bl1ul9g) ([raw file](https://github.com/sillsdev/web-languageforge/blob/f7fba6570569423ef6c74ffbe8fdd518e6faf446/deploy/xenial/Vagrantfile#L78)):*

There actually is something in the Ansible script that's supposed to set permissions. It must not be working correctly or we wouldn't be having some of these issues.
</blockquote></details>

Yes. I was in fact just looking at dependencies.yml and it looks like 'Ensure www-data has permission to folders' is probably supposed to take care of this. So I will not have the vagrant bash do it, and instead we should fix the ansible.

---

*[deploy/xenial/Vagrantfile, line 88 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/277#-LEB-Fxh9JQd71OUmBnX:-LEBY0SW4Qqf5LCfPZoc:b-8l0mb) ([raw file](https://github.com/sillsdev/web-languageforge/blob/dce2f2f35081dabb4fb29b19e683a250971cec85/deploy/xenial/Vagrantfile#L88)):*

OK. In my tests today during builds there was as little as 60 MB of free RAM. That's cutting it too close to be future-proof, so probably best to increase it now.
</blockquote></details>

Yes. Thanks for testing that.

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/277)*

Nateowami:
Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---

*[deploy/xenial/Vagrantfile, line 78 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/277#-LEAzE2O3RdShEL-zLEN:-LEBXG0X5LS4YgrLUDOL:bypkvth) ([raw file](https://github.com/sillsdev/web-languageforge/blob/f7fba6570569423ef6c74ffbe8fdd518e6faf446/deploy/xenial/Vagrantfile#L78)):*
<details><summary><i>Previously, marksvc wrote…</i></summary><blockquote>

Yes. I was in fact just looking at dependencies.yml and it looks like 'Ensure www-data has permission to folders' is probably supposed to take care of this. So I will not have the vagrant bash do it, and instead we should fix the ansible.
</blockquote></details>

Just realized I misunderstood what you were saying. You're right; it does make more sense to just set the owner of particular files/folders rather than mounting with a particular group. It's possible there are issues with that though as it syncs between the two file systems. I'm guessing the files store the ID of the user and group, rather than the name, so it might not be possible to keep them in sync across the two file systems.

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/277)*

marksvc:
Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---

*[deploy/xenial/Vagrantfile, line 78 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/277#-LEAzE2O3RdShEL-zLEN:-LEBeHRLCAHD63YKB9dt:b-gius17) ([raw file](https://github.com/sillsdev/web-languageforge/blob/f7fba6570569423ef6c74ffbe8fdd518e6faf446/deploy/xenial/Vagrantfile#L78)):*

Just realized I misunderstood what you were saying. You're right; it does make more sense to just set the owner of particular files/folders rather than mounting with a particular group. It's possible there are issues with that though as it syncs between the two file systems. I'm guessing the files store the ID of the user and group, rather than the name, so it might not be possible to keep them in sync across the two file systems.
</blockquote></details>

Oops, I think we will have to mount as www-data group after all.

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/277)*

Nateowami:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/277#-:-LEBo9Yc3Sb0pWofBJYQ:bnfp4nl)*

irahopkinson:

---

*[deploy/xenial/Vagrantfile, line 35 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/277#-LEBx_259AFDq1tQg41V:-LEBx_259AFDq1tQg41W:b-3a6zsf) ([raw file](https://github.com/sillsdev/web-languageforge/blob/244d9ca2d074250d17f5122b7ab1cdfe75b48ab2/deploy/xenial/Vagrantfile#L35)):*
> ```
> ln -s web-languageforge ~/src/web-scriptureforge
> ```
We really want to create a relative link to the sibling folder. This new link would not work if we moved the `src` folder.

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/277)*

Nateowami:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[deploy/xenial/Vagrantfile, line 35 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/277#-LEBx_259AFDq1tQg41V:-LEC3lTB7D-9A7AHXOIH:b-9iurvw) ([raw file](https://github.com/sillsdev/web-languageforge/blob/244d9ca2d074250d17f5122b7ab1cdfe75b48ab2/deploy/xenial/Vagrantfile#L35)):*

> ```
> ln -s web-languageforge ~/src/web-scriptureforge
> ```
We really want to create a relative link to the sibling folder. This new link would not work if we moved the `src` folder.
</blockquote></details>

Actually, as it currently stands, the link is relative. The first argument indicates where we are linking to, the second where to put the symlink. It doesn't matter if the second argument is relative or not.
The current functionality is wrong though. When we hit the else case of the if statement above web-languageforge will not resolve to any directory because we will be inside of it. I'll try to push a commit fixing this.

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/277)*

rmunn:
:lgtm: as well. Commit 1a4a9de is not actually needed, as I mention in more detail below.

---

---

*[deploy/xenial/Vagrantfile, line 35 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/277#-LEBx_259AFDq1tQg41V:-LEDgitD1i_Dl7J4cN9h:b-odo66l) ([raw file](https://github.com/sillsdev/web-languageforge/blob/244d9ca2d074250d17f5122b7ab1cdfe75b48ab2/deploy/xenial/Vagrantfile#L35)):*

Actually, as it currently stands, the link is relative. The first argument indicates where we are linking to, the second where to put the symlink. It doesn't matter if the second argument is relative or not.
The current functionality is wrong though. When we hit the else case of the if statement above web-languageforge will not resolve to any directory because we will be inside of it. I'll try to push a commit fixing this.
</blockquote></details>

Nathaniel is correct about the symlink: `ln -s web-languageforge ~/src/web-scriptureforge` would create a symlink named `web-scriptureforge` in the `~/src` directory, which is a relative link pointing to `web-languageforge`. However, he's incorrect about the current functionality being wrong. It doesn't matter whether the target of a symlink exists or not: the `ln -s TARGET SOURCE` command will create a symlink called `SOURCE` pointing to `TARGET` *whether or not TARGET exists*. (Try it yourself and see!) So the change in commit 1a4a9de is not needed (but it doesn't do any harm either).

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/277#-:-LEDgfZIA4UZr2N-q81O:bt1mv5l)*

Nateowami:

---

*[deploy/xenial/Vagrantfile, line 35 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/277#-LEBx_259AFDq1tQg41V:-LEF37rP86DX5q1IMvMA:b-70ug7l) ([raw file](https://github.com/sillsdev/web-languageforge/blob/244d9ca2d074250d17f5122b7ab1cdfe75b48ab2/deploy/xenial/Vagrantfile#L35)):*

Nathaniel is correct about the symlink: `ln -s web-languageforge ~/src/web-scriptureforge` would create a symlink named `web-scriptureforge` in the `~/src` directory, which is a relative link pointing to `web-languageforge`. However, he's incorrect about the current functionality being wrong. It doesn't matter whether the target of a symlink exists or not: the `ln -s TARGET SOURCE` command will create a symlink called `SOURCE` pointing to `TARGET` *whether or not TARGET exists*. (Try it yourself and see!) So the change in commit 1a4a9de is not needed (but it doesn't do any harm either).
</blockquote></details>

Ah, OK. I assumed the target was relative to the CWD, but it's actually relative to the location of the symlink.

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/277)*

irahopkinson:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/277#-:-LEH5uYX0bEF3b4nCIhV:bnfp4nl)*

PR #278, opened 2018-06-05
Make Gulp run Webpack via the API so it will work on Windows

Nateowami:
VirtualBox does not forward file change events on the host to the guest. Watching builds on the Linux side, therefore, does not work.

The current workaround is to run them on the Windows using `node node_modules/.bin/webpack --watch  --env.applicationName=languageforge --colors`. The way the Gulp script currently is written uses the shell and fails on Windows. This changes it to use the Webpack API. Admittedly, some parts of this change are suboptimal. The CLI exposes some options not directly available via the API, but it is possible to recreate the same effect via the API. 

For example, where the CLI lets us use the `-p` flag, to do the same via the CLI, we have to load two plugins explicitly and set NODE_ENV to `'production'`. This is exactly what that flag does to begin with.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/278)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -39,6 +39,12 @@ var del = require('del');
 var data = require('gulp-data');
 var ejs = require('gulp-ejs');
 var dest = require('gulp-dest');
+var webpackConfig = require('./webpack.config.js');
+var webpack = require('webpack');
+
+// -------------------------------------
+//   Global Variables
+// -------------------------------------
 
 // Define release stages that will send errors to bugsnag
 var notifyReleaseStages = "['live', 'qa']";
@@ -48,9 +54,6 @@ var notifyReleaseStages = "['live', 'qa']";
 
 // const secrets_google_api_client_id = require('./secrets/google-api-client-id.json');
 
-// -------------------------------------
-//   Global Variables
-// -------------------------------------
 var srcPatterns = [
   'src/angular-app/**',
   'src/Api/**',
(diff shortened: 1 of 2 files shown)


irahopkinson:
:lgtm:

---

---

*[webpack.config.js, line 129 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/278#-LEH9faI3-w-P_UxRMpT:-LEH9faJ-RmTBJf6Dhdk:b7y9su1) ([raw file](https://github.com/sillsdev/web-languageforge/blob/8774c06538cb68f82d6351d19c88a3208bb71d85/webpack.config.js#L129)):*
> ```
> JSON.stringify('production')
> ```
Wondering why this is necessary?

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/278#-:-LEHA9ozAu9N9KsoBqOv:bnfp4nl)*

irahopkinson:
However, this change is causing the front-end unit test to fail. We only just got them working again so I'm unwilling to let this through while that is broken.

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/278#-:-LEHBowd3wKYH3KNffjN:bd90yb3)*

rmunn:
The reason for the failures on TeamCity is because the Webpack config used during the `build-webpack` step is sticking around in the Gulp process, and so when we run the `test-ts` step that is supposed to use a different Webpack config, it's getting modules from the production config as well as the test config. This results in the `webpack.optimize.CommonsChunkPlugin` plugin running during testing, when it was only *supposed* to run during the build step. (It extracts common functions, including `webpackJsonp`, out of the webpack bundles and stores them in a `manifest.bundle.js` file — but our testing doesn't actually look for `manifest.bundle.js` and so it fails).

Repro steps to verify this yourself:

1. Check out this PR locally (`git fetch origin pull/278/head:pr-278` then `git checkout pr-278`).
1. Edit the `build-and-test` step in `gulpfile.js` and comment out the `test-php` step. (Not required for repro, but saves a bit of time). That means that the `build-and-test` target now runs two steps, `build` and then `test-ts`. (It has other steps as well, but it won't getting to them if the `test-ts` step fails).
1. Run `gulp build-and-test` and notice the Karma tests failing with the error message "Uncaught ReferenceError: webpackJsonp is not defined".
1. Run `gulp build`. Everything succeeds.
1. Run `gulp test-ts`. Everything succeeds.

I'll push a commit that should solve the problem with as few changes as possible.

irahopkinson:
:lgtm: Great sleuthing @rmunn.

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/278#-:-LEJ-drH2bJeHRNMIKuA:b99ep87)*

PR #279, opened 2018-06-06
Integrate Bugsnag to C# part of xforge

ermshiperete:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/279)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1251,12 +1251,25 @@ gulp.task('build-dotnet-host-config', function (cb) {
     jwtKey = 'jwtKey';
   }
 
+  var bugsnagApiKey = process.env.XFORGE_BUGSNAG_API_KEY;
+  if (bugsnagApiKey === undefined) {
+    bugsnagApiKey = 'missing-bugsnag-api-key';
+  }
+
   var params = require('yargs')
     .option('applicationName', {
       demand: true,
       type: 'string' })
+    .option('bugsnagApiKey', {
+      demand: false,
+      default: bugsnagApiKey,
+      type: 'string' })
+    .option('buildNumber', {
+      demand: true,
+      type: 'string' })
     .fail(yargFailure)
     .argv;
+  console.log('version =', params.buildNumber);
 
   var jobDatabase = params.applicationName + '_jobs';
 
@@ -1270,6 +1283,10 @@ gulp.task('build-dotnet-host-config', function (cb) {
     },
     DataAccess: {
       JobDatabase: jobDatabase
+    },
+    Bugsnag: {
+      ApiKey: bugsnagApiKey,
+      Version: params.buildNumber
     }
   }), cb);
 });
@@ -1356,7 +1373,7 @@ gulp.task('build-createWebsiteDefsPhp', function () {
 });
 
 // ------------------------------------------
-// Create website-instances.ts from template
+// Create website-instances.generated-data.ts from template
 // ------------------------------------------
 gulp.task('build-createWebsiteDefsTs', function () {
   return gulp.src('src/angular-app/bellows/core/website-instances.ejs')
--- a/package-lock.json
+++ b/package-lock.json
@@ -586,13 +586,13 @@
       "dev": true
     },
     "are-we-there-yet": {
-      "version": "1.1.4",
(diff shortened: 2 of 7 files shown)


rmunn:
I'd like to see one change (detailed below) before I sign off on this.

---

---

*[src/netcore-api/SIL.XForge.WebApi.Server/ExceptionLogging/BugsnagConfigurationOptions.cs, line 20 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/279#-LENE3Lb5l9VEnVskuNZ:-LENE3Lb5l9VEnVskuN_:b-9rlzno) ([raw file](https://github.com/sillsdev/web-languageforge/blob/a63c2348cafbad031ceaf6c864814091551f62c4/src/netcore-api/SIL.XForge.WebApi.Server/ExceptionLogging/BugsnagConfigurationOptions.cs#L20)):*
> ```Smalltalk
>         public void Configure(Configuration options)
>         {
>             _configurator.Configure(options);
> ```

So the entire interface IConfigureOptions interface is basically a single function. I wish they'd designed it around just passing a delegate around, then we wouldn't need a 23-line file (and yet another class) whose entire purpose is fulfilling a single-function interface. I realize that we have to work with the API we are given, so I'm not saying that we shouldn't write this class. I'm just wishing that more people aimed for simplicity in their API design; passing a delegate around ends up being far simpler than using a single-function interface. Basically, http://blog.ploeh.dk/2014/03/10/solid-the-next-step-is-functional/ is what I'm trying to say.

(Eberhard, this is not a criticism of your code, which looks great. This is a criticism of how the Bugsnag people designed their API).

---

*[src/netcore-api/SIL.XForge.WebApi.Server/ExceptionLogging/BugsnagConfigurator.cs, line 34 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/279#-LENFPjF4qBAPgAbi9oz:-LENFPjF4qBAPgAbi9p-:bpg6510) ([raw file](https://github.com/sillsdev/web-languageforge/blob/a63c2348cafbad031ceaf6c864814091551f62c4/src/netcore-api/SIL.XForge.WebApi.Server/ExceptionLogging/BugsnagConfigurator.cs#L34)):*
> ```Smalltalk
>         {
>             var solutionPath =
>                 Path.GetFullPath(Path.Combine(Path.GetDirectoryName(callerFilePath), "../.."));
> ```

I'm not happy with the assumption being made here that the caller file will *always* be two levels down from the solution file. We might end up moving the BugsnagConfigurationOptions class to a different folder, shallower or deeper in the folder structure, and then this line would break. (And we might not even know why, since there would be no obvious link between moving the BugsnagConfigurationOptions.cs file and suddenly having Bugsnag not being able to find the project root correctly). I'd prefer to see logic (in a separate function) that starts at the caller file path and moves one level up the directory tree at a time, looking for the solution file. That would be more robust against later refactoring that might change the location of the BugsnagConfigurationOptions.cs file.

---

*[src/netcore-api/SIL.XForge.WebApi.Server/ExceptionLogging/BugsnagConfigurator.cs, line 93 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/279#-LENGc5T5Yd7Fm-9Jyc1:-LENGc5T5Yd7Fm-9Jyc2:b9o5yau) ([raw file](https://github.com/sillsdev/web-languageforge/blob/a63c2348cafbad031ceaf6c864814091551f62c4/src/netcore-api/SIL.XForge.WebApi.Server/ExceptionLogging/BugsnagConfigurator.cs#L93)):*
> ```Smalltalk
>         }
> 
>         public static void OnBeforeNotify(Report report)
> ```

Why do we have an event handler that does nothing? Does Bugsnag require you to hook up the BeforeNotify event or else their code won't grab exceptions?

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/279#-:-LENGyKJC-LvRXC-HvFD:bm4ydi5)*

rmunn:
:lgtm:

---

Reviewed 3 of 3 files at r2.

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/279#-:-LEO_3ezBA0R6f7JLLQc:bnfp4nl)*

PR #282, opened 2018-06-07
Upgrade test pausing to work with new debugger in Node 8

Nateowami:
At present it is not possible to pause the tests while running because `browser.pause()` is broken as of Node 8.x. The old debugger has been replaced with one that works with the Chrome devtools.

This adds a script that signals the Node process running the tests to tell it to enable its debugger and start listening. Also adds usage instructions to the README.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/282)

### DIFF
--- a/README.md
+++ b/README.md
@@ -546,6 +546,16 @@ will run the the *lexicon-new-project.e2e-spec.ts* tests on **languageforge**.
 
 To add more verbosity during E2E tests, add a parameter `--verbosity true`
 
+To debug the tests:
+- Place breakpoints in your code (`debugger;` statements).
+- Open Chrome to `chrome://inspect/#devices`.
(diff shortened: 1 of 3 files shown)


irahopkinson:
FYI I like this only I can't get it to work. Will paste error in slack.

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/282#-:-LES8BDq7akuGkVJLoY1:bb5dv0y)*

Nateowami:
After thinking about this a bit more, I think this could be a `--debug` flag to the e2e tests rather than a separate script. I think that would be a better user experience.

PR #287, opened 2018-06-13
Ensure activity log DTO always processes difference data

rmunn:
The `instanceof` logic in `prepareDto` assumes that the project model instance received is always the most specific instance possible (e.g., was created by `ProjectModel::getById`). But there are some code paths (for example, `ActivityListDto::getActivityForUser`) where `new ProjectModel($id)` is called, which creates a base ProjectModel instance instead of a more specific instance like LexProjectModel. The simplest fix is to rely on the `appName` field, rather than the PHP class instance.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/287)

### DIFF
--- a/src/Api/Model/Shared/Dto/ActivityListDto.php
+++ b/src/Api/Model/Shared/Dto/ActivityListDto.php
@@ -243,14 +243,15 @@ class ActivityListDto
         return ((new \DateTime($a['date'])) < (new \DateTime($b['date']))) ? 1 : -1;
     }
 
-    private static function prepareDto(&$dto, $projectModel)
+    private static function prepareDto(&$dto, ProjectModel $projectModel)
     {
         foreach ($dto['entries'] as &$item) {
             $item['content'] = $item['actionContent'];
             $item['type'] = 'project';  // FIXME: Should this always be "project"? Should it sometimes be "entry"? 2018-02 RM
             unset($item['actionContent']);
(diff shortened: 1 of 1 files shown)


rmunn:
Review status: all files reviewed, 1 unresolved discussion (waiting on @irahopkinson)

---

*[src/Api/Model/Shared/Dto/ActivityListDto.php, line 252 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/287#-LErO7gI6ECwgl5jbfDt:-LEsAmGt86rxAs1KODKM:b8vrpqc) ([raw file](https://github.com/sillsdev/web-languageforge/blob/31171fd2cf678d733991ad45d95654b18466a387/src/Api/Model/Shared/Dto/ActivityListDto.php#L252)):*

> ```
> 'lexicon'
> ```
Use `LfProjectModel::LEXICON_APP` here.
</blockquote></details>

*Slaps forehead* I know better than to use magic strings when there's already a constant available. Guess I was writing in too much haste. Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/287)*

irahopkinson:
:lgtm: Thanks for cleaning this bug up before it happens.

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/287#-:-LEsKzF_ChxWcuyptcVZ:b-c0oq5s)*

PR #288, opened 2018-06-14
Fix user update and user profile update

irahopkinson:
- allow update of a single ordinary parameter

The old code probably worked fine in PHP5 because `$params['username']` would have returned `null` rather than throw an exception. The two new tests were failing before the other changes.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/288)

### DIFF
--- a/src/Api/Model/Shared/Command/UserCommands.php
+++ b/src/Api/Model/Shared/Command/UserCommands.php
@@ -62,14 +62,18 @@ class UserCommands
     {
         $user = new UserModel($params['id']);
 
+        $checkUsername = '';
+        $checkEmail = '';
         if (array_key_exists('username', $params)) {
-            $params['username'] = UserCommands::sanitizeInput($params['username']);
+            $checkUsername = UserCommands::sanitizeInput($params['username']);
+            $params['username'] = $checkUsername;
         }
         if (array_key_exists('email', $params)) {
-            $params['email'] = UserCommands::sanitizeInput($params['email']);
+            $checkEmail = UserCommands::sanitizeInput($params['email']);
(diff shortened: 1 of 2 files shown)


rmunn:
:lgtm: overall, though I have one style comment (and question about PHP's handling of "index not found" errors). But it doesn't need to block merging this PR.

---

Review status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @irahopkinson)

---

*[src/Api/Model/Shared/Command/UserCommands.php, line 68 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/288#-LEvxMHN6AEIY-VYwMnY:-LEvxMHN6AEIY-VYwMnZ:b-jofa9s) ([raw file](https://github.com/sillsdev/web-languageforge/blob/a7d9779cd98f985039675a8cfd156f5636c0d002/src/Api/Model/Shared/Command/UserCommands.php#L68)):*
> ```PHP
>         $checkEmail = '';
>         if (array_key_exists('username', $params)) {
>             $checkUsername = UserCommands::sanitizeInput($params['username']);
> ```

If accessing a non-existent array index like `$params['username']` throws an exception, then why does this change work? *Shakes head*. Sometimes I think I'll never understand PHP...

Also, wouldn't it be better to just do something like `$params['username'] = UserCommands::sanitizeInput($params['username'] ?? '');` here?

Same comment applies to everywhere where we use the `$checkFoo` pattern below.

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/288#-:-LEvxtdtB_J2dOT3gLGB:bmj7ofc)*

PR #291, opened 2018-06-15
Comments now have labels in activity log DTO

rmunn:
For "add_lex_comment" and "update_lex_comment" activities, activity log DTO now returns a "fieldLabel" field whose contents look like:

```json
"fieldLabel": { "label": "Translation", "sense": 1, "example": 2 }
```

Or, at a minimum:

```json
"fieldLabel": { "label": "Word" }
```

The "fieldLabel.label" field should always be there, but "fieldLabel.sense" and "fieldLabel.example" are optional (though if "fieldLabel.example" appears, then "fieldLabel.sense" is guaranteed to also be there).

The DTO will also contain a "lexCommentFieldValue" which will contain the value of the field at the time the comment was made; this may be useful for displaying in the activity log.

Feedback requested on naming consistency: should "fieldLabel" be called "fieldLabel", or "lexCommentLabel", or "lexCommentFieldLabel"? The advantage of the "fieldLabel" name is that this is consistent with what's returned in the case of entry updates. The advantage of the "lexCommentLabel" or "lexCommentFieldLabel" names would be more internal consistency within the activity log content for a comment update (or a comment addition). Since I couldn't make up my mind which name to use, I've gone for the shorter name, "fieldLabel", for now.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/291)

### DIFF
--- a/src/Api/Model/Shared/ActivityModel.php
+++ b/src/Api/Model/Shared/ActivityModel.php
@@ -40,6 +40,8 @@ class ActivityModel extends MapperModel
     const COMMENT = 'comment';
     const LEX_COMMENT = 'lexComment';
     const LEX_COMMENT_CONTEXT = 'lexCommentContext';
+    const LEX_COMMENT_LABEL = 'lexCommentLabel';
+    const LEX_COMMENT_FIELD_VALUE = 'lexCommentFieldValue';
     const LEX_COMMENT_STATUS = 'lexCommentStatus';
     const LEX_COMMENT_INCREASE_SCORE = 'lexCommentIncreaseScore';
     const LEX_COMMENT_DECREASE_SCORE = 'lexCommentDecreaseScore';
--- a/src/Api/Model/Shared/Command/ActivityCommands.php
+++ b/src/Api/Model/Shared/Command/ActivityCommands.php
@@ -2,10 +2,13 @@
 
 namespace Api\Model\Shared\Command;
 
+use Api\Library\Shared\Palaso\StringUtil;
 use Api\Model\Languageforge\Lexicon\Command\LexEntryCommands;
 use Api\Model\Languageforge\Lexicon\LexCommentModel;
 use Api\Model\Languageforge\Lexicon\LexCommentReply;
 use Api\Model\Languageforge\Lexicon\LexEntryModel;
+use Api\Model\Languageforge\Lexicon\LexExample;
+use Api\Model\Languageforge\Lexicon\LexSense;
 use Api\Model\Scriptureforge\Sfchecks\AnswerModel;
 use Api\Model\Scriptureforge\Sfchecks\QuestionModel;
 use Api\Model\Scriptureforge\Sfchecks\TextModel;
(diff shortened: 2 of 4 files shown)


nigel-wells:
@rmunn I think it would be ideal if we kept the response from the DTO to always provide the same results. So, even though comments will only ever have a single activity, because we have `changes` used for entry updates I think we should keep it consistent. Otherwise the frontend will need to decipher how to interpret the various types of activities coming in. It would be much simpler if the frontend can just rely on an array being available in the `changes` variable regardless of the type of activity. Easy enough to do? This way when potentially other types of activities make there way through the front end won't have to be adjusted at all as it knows what to do. Doable?

rmunn:
@nigel-wells I've done an initial implementation of this (where comments always have a `changes` array) and it feels wrong to me. The `action` field of the activity log DTO is what we should be using to determine the rest of the contents. I'll write up my thoughts on this tomorrow since it's getting rather late here.

rmunn:
@nigel-wells My thoughts, since I had some time this evening to write them up:

- The `action` field acts like a `type` field, determining how the rest of the fields in the activity entry should be parsed.
- An action of `update_entry` means that there's a `changes` array in the content, alongside the usual userref, entryref, etc. fields.
- An action of `add_lex_comment` or `update_lex_comment` means that there's no `changes` array, but there's a `fieldLabel` entry in the content.
- And so on.

If it would be useful to the frontend, I can make sure that a `changes` array always appears, but is empty. But my first reading of your request was that you were asking for the field label in `add_lex_comment` or `update_lex_comment` to contain something like this:

```
changes: [
    {
        fieldLabel: { label: "Meaning", sense: 1}
    }
]
```

And that feels wrong to me. There shouldn't be a `changes` array unless the activity relates to a change, and adding a comment isn't a change (the way we've been using the term). The `content` object is where things like "Which field was commented on?" belong.

nigel-wells:
@rmunn you're quite right, `changes` is not the right variable to use for this. My line of thinking was more around the front end not having to worry interpreting too much information from the DTO based on the type i.e. each activity returned was a single line for the front end rather than potentially having another array contained in the activity representing multiple lines. 

I have re-written this response a few times now and have convinced myself that what you've done is the better option in the long run. Although I still have my initial concerns I think you're approach will give more flexibility even if it means the front end may need to be tweaked more if a new activity type is introduced.

Let's go for it and I'll see how I get on :)

rmunn:
Unit tests written, and used PhpStorm's code-coverage button (which I hadn't noticed until now) to be sure that I've covered all the code paths. (Except for the "moved field" code path, which I won't be able to cover until I implement the PHP backend changes needed for https://github.com/sillsdev/web-languageforge/pull/209). Also addressed code review comments.

---

Review status: all files reviewed, 4 unresolved discussions (waiting on @irahopkinson)

---

*[src/Api/Model/Shared/Command/ActivityCommands.php, line 312 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/291#-LFFs3ks07_Pu_nmvdkY:-LFM2S_j-IJJNjw0TGpk:b-896fix) ([raw file](https://github.com/sillsdev/web-languageforge/blob/c6b2cde147106946ba69995072f0499efcaa067a/src/Api/Model/Shared/Command/ActivityCommands.php#L312)):*

> ```
> public static function prepareActivityLabel
> ```
Does this need to be public? If not move to below the public methods.
</blockquote></details>

Done.

---

*[src/Api/Model/Shared/Command/ActivityCommands.php, line 322 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/291#-LFFrVeV9JvCsQ7XSVp_:-LFM3o3gCejrDBYMXNSo:b-52hyu3) ([raw file](https://github.com/sillsdev/web-languageforge/blob/c6b2cde147106946ba69995072f0499efcaa067a/src/Api/Model/Shared/Command/ActivityCommands.php#L322)):*

> ```
> substr($part, 0, 6) === 'sense#'
> ```
`StringUtil::startsWith($part, 'sense#')` would be clearer. Here and below.
</blockquote></details>

Done. I looked for a PHP function like `starts_with()` and didn't find one, but I missed that we already had one.

---

*[src/Api/Model/Shared/Command/ActivityCommands.php, line 332 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/291#-LFFswQD7p3aPKvwWM6n:-LFM4M_k4UVD8yO4rH9s:b-896fix) ([raw file](https://github.com/sillsdev/web-languageforge/blob/c6b2cde147106946ba69995072f0499efcaa067a/src/Api/Model/Shared/Command/ActivityCommands.php#L332)):*

> ```
> /** LexSense $sense */
> ```
Did you mean `/** @var LexSense $sense */`? Here and below. Remember the `use` statement. 
</blockquote></details>

Done.

---

*[src/Api/Model/Shared/Dto/ActivityListDto.php, line 281 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/291#-LFFrExf5A4lWf_pgLfK:-LFM4hpv7hnBSJae7FVt:b-896fix) ([raw file](https://github.com/sillsdev/web-languageforge/blob/c6b2cde147106946ba69995072f0499efcaa067a/src/Api/Model/Shared/Dto/ActivityListDto.php#L281)):*

> ```
> substr($part, 0, 6) === 'sense@'
> ```
`StringUtil::startsWith($part, 'sense@')` would be clearer. Here and below.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/291#-:-LFM4l4z7xeUnVD-QakZ:b-ikses8)*

irahopkinson:
BTW Yeah that PhpStorm's code-coverage button is quite handy; I might use it myself.

Everything else is looking good only you have one test failing though: `Failed asserting that two arrays are equal at ActivityListDtoTest.php:669`. Its expecting `['label' => 'Sentence', 'sense' => 2, 'example' => 1]` but gets `['label' => 'Sentence']`.

---

Reviewed 3 of 3 files at r2.

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/291#-:-LFM7uWmBQL1KbElWKny:b-gs9riq)*

rmunn:
That was a bug introduced by copying and pasting: I copied a line looking for `sense#` over a line looking for `sense@`, but the unit test caught it for me.

---

Review status: 3 of 4 files reviewed, all discussions resolved (waiting on @irahopkinson)

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/291#-:-LFMNs6JE-Z98Ge-_AJI:bfvsams)*

irahopkinson:
:lgtm: good work, go the tests!

---

Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/291#-:-LFMdYp97djBsNQqP43q:bv0cnda)*

irahopkinson:
Reviewed 3 of 3 files at r4, 2 of 3 files at r5, 1 of 1 files at r6.

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/291)*

PR #292, opened 2018-06-18
Unsticky footer on mobile

Nateowami:
Currently the xForge footer is sticky. I can understand the purpose of a sticky header, but the footer seems unnecessary (looking at commit history, there's a possibility it may have even been changed by accident). On mobile this is especially problematic, especially when typing causes half the screen to be taken by the keyboard.

Surprisingly, this small change to the CSS breaks the e2e tests. Frequently buttons are hidden behind the sticky header and we have to scroll to the top before clicking them. Tests are passing as well as they do on master.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/292)

### DIFF
--- a/src/Site/views/languageforge/theme/default/sass/_global.scss
+++ b/src/Site/views/languageforge/theme/default/sass/_global.scss
@@ -14,9 +14,15 @@ html {
 
 body {
   background: white;
-  height: 100%;
+  min-height: 100%;
   display: flex;
   flex-direction: column;
+}
+
+@include media-breakpoint-up(lg) {
+  body {
+    height: 100%;
+  }
 }
 
 h1.mobile-title {
--- a/src/Site/views/scriptureforge/theme/default/sass/_global.scss
+++ b/src/Site/views/scriptureforge/theme/default/sass/_global.scss
@@ -14,9 +14,15 @@ html {
 
 body {
   background: white;
-  height: 100%;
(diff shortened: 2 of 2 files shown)


megahirt:
@Nateowami Is there a way to make the footer sticky on desktop but not mobile?  The sticky footer (on desktop) is a move toward a more desktop application-oriented UI.  We'd like to move away from the web app that behaves like a web page (lots of vertical page scrolling) and more toward how a desktop application behaves with a fixed footer.  The mobile experience should behave differently and I think that's what is broken here that you're aiming to fix.  I appreciate that.  The sticky footer came back in in part to facilitate the UI that @ddaspit designed for the translate app.  But that app was not designed or tested for a mobile UI so we've got the sticky footer problem on mobile (and potentially tablet with an on-screen keyboard).

Nateowami:
@megahirt Yes; that would certainly be possible. I'll make that change.

Nateowami:
The footer is now sticky on desktop and unsticky on smaller screens.

megahirt:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/292#-:-LFZy9OF3F3YzmOaiVZG:bnfp4nl)*

PR #297, opened 2018-06-19
Make the Node.js debugger listen when the e2e tests start

Nateowami:
- Adds a debug configuration for debugging the e2e tests
- Moves the signaling of the Node.js process to `rune2e.sh`. The result is that the debugger always listens for connections. Nothing actually happens though unless you attach a debugger.

To use, one only needs to start the e2e tests and then in VSCode run the new debug configuration, after Protractor starts.

I initially wrote this to have `rune2e.sh` accept a `--debug` command line flag. That was ugly and brittle though (the order of arguments had to be just so), so I just made enabling the debugger the default. I don't expect that to cause any problems, but if it does it can just be commented out.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/297)


irahopkinson:
FYI this looks good. Please update the `README.md` to match with instructions above.

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/297#-:-LFPLkxi-CfxUnyA8GoR:bmcq0xp)*

Nateowami:
Done.

---

Review status: 3 of 4 files reviewed, all discussions resolved

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/297#-:-LFSiLcS5sOKMlLwjCXH:b-896fix)*

PR #299, opened 2018-06-21
Apex/activity features

nigel-wells:
Trello Card: https://trello.com/c/8LvnfVGe

Updated the activity messages to provide more insight as to what changed and where it changed particularly around the lexicon editor and the comments.
The activity feed now also updates automatically within the lexicon editor without the need to refresh.

There are some issues that still need to be resolved either in the API and I figured it would be best to create this PR so it can be more easily checked against when trying to resolve them.
Issues to resolve:
- Creating a new sense and/or example, adding text to various fields, and then saving doesn't always track changes made in that activity save.
- Not all comment activities are providing the new `fieldLabel` object. Needs to work on any activity within a comment i.e. todo status change, like, reply, etc...
- Sometimes adding a new value on a new sense/example doesn't provide the new value in the  `newValue` field

I'm also open to feedback on how we should label a change for a field contained within a sense and/or example. Currently I've done, "Meaning x Example x {label}" which looks/feels a bit odd - maybe?

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/299)


rmunn:
My comments:

* Creating a new sense and/or example, adding text to various fields, and then saving doesn't always track changes made in that activity save.
    * This appears to be a failing of the "differences" code that I wrote. When there's a brand-new field, it just says "There's a new field" and doesn't dig into the field to record all the field's contents. When that field is something that contains other fields, like a sense or example, then that's not actually the desired behavior and I'll need to improve what the "differences" code returns.
* Not all comment activities are providing the new fieldLabel object. Needs to work on any activity within a comment i.e. todo status change, like, reply, etc...
    * This should be relatively straightforward, as long as the frontend supplies `contextGuid` and `fieldNameForDisplay` (the latter lives in the `regarding` object) for those actions (status change, like, reply, etc). Those two are required for building `fieldLabel`: the `contextGuid` is used for determining which sense and/or example is the field's parent, and the `fieldNameForDisplay` is passed directly into the label. It's possible that the frontend already supplies those things, or it might need to be adjusted to supply them; I haven't looked yet.
* Sometimes adding a new value on a new sense/example doesn't provide the new value in the newValue field
    * Is this the same scenario as the first bullet point, where the new sense/example has just been created and hasn't been saved yet? Or is this a new sense that has been saved once, but then adding a new value *after it's been saved* is still not providing the newValue field in the activity log?

nigel-wells:
Thanks for looking in to the first 2 @rmunn. The last point only appears to contain the first change in the mongo so I'd say it relates to the saving of them along with the ids not being linked up yet? If I refresh the page, and adjust the same 3 fields, everything saves to mongo fine and then all 3 changes come through the DTO correctly - just tested again now as a sanitiy check.

rmunn:
@nigel-wells https://github.com/sillsdev/web-languageforge/pull/300 should address the first of those points.

rmunn:
PR #316 should address the second point from https://github.com/sillsdev/web-languageforge/pull/299#issue-196313753

nigel-wells:
Ok I've just run E2E tests for SF on `master` and I get the same amount of fails (191). Anyone else getting these as well? I've made sure to run `refreshDeps.sh` before I run the tests as well

PR #301, opened 2018-06-21
Add Mongo apt key to Ansible setup and add JRE

Nateowami:
Previously install scripts have had to add
``` shell
sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 2930ADAE8CAF5059EE73BB4B58712A2291FA4AD5
```
in order to have Mongo installed. Adding the apt key is now done in Ansible.

Additionally, I discovered today that we also need Java in order to run WebDriver. Hence the installation of default-jre-headless. I'm not very aware of the package situation in the Java world, so it's possible there's a different package that's more appropriate to install and I wouldn't know. On xenial this package depends on openjdk-8-jre-headless.


jasonleenaylor:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/301)*

jasonleenaylor:
'java-runtime' might be safer if no specific version is required and it wouldn't add any extra package downloads to systems which already had a java runtime installed.

Nateowami:
@jasonleenaylor I like that idea. But what about java-runtime-headless? It's also a virtual package, and it defaults to default-jre-headless instead of default-jre (default-jre depends on default-jre-headless anyway).

Nateowami:
@jasonleenaylor I'm not very familiar with virtual packages, but after a bit more research it looks like there's not a straightforward way to ensure a virtual package is installed. If my research is correct, the main use case would be to specify the dependencies of a package e.g. `default-jre-headless | java-runtime-headless`. Then the first package would be installed, unless the second package (a virtual package) is already satisfied. We're not building a package though, and I can't find anything specifying how to do likewise when installing packages directly (i.e. not as dependencies of a package).

PR #302, opened 2018-06-22
interface language picker and lazy-load other language semantic domains

irahopkinson:
- language picker in top-right navbar (styled to fit site). Changes semantic domain language (if available).
- **Project Settings > Default Interface Language** is connected. If a user has not yet picked a language the project default language will be used for interface and semantic domains (if available).
- semantic domain data is lazy-loaded on language change otherwise only one is loaded.

LF E2E passing.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/302)


rmunn:
:lgtm: - nice work! Testing on languageforge.local I notice that the Transifex translations for French, etc., aren't connected to the dropdown, but that's why we need to test on QA. Everything looks good from a code standpoint.

---

Reviewed 6 of 6 files at r1, 21 of 21 files at r2, 8 of 8 files at r3, 12 of 12 files at r4.

---

*[src/angular-app/languageforge/core/semantic-domains/semantic-domains.service.ts, line 13 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/302#-LFapNMU0jN1Bk97qxrr:-LFapNMU0jN1Bk97qxrs:bdf17d7) ([raw file](https://github.com/sillsdev/web-languageforge/blob/f728da83bd8c7a1a39fb5bb155a9e974a7a9973b/src/angular-app/languageforge/core/semantic-domains/semantic-domains.service.ts#L13)):*
> ```XML
> }
> 
> export interface SemanticDomains { [key: string]: SemanticDomain; }
> ```

I hadn't seen this syntax before, and I had to read the Typescript docs to understand it. Basically, this means that the SemanticDomains interface can have any number of properties named anything (as long as the names are strings), and each of those properties will be a SemanticDomain. They can then be referenced as either `object['foo']` or `object.foo`. (Though in this particular case where semantic domain keys are things like "1.2.3.4", the `object.foo` notation would not be possible.)

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/302#-:-LFax2xe9jtrRk_WrhYQ:bfunrqh)*

PR #303, opened 2018-06-22
Add Vagrant GUI install method and update local Linux setup instructions

Nateowami:
I've created a Vagrant install method where everything is inside the box, including development tools. I initially thought this would be too slow, but Mark does all his development in VirtualBox and says it is not a problem. It's not ideal, but it's likely faster than using VirtualBox shared folders. At the very least it avoids the numerous problems with having the source code on the host and the application in the box.

The Vagrant GUI install method downloads a box with xForge already installed. This box is generated using `deploy/virtualbox_xenial_gui_creation/Vagrantfile` and should be updated whenever our installation method changes. Currently it's hosted on my Vagrant Cloud account at [NathanielPaulus/xenial64-xforge](https://app.vagrantup.com/NathanielPaulus/boxes/xenial64-xforge). I asked Jason about it and was told for now just host it there and it can be changed in the future if necessary (it would be a simple matter up building the box, uploading it elsewhere, and updating `deploy/vagrant_xenial_gui/Vagrantfile` to point to the new location). So far I am the only person to have tried it out.

The Ansible setup has been changed to install the npm package instead of the nodejs package. This is because npm depends on nodejs, but nodejs does not depend on npm (I believe npm used to be included in the nodejs package). What we really need to install is npm, because we intend to use it to install n, which in turn will install Node.js.

I've also updated the local Linux setup instructions.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/303)

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,35 +170,33 @@ Proceed to [Language Forge Configuration File](#LFConfig) and follow the rest of
 
 ### Local Linux Development Setup <a id="LocalSetup"></a> ###
 
-Start with the Ansible-assisted setup [described here](https://github.com/sillsdev/ops-devbox) to install and configure a basic development environment.
+Start by installing Git and Ansible:
+``` shell
+sudo add-apt-repository ppa:ansible/ansible
+sudo apt-get update
+sudo apt-get install -y git ansible
+```
 
-#### Installation and Deployment ####
-
-After creating your Ansible-assisted setup, clone this repository from your *home* folder...
-
-```` bash
-mkdir src
-cd src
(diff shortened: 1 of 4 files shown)


irahopkinson:
:lgtm:

---

Reviewed 3 of 3 files at r1, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/303#-:-LFopHpI97-_AYcY-r6w:bnfp4nl)*

PR #308, opened 2018-06-26
Retaining new empty entry while refresh in screen

DineshKumarGovindarajan:
In Editor Entry Screen, When New Entry is not saved, Data is retaining while refreshing the browser.

Restricted the duplicate new entry.

Data is retaining in local storage, once it is saved local storage will be cleared.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/308)

### DIFF
--- a/src/angular-app/languageforge/lexicon/editor/editor-entry.view.html
+++ b/src/angular-app/languageforge/lexicon/editor/editor-entry.view.html
@@ -29,7 +29,7 @@
                                                search="$ctrl.typeahead.searchEntries" select="$ctrl.typeahead.searchSelect">
                                     <ul data-ng-if="$ctrl.typeahead.searchResults.length > 0" class="list-group">
                                         <li data-typeahead-item="e" class="typeahead-item list-group-item"
-                                            data-ng-repeat="e in $ctrl.typeahead.searchResults | limitTo: $ctrl.typeahead.limit">
+                                            data-ng-repeat="e in $ctrl.typeahead.searchResults track by $index | limitTo: $ctrl.typeahead.limit">
                                             <div class="listItemPrimary" data-ng-bind-html="$ctrl.getPrimaryListItemForDisplay($ctrl.lecConfig, e)"></div>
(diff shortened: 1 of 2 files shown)


irahopkinson:
BTW this PR doesn't state if the LF E2E tests are passing. Are they?
There is also no tests for this feature.
Explain your reasons for using `localStorage` to implement this feature?

---

Review status: 0 of 2 files reviewed, all discussions resolved

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/308#-:-LFy5VZY8TTT7bneb0eK:b-u30pkr)*

DineshKumarGovindarajan:
```
I have done some changes and now all test cases got passed.
```
When we are creating a new entry and while word field is empty and some other fields are filled, autosave has been restricted in this scenario.
```
Without saving if we are refreshing the screen data should be retained on browser side (LocalStorage has been used).
```

megahirt:
The use of localstorage is interesting to solve this particular problem, but I'm not in favor of the way it's written since it adds considerable complexity to the methods.  We already use localforage to persist data, so lets use that as the storage mechanism.  Also, the getitem/saveitem/removeitem should be moved to separate functions so it's clear what is being done in the onchange method.

Does this save data instantly on every keystroke?

As a general comment though, I think using localstorage caching is the right approach for persisting user data that hasn't made it to the server.  Currently we aren't doing this which is why this PR exists.

But the future direction we need to head is offline, which means that all entries created offline need to be persisted in localforage and then uploaded to the server when a connection is available.

Not for this PR, but future work build on this workflow as follows.  All new entries, delete entries, and edits to existing entries need to be persisted in localforage *first*.  These would likely have a dirty property that marks them as needing to be uploaded.  The "save timer" needs to become a full-blown "process offline edits".  This could be done every 10 seconds when online, and also on specific triggers (like switching views).

megahirt:
The PR should be split into two PRs: one that addresses the security issue, and the other that addresses the title.  It looks like @DineshKumarGovindarajan already closed this PR so that's probably what he's doing.

PR #311, opened 2018-06-26
Fix blank entry in SF reports select tag

Nateowami:
This is somewhat unintuitive. The original code looks more "correct." However, it appears that AngularJS maps null/undefined in the model to an empty string value in the select tag. I'm guessing this is because it needs a concept of the selection value being null. 

This isn't clearly documented, but the examples in [the documentation](https://docs.angularjs.org/api/ng/directive/ngOptions) don't define the property in the scope, making it undefined by default.

Before | After
---------|-------
sf_select_before | sf_select_after

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/311)

### DIFF
--- a/src/angular-app/bellows/directive/pui-run-report.js
+++ b/src/angular-app/bellows/directive/pui-run-report.js
@@ -10,7 +10,7 @@ angular.module('palaso.ui.runReport', ['coreModule'])
         function ($scope, projectService) {
           $scope.report = {
             output: '',
-            currentId: ''
+            currentId: null
           };
           $scope.reportOutput = '';
 


megahirt:
I'm not sure if this has anything to do with it, but it's possible this has
been broken since we upgraded to AngularJS 1.6 from 1.5  See
https://docs.angularjs.org/guide/migration#-select-

Thanks for the PR.

irahopkinson:
:lgtm:
Yes its broken all over the place since AngularJS 1.6 which expects `null` for `<option value="">`. AngularJS v1.5 and below used the empty string. I've fix some but plenty to go in both LF and SF and they are not all as simple as this as the data may be stored in Mongo as empty string instead of null.

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/311#-:-LFysZ5U8X10G0xzuMY0:b-sup0yf)*

megahirt:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/311#-:-LFyxB5JD_5qW8DRr2QK:bnfp4nl)*

PR #312, opened 2018-06-26
Fix imageElement is null error

Nateowami:
LF & SF used to both have two avatars on the page, with IDs mobileSmallAvatarURL and smallAvatarURL. A recent change to LF makes it use the same image element for both mobile and desktop, so one of the elements was removed, hence the error "imageElement is null".

Unfortunately the way the e2e tests run this error is masked, likely because console errors are not check until the end of an individual test, and by that time the page has been unloaded.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/312)

### DIFF
--- a/src/angular-app/bellows/apps/userprofile/user-profile-app.component.ts
+++ b/src/angular-app/bellows/apps/userprofile/user-profile-app.component.ts
@@ -212,7 +212,7 @@ export class UserProfileAppController implements angular.IController {
           const newAvatarUrl = this.getAvatarUrl(this.user.avatar_ref);
           ['mobileSmallAvatarURL', 'smallAvatarURL'].forEach(id => {
             const imageElement = this.$window.document.getElementById(id) as HTMLImageElement;
-            imageElement.src = newAvatarUrl;
+            if (imageElement) imageElement.src = newAvatarUrl;
           });
         }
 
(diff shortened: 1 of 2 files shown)


megahirt:
Nice find, Nathaniel.  I'm guessing that one way to tighten up our E2E
tests is to ensure that no single it() block accesses more than one view.
Is that correct?

Nateowami:
Investigating more, I think my analysis was incorrect. We just don't have tests for this. I'll work on adding one.

Nateowami:
Turns out I was right the first time. The error was being masked, just not where I thought it was. I have updated the tests so they fail, then fixed the issue that makes them fail.

irahopkinson:
:lgtm:
Well done for persisting in diagnosing the actual problem and adding the regression test (esp when the fix is so easy). Excellent work.

---

Reviewed 1 of 1 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/312#-:-LFywCMdD7OeifJXemAh:b-8cso4u)*

irahopkinson:
Reviewed 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/312)*

PR #317, opened 2018-06-28
Retain the value while refreshing the screen

DineshKumarGovindarajan:
I have done some changes and now all test cases got passed.

When we are creating a new entry and while word field is empty and some other fields are filled, autosave has been restricted in this scenario.

Without saving if we are refreshing the screen data should be retained on browser side (LocalStorage has been used).

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/317)


DineshKumarGovindarajan:
i have changed the local storage to localforage concept and given pull request

Umasengoden:

PR #327, opened 2018-07-06
Locally store and use interface language code

irahopkinson:
- this improves behaviour when no project is in scope, eg login

LF E2E tests passing.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/327)


Nateowami:
E2e tests are mostly passing for me. There are more failures now than there have been in the past though (eight in editor-comments), but those failures are on master too. I did have two of the new tests fail once, but they don't consistently fail.

PR #329, opened 2018-07-10
Assignuserrole

Umasengoden:
newly invited/added members in LF should be observer role --- Assigning user role as per selection or else by default user role (commenter) will be assigned

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/329)


Umasengoden:

PR #349, opened 2018-07-31
Fix activity feed

irahopkinson:
- reworded activity summary when a user is added to project
- add project name to activity migration

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/349)

### DIFF
--- a/src/Api/Library/Shared/Script/Migration/FixActivityUserRelated.php
+++ b/src/Api/Library/Shared/Script/Migration/FixActivityUserRelated.php
@@ -35,6 +35,7 @@ class FixActivityUserRelated
         foreach ($projectList->entries as $projectParams) { // foreach existing project
             $projectId = $projectParams['id'];
             $project = new ProjectModel($projectId);
+            print("\n-------------  $project->projectName.\n");
             $activityEntries = ActivityListDto_lf_v1_5::getActivityForProject($project);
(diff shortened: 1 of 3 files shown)


irahopkinson:
___
*[src/Api/Library/Shared/Script/Migration/FixActivityUserRelated.php, line 38 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/349#-LIibWd14LoGrWo8LuG2:-LIisA3G4WfWHmLbEjsQ:b-strk59) ([raw file](https://github.com/sillsdev/web-languageforge/blob/15442d36b917b560587d546f18087c78072dd78a/src/Api/Library/Shared/Script/Migration/FixActivityUserRelated.php#L38)):*
<details><summary><i>Previously, nigel-wells (Nigel Wells) wrote…</i></summary><blockquote>

Does the project name need to be printed even if no activities require updating? Might get confusing as you won't know which projects do and which projects won't.
</blockquote></details>

I added this back in because when running the migration script in 'test' mode on the *QA* server I got 9 warnings about trying to unset indexes that didn't exist. After adding this line of code I could see that all of these warnings were on test projects. So I just ran the script in 'run' mode anyway. I want this line included in case the same thing happens when we run it on the *live* server.

PR #375, opened 2018-09-14
SF-42 Forgot Password

JamesPrabu:
Uma and I completed following points

* Disigned ForgotPassword Page
* Added get and post methods
* Verified username/email
* Added logic for ResetPasswordKey generation
* Saved ResetPasswordKey and ResetPasswordExpirationDate
* Added method for email sending

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/375)

### DIFF
--- a/src/SIL.XForge.Identity/Controllers/Account/AccountController.cs
+++ b/src/SIL.XForge.Identity/Controllers/Account/AccountController.cs
@@ -40,6 +40,7 @@ namespace SIL.XForge.Identity.Controllers.Account
         private readonly IAuthenticationSchemeProvider _schemeProvider;
         private readonly IEventService _events;
         private readonly IOptions<SiteOptions> _options;
+        private IEmailService _emailService;
 
         public AccountController(
             IIdentityServerInteractionService interaction,
@@ -47,7 +48,8 @@ namespace SIL.XForge.Identity.Controllers.Account
             IAuthenticationSchemeProvider schemeProvider,
             IEventService events,
             IRepository<UserEntity> users,
-            IOptions<SiteOptions> options)
+            IOptions<SiteOptions> options,
+            IEmailService emailService)
         {
             _users = users;
             _interaction = interaction;
@@ -55,6 +57,7 @@ namespace SIL.XForge.Identity.Controllers.Account
             _schemeProvider = schemeProvider;
             _events = events;
             _options = options;
+            _emailService = emailService;
         }
 
         /// <summary>
@@ -218,8 +221,7 @@ namespace SIL.XForge.Identity.Controllers.Account
                     "<p>Please click this link to <a href='https://beta.scriptureforge.local/account/resetpassword?token=" + resetPasswordKey + "' target='_blank'>Reset Your Password</a>.</p> " +
                     "<p>This link will be valid for 1 week only.</p><p>Regards,<br>The Scripture Forge team</p></div>";
 
-                IEmailService emailService = new EmailService(_options.Value);
-                emailService.SendEmail(emailId, subject, body);
+                _emailService.SendEmail(emailId, subject, body);
 
                 LoginViewModel vm = await BuildLoginViewModelAsync("");
                 vm.ResetPasswordMessage = "Password Reset email sent for username " + model.UsernameOrEmail;
--- a/src/SIL.XForge.Identity/Controllers/Account/AccountController.cs
+++ b/src/SIL.XForge.Identity/Controllers/Account/AccountController.cs
@@ -40,7 +40,6 @@ namespace SIL.XForge.Identity.Controllers.Account
         private readonly IAuthenticationSchemeProvider _schemeProvider;
         private readonly IEventService _events;
         private readonly IOptions<SiteOptions> _options;
-        private readonly IEmailService _emailService;
 
         public AccountController(
             IIdentityServerInteractionService interaction,
@@ -48,8 +47,7 @@ namespace SIL.XForge.Identity.Controllers.Account
             IAuthenticationSchemeProvider schemeProvider,
             IEventService events,
             IRepository<UserEntity> users,
-            IOptions<SiteOptions> options,
-            IEmailService emailService)
+            IOptions<SiteOptions> options)
         {
             _users = users;
             _interaction = interaction;
@@ -57,7 +55,6 @@ namespace SIL.XForge.Identity.Controllers.Account
             _schemeProvider = schemeProvider;
             _events = events;
             _options = options;
-            _emailService = emailService;
         }
 
         /// <summary>
@@ -215,13 +212,15 @@ namespace SIL.XForge.Identity.Controllers.Account
                 .Set(u => u.ResetPasswordExpirationDate, DateTime.Now.AddDays(7)));
             if (user != null)
             {
-                string emailId = "jamesprabud@gmail.com";
+                string emailId = user.Email;
                 string subject = "Scripture Forge Forgotten Password Verification";
                 string body = "<div class=''><h1>Reset Password for " + user.Username + "</h1> " +
                     "<p>Please click this link to <a href='https://beta.scriptureforge.local/account/resetpassword?token=" + resetPasswordKey + "' target='_blank'>Reset Your Password</a>.</p> " +
                     "<p>This link will be valid for 1 week only.</p><p>Regards,<br>The Scripture Forge team</p></div>";
-                var siteOptions = _options.Value;
-                _emailService.SendEmail(emailId, subject, body, siteOptions.Domain, siteOptions.Name, siteOptions.SmtpServer, siteOptions.PortNumber);
+
+                IEmailService emailService = new EmailService(_options.Value);
+                emailService.SendEmail(emailId, subject, body);
+
                 LoginViewModel vm = await BuildLoginViewModelAsync("");
                 vm.ResetPasswordMessage = "Password Reset email sent for username " + model.UsernameOrEmail;
                 return View("Login", vm);
--- a/src/SIL.XForge.Identity/Controllers/Account/AccountController.cs
+++ b/src/SIL.XForge.Identity/Controllers/Account/AccountController.cs
@@ -19,12 +19,9 @@ using MongoDB.Driver.Linq;
 using MongoDB.Driver;
 using SIL.XForge.DataAccess;
 using SIL.XForge.Models;
-using MailKit;
-using MailKit.Net.Smtp;
-using MimeKit;
+using SIL.XForge.Services;
 using Microsoft.Extensions.Options;
 using SIL.XForge.Configuration;
-using Microsoft.AspNetCore.Hosting;
 
 namespace SIL.XForge.Identity.Controllers.Account
 {
@@ -43,7 +40,7 @@ namespace SIL.XForge.Identity.Controllers.Account
         private readonly IAuthenticationSchemeProvider _schemeProvider;
         private readonly IEventService _events;
         private readonly IOptions<SiteOptions> _options;
-        public IHostingEnvironment _environment { get; }
+        private readonly IEmailService _emailService;
 
         public AccountController(
             IIdentityServerInteractionService interaction,
@@ -52,7 +49,7 @@ namespace SIL.XForge.Identity.Controllers.Account
             IEventService events,
             IRepository<UserEntity> users,
             IOptions<SiteOptions> options,
-            IHostingEnvironment environment)
+            IEmailService emailService)
         {
             _users = users;
             _interaction = interaction;
@@ -60,7 +57,7 @@ namespace SIL.XForge.Identity.Controllers.Account
             _schemeProvider = schemeProvider;
             _events = events;
             _options = options;
-            _environment = environment;
+            _emailService = emailService;
         }
 
         /// <summary>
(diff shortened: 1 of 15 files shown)


src/SIL.XForge.Scripture/appsettings.json
```diff
@@ -5,8 +5,11 @@
     }
   },
   "AllowedHosts": "*",
-  "System": {
-    "Hostname": "beta.scriptureforge.org"
+  "Site": {
```
jasonleenaylor: I know this is still in process, but there's a couple of things to look at here.
I think the System section was removed by accident and the SmtpServer probably still needs an update.

src/SIL.XForge.Identity/Controllers/Account/AccountController.cs
```diff
@@ -167,7 +176,91 @@ public async Task<IActionResult> Login(LoginInputModel model, string button)
...
+            else
+            {
+                model.EnableErrorMessage = true;
+                return View("ForgotPassword", model);
+            }
+        }
+
+        public void SendEmail(string subject, string body, string emailId)
+        {
+            string fromAddress = "no-reply@" + _config.GetSection("Site").GetSection("Domain").Value;
```
ddaspit: You should use SiteOptions instead of IConfiguration to get these config values.

src/SIL.XForge.Identity/Controllers/Account/AccountController.cs
```diff
@@ -284,6 +378,15 @@ private async Task<LoginViewModel> BuildLoginViewModelAsync(LoginInputModel mode
             return vm;
         }
 
+        private async Task<ForgotPasswordViewModel> BuildForgotPasswordViewModelAsync()
```
ddaspit: This does not need to be an async method.

src/SIL.XForge.Identity/Controllers/Account/ForgotPasswordInputModel.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Brock Allen & Dominick Baier. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+
+
+using System.ComponentModel.DataAnnotations;
+
+
+namespace SIL.XForge.Identity.Controllers.Account
+{
+    public class ForgotPasswordInputModel
```
ddaspit: This class isn't used anywhere. You can remove it.

src/SIL.XForge.Identity/Controllers/Account/AccountController.cs
```diff
@@ -167,7 +176,91 @@ public async Task<IActionResult> Login(LoginInputModel model, string button)
...
+            return View("ForgotPassword", vm);
+        }
+
+        /// <summary>
+        /// Validate username or email and send email
+        /// </summary>
+        [HttpPost]
+        public async Task<IActionResult> ForgotPassword(ForgotPasswordViewModel model)
+        {
+            UserEntity user = await _users.Query().SingleOrDefaultAsync(u => u.Username == model.UsernameOrEmail || u.Email == model.UsernameOrEmail);
```
ddaspit: You can combine the `Query` and `UpdateAsync` calls into a single `UpdateAsync` call. Use the `UpdateAsync` overload that allows you to specify a filter.

src/SIL.XForge.Identity/Controllers/Account/AccountController.cs
```diff
@@ -284,6 +382,15 @@ private async Task<LoginViewModel> BuildLoginViewModelAsync(LoginInputModel mode
             return vm;
         }
 
+        private ForgotPasswordViewModel BuildForgotPasswordViewModelAsync()
```
ddaspit: Since this is no longer an async method, you can rename this to `BuildForgotPasswordViewModel`.

src/SIL.XForge.Identity/Controllers/Account/AccountController.cs
```diff
@@ -167,7 +179,92 @@ public async Task<IActionResult> Login(LoginInputModel model, string button)
...
+            else
+            {
+                model.EnableErrorMessage = true;
+                return View("ForgotPassword", model);
+            }
+        }
+
+        public void SendEmail(string subject, string body, string emailId)
+        {
+            var siteOptions = _config.GetOptions<SiteOptions>();
```
ddaspit: You should be able to use the `_options` field instead of getting a `SiteOptions` from `_config`. You should need `IConfiguration` at all, so that can be removed entirely.

src/SIL.XForge.Identity/Controllers/Account/AccountController.cs
ddaspit: You should be able to do something like this:
``` cs
byte[] time = BitConverter.GetBytes(DateTime.UtcNow.ToBinary()); 
byte[] key = Guid.NewGuid().ToByteArray(); 
string resetPasswordKey = Convert.ToBase64String(time.Concat(key).ToArray());
UserEntity user = await _users.UpdateAsync(u => u.Username == model.UsernameOrEmail || u.Email == model.UsernameOrEmail, update => update
    .Set(u => u.ResetPasswordKey, resetPassword)
    .Set(u => u.ResetPasswordExpirationDate, DateTime.Now.AddDays(7)));
if (user != null)
{
    // Send email and redirect to login page
}
else
{
    // Redirect to forgot password page with error message
}
````

File: src/SIL.XForge.Identity/Controllers/Account/AccountController.cs (hunk 6 above)
Comment (JamesPrabu) [reply src/SIL.XForge.Identity/Controllers/Account/AccountController.cs]: Yes, we removed async.

File: src/SIL.XForge.Identity/Controllers/Account/AccountController.cs (hunk 7 above)
Comment (JamesPrabu) [reply src/SIL.XForge.Identity/Controllers/Account/AccountController.cs]: Yes, We removed config related code.

File: src/SIL.XForge.Identity/Controllers/Account/AccountController.cs (hunk 5 above)
Comment (JamesPrabu) [reply src/SIL.XForge.Identity/Controllers/Account/AccountController.cs]: We implemented your logic and joined query and updateAsync together. Thank you for your logic and code.

**Comment by ddaspit** [comment]
This is looking good. Just need some unit tests.

File: src/SIL.XForge.Identity/Controllers/Account/AccountController.cs [hunk 8]
```diff
@@ -166,7 +179,92 @@ public async Task<IActionResult> Login(LoginInputModel model, string button)
...
+        /// Validate username or email and send email
+        /// </summary>
+        [HttpPost]
+        public async Task<IActionResult> ForgotPassword(ForgotPasswordViewModel model)
+        {
+            byte[] time = BitConverter.GetBytes(DateTime.UtcNow.ToBinary());
+            byte[] key = Guid.NewGuid().ToByteArray();
+            string resetPasswordKey = Convert.ToBase64String(time.Concat(key).ToArray());
+            UserEntity user = await _users.UpdateAsync(u => u.Username == model.UsernameOrEmail || u.Email == model.UsernameOrEmail, update => update
+            .Set(u => u.ResetPasswordKey, resetPasswordKey)
```
ddaspit: I would indent these two lines, since it is a continuation of the previous line of code.

src/SIL.XForge.Identity/Controllers/Account/AccountController.cs
```diff
@@ -166,7 +179,92 @@ public async Task<IActionResult> Login(LoginInputModel model, string button)
...
+            .Set(u => u.ResetPasswordExpirationDate, DateTime.Now.AddDays(7)));
+
+            if (user != null)
+            {
+                if (_environment.IsDevelopment())
```
ddaspit: Instead of disabling this code in the development environment, I would create an `EmailService` class and a corresponding `IEmailService` interface in the SIL.XForge project. You could then mock the email service. You could use NSubstitute to verify that the correct data is passed to the EmailService.

test/SIL.XForge.Identity.Tests/Controllers/Account/AccountControllerTests.cs
```diff
@@ -62,6 +65,37 @@ public async Task Login_IncorrectPassword()
             await env.Events.Received().RaiseAsync(Arg.Any<UserLoginFailureEvent>());
         }
 
+       [Test]
+        public async Task ForgotPassword_CorrectUsernameOrEmail()
```
ddaspit: I would check to make sure that the user was updated correctly in the database.

src/SIL.XForge.Identity/Controllers/Account/AccountController.cs
ddaspit: Instead of disabling this code in the development environment, I would create an `EmailService` class and a corresponding `IEmailService` interface in the SIL.XForge project. You could then mock the email service. You could use NSubstitute to verify that the correct data is passed to the EmailService.

test/SIL.XForge.Identity.Tests/Controllers/Account/AccountControllerTests.cs
ddaspit: I would check to make sure that the user was updated correctly in the database.

src/SIL.XForge.Identity/Controllers/Account/AccountController.cs
ddaspit: I would indent these two lines, since it is a continuation of the previous line of code.

test/SIL.XForge.Identity.Tests/Controllers/Account/AccountControllerTests.cs
JamesPrabu: We have written new unit test to ensure ResetPasswordKey and ResetPasswordExpirationDate updated in database.

src/SIL.XForge.Identity/Controllers/Account/AccountController.cs
JamesPrabu: We have indented those lines.

src/SIL.XForge.Identity/Controllers/Account/AccountController.cs
JamesPrabu: We have written EmailService and send email through that. Also we have added one test case for that.

test/SIL.XForge.Identity.Tests/Controllers/Account/AccountControllerTests.cs
```diff
@@ -97,6 +97,43 @@ public async Task ForgotPassword_IncorrectUsernameOrEmail()
             Assert.IsTrue(model.EnableErrorMessage == true);
         }
 
+        [Test]
+        public async Task ForgotPassword_UpdateResetPasswordKeyAndExpirationDate()
```
ddaspit: I would combine this test with the `ForgotPassword_CorrectUsernameOrEmail` test. I don't think we need to separate tests.

test/SIL.XForge.Identity.Tests/Controllers/Account/AccountControllerTests.cs
```diff
@@ -145,6 +184,7 @@ public TestEnvironment()
             public IAuthenticationService AuthService { get; }
             public IEventService Events { get; }
             public AccountController Controller { get; }
+            public MemoryRepository<UserEntity> users {get; set;}
```
ddaspit: Change this property name to `Users` instead of `users`.

test/SIL.XForge.Identity.Tests/Controllers/Account/AccountControllerTests.cs
```diff
@@ -97,6 +97,43 @@ public async Task ForgotPassword_IncorrectUsernameOrEmail()
...
+            var env = new TestEnvironment();
+            var model = new ForgotPasswordViewModel
+            {
+                UsernameOrEmail = "user",
+                EnableErrorMessage = false
+            };
+            IActionResult result = await env.Controller.ForgotPassword(model);
+            UserEntity user = await env.users.Query().SingleOrDefaultAsync(x => x.Username == model.UsernameOrEmail);
+            Assert.True(user.ResetPasswordKey != "jGc6Qe4i1kgM+aA4LVczTJwfHx2YuDR/", "ResetPasswordKey not saved.");
+            Assert.True(user.ResetPasswordExpirationDate != DateTime.Now, "ResetPasswordExpirationDate not saved.");
```
ddaspit: You should test `ResetPasswordExpirationDate != default(DateTime)`. This will ensure that the property was set.

src/SIL.XForge/Services/EmailService.cs
```diff
@@ -0,0 +1,32 @@
+using System;
+using MailKit;
+using MailKit.Net.Smtp;
+using MimeKit;
+
+namespace SIL.XForge.Services
+{
+    public class EmailService : IEmailService
+    {
+        public string SendEmail(string emailId, string subject, string body, string domain, string name, string smtpServer, string portNumber)
```
ddaspit: I would inject `SiteOptions` into this class instead of passing `smtpServer` and `portNumber` into this method.

test/SIL.XForge.Identity.Tests/Controllers/Account/AccountControllerTests.cs
```diff
@@ -97,6 +97,43 @@ public async Task ForgotPassword_IncorrectUsernameOrEmail()
...
+        }
+
+        [Test]
+        public async Task ForgotPassword_SendEmail()
```
ddaspit: You don't need to test the `EmailService` class. In the unit test `ForgotPassword_CorrectUsernameOrEmail`, you can check that the `SendEmail` method of the `IEmailService` mock was called with the correct parameters. See the NSubstitute [docs](http://nsubstitute.github.io/help/received-calls/) to see how to check for a received call with specified arguments. I will clarify what I want in the standup meeting.

test/SIL.XForge.Identity.Tests/Controllers/Account/AccountControllerTests.cs
JamesPrabu: We added this test with CorrectusernameOrEmail test.

test/SIL.XForge.Identity.Tests/Controllers/Account/AccountControllerTests.cs
JamesPrabu: Yes, we change users into Users.

test/SIL.XForge.Identity.Tests/Controllers/Account/AccountControllerTests.cs
JamesPrabu: We changes to default(DateTime).

src/SIL.XForge/Services/EmailService.cs
JamesPrabu: We injected SiteOptions to EmailServer's constructor.

test/SIL.XForge.Identity.Tests/Controllers/Account/AccountControllerTests.cs
JamesPrabu: We removed unit test for EmaiService and added EmailServer mock test in CorrectusernameOrEmail unit test. Thank you.

test/SIL.XForge.Identity.Tests/Controllers/Account/AccountControllerTests.cs
```diff
@@ -62,6 +65,46 @@ public async Task Login_IncorrectPassword()
...
+
+            Assert.That(result, Is.TypeOf<ViewResult>());
+            Assert.IsTrue(model.EnableErrorMessage == false);
+
+            UserEntity user = await env.Users.Query().SingleOrDefaultAsync(x => x.Username == model.UsernameOrEmail);
+            Assert.True(user.ResetPasswordKey != "jGc6Qe4i1kgM+aA4LVczTJwfHx2YuDR/", "ResetPasswordKey not saved.");
+            Assert.True(user.ResetPasswordExpirationDate != default(DateTime), "ResetPasswordExpirationDate not saved.");
+
+            string emailId = "abc@fakegmail.com";
+            string subject = "Scripture Forge Forgotten Password Verification";
```
ddaspit: Did you intend to check `subject` and `body`? You specified the variables, but you don't check them.

src/SIL.XForge.Identity/Controllers/Account/AccountController.cs
```diff
@@ -34,20 +39,25 @@ public class AccountController : Controller
         private readonly IClientStore _clientStore;
         private readonly IAuthenticationSchemeProvider _schemeProvider;
         private readonly IEventService _events;
+        private readonly IOptions<SiteOptions> _options;
+        private IEmailService _emailService;
```
ddaspit: You can make this `readonly`.

src/SIL.XForge.Identity/SIL.XForge.Identity.csproj
```diff
@@ -11,6 +11,7 @@
     <PackageReference Include="Microsoft.AspNetCore.Mvc" Version="2.1.2" />
     <PackageReference Include="Microsoft.AspNetCore.StaticFiles" Version="2.1.1" />
     <PackageReference Include="Microsoft.Extensions.FileProviders.Embedded" Version="2.1.1" />
+    <PackageReference Include="NETCore.MailKit" Version="2.0.2" />
```
ddaspit: I don't think you need this anymore, since you moved the send email code to `SIL.XForge`.

PR #377, opened 2018-09-18
Fix: SF-39 Change Password feature included

DineshKumarGovindarajan:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/377)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/home/home.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/home/home.component.html
@@ -16,3 +16,5 @@
 <p>
 <p><button mat-stroked-button [routerLink]="['/connect-project']">Connect Project</button></p>
 <p><button mat-stroked-button (click)="logOut()" id="logout">Log Out</button></p>
+
+<button mat-stroked-button [routerLink]="['/change-password']">Change Password</button>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/home/home.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/home/home.component.spec.ts
@@ -1,7 +1,6 @@
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 import { RouterTestingModule } from '@angular/router/testing';
 import { OAuthService } from 'angular-oauth2-oidc';
-
 import { HomeComponent } from './home.component';
 
 class HomeComponentPage {
@@ -30,9 +29,7 @@ describe('HomeComponent', () => {
 
   beforeEach(async(() => {
     TestBed.configureTestingModule({
-      imports: [
-        RouterTestingModule
-      ],
+      imports: [RouterTestingModule],
       declarations: [ HomeComponent ],
       providers: [
         { provide: OAuthService, useValue: oauthServiceStub }
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/user.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/user.service.ts
@@ -21,6 +21,10 @@ export class UserService extends ResourceService<User, UserAttributes, ResourceR
     return claims['sub'];
   }
 
+  changePassword(newPassword: string): Promise<void> {
+    return this.jsonApiService.updateAttributes(this.identity(this.currentUserId), { password: newPassword }, false, true);
+  }
+
   currentUserIdentity(): RecordIdentity {
     return this.identity(this.currentUserId);
   }
--- a/src/SIL.XForge.Scripture/ClientApp/package-lock.json
+++ b/src/SIL.XForge.Scripture/ClientApp/package-lock.json
@@ -2847,6 +2847,14 @@
         "tslib": "^1.9.0"
       }
     },
+    "angular-password-strength-meter": {
+      "version": "0.0.4",
+      "resolved": "https://registry.npmjs.org/angular-password-strength-meter/-/angular-password-strength-meter-0.0.4.tgz",
+      "integrity": "sha512-EPo0R5xPUST/q1hjEVnn5S2IIqPTfLkuTeYSJTKmas2lchADiiNOUw2g5CZGJy+Kn41m7fuW6kHt8EVD8jxgAQ==",
+      "requires": {
+        "tslib": "^1.9.0"
+      }
+    },
     "ansi-html": {
       "version": "0.0.7",
       "resolved": "https://registry.npmjs.org/ansi-html/-/ansi-html-0.0.7.tgz",
@@ -14405,6 +14413,11 @@
       "version": "0.8.26",
       "resolved": "https://registry.npmjs.org/zone.js/-/zone.js-0.8.26.tgz",
       "integrity": "sha512-W9Nj+UmBJG251wkCacIkETgra4QgBo/vgoEkb4a2uoLzpQG7qF9nzwoLXWU5xj3Fg2mxGvEDh47mg24vXccYjA=="
+    },
+    "zxcvbn": {
+      "version": "4.4.2",
+      "resolved": "https://registry.npmjs.org/zxcvbn/-/zxcvbn-4.4.2.tgz",
+      "integrity": "sha1-KOwXzwl0PtyrBW3dixsGJizHPDA="
     }
   }
 }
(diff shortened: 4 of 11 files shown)


src/SIL.XForge.Scripture/ClientApp/src/xforge-common/json-api.service.ts
```diff
@@ -316,4 +317,11 @@ export class JSONAPIService {
       return ops;
     }, { update: [JSONAPIService.BACKUP] });
   }
+
+  changePassword(newPassword: string): Observable<any> {
```
ddaspit: This should be implemented in `UserService`. Changing a user's password is already implemented in the JSON API backend, so you should be able to use the `updateAttributes` method from `JSONAPIService` to update the password. The method would look something like this:
``` ts
changePassword(newPassword: string): Promise<void> {
  return this.jsonApiService.updateAttributes(this.currentUserId, { password: newPassword }, false, true);
}
```
The last two boolean parameters in the `updateAttributes` call tells Orbit to not persist the change to IndexedDB and to block the promise until the server response comes back.

src/SIL.XForge.Scripture/ClientApp/src/app/change-password/change-password.component.ts
```diff
@@ -0,0 +1,60 @@
...
+
+@Component({
+  selector: 'app-change-password',
+  templateUrl: './change-password.component.html',
+  styleUrls: ['./change-password.component.scss']
+})
+export class ChangePasswordComponent implements OnInit {
+  changePasswordForm: FormGroup;
+  isSubmitted = false;
+  btnChangePasswordEnabled: Boolean = true;
```
ddaspit: I would just use `boolean` as the type instead of `Boolean` to be consist with existing code.

src/SIL.XForge.Scripture/ClientApp/src/app/change-password/change-password.component.ts
```diff
@@ -0,0 +1,60 @@
+import { HttpClient, HttpHeaders } from '@angular/common/http';
```
ddaspit: This component should be moved to the `src/xforge-common` folder, since it will be shared between Language Forge and Scripture Forge.

src/SIL.XForge.Scripture/ClientApp/src/xforge-common/user.service.ts
```diff
@@ -22,4 +22,7 @@ export class UserService extends BaseResourceService<User, UserAttributes> {
     return claims['sub'];
   }
 
+  changePassword(newPassword: string): Promise<any> {
```
ddaspit: The return type should be `Promise<void>` not `Promise<any>`.

src/SIL.XForge.Identity/Controllers/Account/AccountController.cs
```diff
@@ -34,6 +34,7 @@ public class AccountController : Controller
         private readonly IClientStore _clientStore;
         private readonly IAuthenticationSchemeProvider _schemeProvider;
         private readonly IEventService _events;
+        private static string _loggedusername { get; set; }
```
ddaspit: You can remove all of the changes to the backend. They are not necessary.

src/SIL.XForge.Scripture/ClientApp/src/app/core/sfproject.service.ts
```diff
@@ -2,17 +2,23 @@ import { Injectable } from '@angular/core';
...
-  constructor(jsonApiService: JSONAPIService) {
+  constructor(jsonApiService: JSONAPIService, private userService: UserService) {
     super(jsonApiService);
   }
 
   get type(): string {
     return SFProjectConstants.TYPE;
   }
+
+  changePassword(newPassword: string): Promise<any> {
```
ddaspit: You can remove this method from `SFProjectService`. It is not necessary.

src/SIL.XForge/Models/UserEntity.cs
```diff
@@ -38,5 +38,10 @@ public bool VerifyPassword(string password)
         {
             return BCrypt.Net.BCrypt.Verify(password, Password);
         }
+
+        public static string ChangePassword(string password)
```
ddaspit: You can remove this code. It is not being used.

src/SIL.XForge.Scripture/ClientApp/src/app/change-password/change-password.component.ts
```diff
@@ -0,0 +1,53 @@
...
+    });
+  }
+
+  onSubmit() {
+    this.isSubmitted = true;
+    if (this.changePasswordForm.invalid) {
+      return;
+    }
+    if (this.changePasswordForm.value.newPassword && this.changePasswordForm.value.confirmPassword) {
+      this.projectService.changePassword(this.changePasswordForm.value.newPassword).then(data => {
```
ddaspit: You can just call `changePassword` from `UserService` directly.

src/SIL.XForge.Identity/Controllers/Account/StatusModel.cs
```diff
@@ -0,0 +1,13 @@
+
+
+using System.ComponentModel.DataAnnotations;
+
+
+namespace SIL.XForge.Identity.Controllers.Account
```
ddaspit: This class is not used, so you can remove it. I would just revert all code changes on the backend, since the backend doesn't need any changes for this PR.

src/SIL.XForge.Scripture/ClientApp/src/app/core/sfproject.service.ts
```diff
@@ -2,13 +2,15 @@ import { Injectable } from '@angular/core';
 
 import { BaseResourceService } from '@xforge-common/base-resource-service';
 import { JSONAPIService } from '@xforge-common/json-api.service';
+import { UserService } from '@xforge-common/user.service';
```
ddaspit: You can revert all changes to this file. They are not necessary.

src/SIL.XForge.Scripture/ClientApp/src/app/home/home.component.html
```diff
@@ -16,3 +16,5 @@ <h1 *ngIf="name">Hello, {{name}}!</h1>
 <button mat-stroked-button (click)="logoff()">
   Logout
 </button>
+
+<button mat-stroked-button (click)="changePassword()">Change Password</button>
```
ddaspit: You can use `[routerLink]="['/change-password']"` in the `button` element to route to the change password component. You don't need to inject `Router` into the `HomeComponent` class.

src/SIL.XForge.Scripture/ClientApp/src/environments/environment.ts
```diff
@@ -4,5 +4,6 @@
 // The list of which env maps to which file can be found in `.angular-cli.json`.
 
 export const environment = {
-  production: false
+  production: false,
+  baseUrl: 'https://beta.scriptureforge.local/'
```
ddaspit: You can revert all changes to this file. They are not necessary.

src/SIL.XForge.Scripture/ClientApp/src/xforge-common/json-api.service.ts
```diff
@@ -1,4 +1,4 @@
-import { HttpClient } from '@angular/common/http';
+import { HttpClient, HttpHeaders } from '@angular/common/http';
```
ddaspit: You can revert all changes to this file. They are not necessary.

src/SIL.XForge.Identity/Controllers/Account/AccountController.cs
```diff
@@ -83,7 +83,7 @@ public async Task<IActionResult> Login(LoginInputModel model, string button)
             {
```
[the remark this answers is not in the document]
DineshKumarGovindarajan: Changed the following suggestions.

Thank you.
Regards,

Dinesh

src/SIL.XForge.Scripture/ClientApp/src/app/home/home.component.spec.ts
```diff
@@ -29,6 +30,7 @@ describe('HomeComponent', () => {
 
   beforeEach(async(() => {
     TestBed.configureTestingModule({
+      imports: [RouterTestingModule],
```
ddaspit: I don't think you need this since you are using `[routerLink]`. If you don't need it, revert all changes to this file.

DineshKumarGovindarajan: if we use this line (imports: [RouterTestingModule],) in home.component.spec.ts file. home component all test cases will be passed otherwise some test cases got failed. that's why we added.

jasonleenaylor:
Trusting that all passing tests indicates a good merge. LGTM

PR #378, opened 2018-09-18
Revise vagrant

marksvc:
* More RAM
* Updated system packages
* Additional desktop and development settings
* Development tools
* Fix permissions.
* Additional cleanup, to ssh host keys and blanking disk.
* Add some instructions.
* Try to avoid accidentally using a slow mirror.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/378)

### DIFF
--- a/deploy/vagrant_xenial_gui/Vagrantfile
+++ b/deploy/vagrant_xenial_gui/Vagrantfile
@@ -15,7 +15,7 @@ Vagrant.configure("2") do |config|
   end
 
   config.vm.provision "shell", privileged: false, inline: <<~SHELL
-    set -eux
+    set -eux -o pipefail
 
     # Delete and generate unique host keys until handled by base box.
     sudo rm -v /etc/ssh/ssh_host_*
@@ -27,5 +27,7 @@ Vagrant.configure("2") do |config|
     cd ~/src/web-languageforge
     git pull --ff-only --recurse-submodules
     ./refreshDeps.sh
+
+    sudo npm install -g @angular/cli@6.1.5
   SHELL
 end
--- a/deploy/vagrant_xenial_gui/Vagrantfile
+++ b/deploy/vagrant_xenial_gui/Vagrantfile
@@ -5,9 +5,13 @@ Vagrant.configure("2") do |config|
 
   config.vm.box = "NathanielPaulus/xenial64-xforge"
 
+  config.vm.hostname = "vag-u1604-xf"
+
   config.vm.provider "virtualbox" do |vb|
+    vb.name="vag-u1604-xf"
     vb.gui = true
-    vb.memory = "2048"
+    vb.cpus = 4
+    vb.memory = "4096"
   end
 
   config.vm.provision "shell", privileged: false, inline: <<~SHELL
@@ -23,7 +27,5 @@ Vagrant.configure("2") do |config|
     cd ~/src/web-languageforge
     git pull --ff-only --recurse-submodules
     ./refreshDeps.sh
-
-    SHELL
+  SHELL
 end
-
--- a/deploy/vagrant_xenial_gui/Vagrantfile
+++ b/deploy/vagrant_xenial_gui/Vagrantfile
@@ -11,6 +11,7 @@ Vagrant.configure("2") do |config|
     vb.name="vag-u1604-xf"
     vb.gui = true
     vb.cpus = 4
+    # At least 3 GiB of RAM is needed to run tests.
     vb.memory = "4096"
   end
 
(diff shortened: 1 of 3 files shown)


marksvc:
In my testing, LF tests passed, and all but one SF tests passed. 
dotnet run compiles.

deploy/vagrant_xenial_gui/Vagrantfile
```diff
@@ -5,13 +5,17 @@ Vagrant.configure("2") do |config|
...
   config.vm.box = "NathanielPaulus/xenial64-xforge"
 
+  config.vm.hostname = "vag-u1604-xf"
+
   config.vm.provider "virtualbox" do |vb|
+    vb.name="vag-u1604-xf"
     vb.gui = true
-    vb.memory = "2048"
+    vb.cpus = 4
+    vb.memory = "4096"
```
irahopkinson: First I had to reduce the CPUs and RAM to get it to work at all on my laptop (it only has 4GB RAM and 4 CPUs). I'm not sure how wise it is to specify this so high when you don't know what machine it will be run on. A user could set those appropriately after the box is created. We would have to add that to the docs.

deploy/virtualbox_xenial_gui_creation/Vagrantfile
```diff
@@ -1,67 +1,205 @@
...
-      gsettings set org.gnome.desktop.screensaver lock-enabled false
-      gsettings set org.gnome.desktop.session idle-delay 0
-      gsettings set com.canonical.Unity.Lenses remote-content-search none
-      gsettings set com.canonical.Unity.Launcher favorites "['application://org.gnome.Nautilus.desktop', 'application://firefox.desktop', 'application://unity-control-center.desktop', 'unity://running-apps', 'application://gnome-terminal.desktop', 'unity://expo-icon', 'unity://devices']"
-    else
-      echo ***COULD NOT MODIFY GSETTINGS***
-    fi
+    [[ -n "$DBUS_SESSION_BUS_ADDRESS" ]] || exit 1
+    # Somehow dconf was owned by root.
+    sudo chown -R vagrant:vagrant /run/user/1000/dconf
```
irahopkinson: I think it failed here with this error:
```
    default: +++ sudo cat /proc/1227/environ
    default: ++ export DBUS_SESSION_BUS_ADDRESS=unix:abstract=/tmp/dbus-FIRX4dFQoA,guid=3f0788233baf403c328f7aaf5ba6332f
    default: ++ DBUS_SESSION_BUS_ADDRESS=unix:abstract=/tmp/dbus-FIRX4dFQoA,guid=3f0788233baf403c328f7aaf5ba6332f
    default: ++ [[ -n unix:abstract=/tmp/dbus-FIRX4dFQoA,guid=3f0788233baf403c328f7aaf5ba6332f ]]
    default: ++ sudo chown -R vagrant:vagrant /run/user/1000/dconf
    default: chown: cannot access '/run/user/1000/dconf'
    default: : No such file or directory
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
```

jasonleenaylor: This is the error you get if you don't log into the machine while provisioning is running. An improvement might be to have the provisioning script pause and prompt the user to login if it gets to this point and finds the directory missing.

deploy/vagrant_xenial_gui/Vagrantfile
jasonleenaylor: We know that the e2e tests won't pass without at least 3 gigabytes of memory. Maybe your laptop doesn't meet our minimum development specs. :-)

deploy/virtualbox_xenial_gui_creation/instructions.md
```diff
@@ -0,0 +1,36 @@
+# Base box creation instructions
+
+When creating the base box, log in to the desktop immediately so gsettings will work.
```
jasonleenaylor: @irahopkinson Failing to do this is what caused the error you ran into.

irahopkinson: Thanx

deploy/virtualbox_xenial_gui_creation/Vagrantfile
irahopkinson: Yes I did see that instruction and had logged in previously but must have forgotten on the one that actually failed. Thanx

marksvc:
So it sounds like we should submit this pull request and then consider further enhancements in additional PRs. I saw in chat that the ssl problem is working now.

PR #372, opened 2018-09-12
fix for LF-184 and a fix for php7

tambi-soft:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/372)

PR #381, opened 2018-09-26
 Fix RTL in compact list view (LF-184)

ermshiperete:
Display RTL scripts on the right hand side of the compact list view.

This supersedes PR #372

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/381)


ermshiperete:
@irahopkinson Thanks for kicking off a build. Good to know that they're currently not started automatically.

I tested this change and it fixes the problem, so I'll merge this PR.

PR #382, opened 2018-10-05
Fix: SF-41 Reset Password

Umasengoden:
* Page disigned
* Added validation
* Saved password
* Added unit tests

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/382)

### DIFF
--- a/src/SIL.XForge.Identity/Controllers/Account/AccountController.cs
+++ b/src/SIL.XForge.Identity/Controllers/Account/AccountController.cs
@@ -177,17 +177,30 @@ namespace SIL.XForge.Identity.Controllers.Account
         }
 
         /// <summary>
-        /// Show reset password page
+        /// ResetPassword from the token produced by ForgotPassword.
         /// </summary>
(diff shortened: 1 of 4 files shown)


jasonleenaylor:
___
*[src/SIL.XForge.Identity/Controllers/Account/AccountController.cs, line 170 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/382#-LO3wB8rB1XJaCHou00q:-LO3wB8rB1XJaCHou00r:b-ytj30t) ([raw file](https://github.com/sillsdev/web-languageforge/blob/a17814d1b45a22a5339ec49c544de1723db114a8/src/SIL.XForge.Identity/Controllers/Account/AccountController.cs#L170)):*
> ```Smalltalk
>                 }
> 
>                 await _events.RaiseAsync(new UserLoginFailureEvent(model.Username, "invalid credentials"));
> ```

This looks like a mistake. I think it may also be the source of the unit test failures.

PR #384, opened 2018-10-05
Correct appsettings file to allow login on development machines

RaymondLuong3:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/384)

### DIFF
--- a/src/SIL.XForge.Scripture/appsettings.Development.json
+++ b/src/SIL.XForge.Scripture/appsettings.Development.json
@@ -8,11 +8,8 @@
     }
   },
   "Site": {
-    "Domain": "beta.scriptureforge.local"
-  },
-  "Site": {
     "Name": "Scripture Forge (beta)",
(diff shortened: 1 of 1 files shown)


src/SIL.XForge.Scripture/appsettings.Development.json
```diff
@@ -7,12 +7,9 @@
...
     }
   },
-  "Site": {
-    "Domain": "beta.scriptureforge.local"
-  },
   "Site": {
     "Name": "Scripture Forge (beta)",
-    "Domain": "beta.scriptureforge.org",
+    "Domain": "beta.scriptureforge.local",
     "SmtpServer": "mail.ecgroup.in",
```
jasonleenaylor: We probably want to find a better default value for this. This will allow actual e-mail sending from inside ecgroup, but isn't helpful elsewhere.

PR #391, opened 2018-10-12
Refactor frontend resource model to use classes

ddaspit:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/391)


src/SIL.XForge.Scripture/ClientApp/src/app/core/sfproject.service.ts
```diff
@@ -2,19 +2,18 @@ import { Injectable } from '@angular/core';
...
+export class SFProjectService extends OptimisticResourceService<SFProject> {
 
   constructor(jsonApiService: JSONAPIService) {
-    super(jsonApiService, SFPROJECT);
+    super(jsonApiService, SFProject.TYPE);
   }
 
-  onlineCreate(resource: Partial<SFProject>): Promise<string> {
-    return this.jsonApiService.create(record(this.type, resource), true, true);
+  onlineCreate(resource: SFProject): Promise<string> {
```
jasonleenaylor: Why a promise instead of an observable here?

src/SIL.XForge.Scripture/ClientApp/src/xforge-common/jsonapi.service.ts
```diff
@@ -5,19 +5,20 @@ import Coordinator, {
...
 import { ObjectId } from 'bson';
 import { from, fromEventPattern, merge, Observable } from 'rxjs';
 import { map, startWith } from 'rxjs/operators';
 
 import { XForgeJSONAPISource } from './jsonapi/xforge-jsonapi-source';
 import { LiveQueryObservable } from './live-query-observable';
+import { getResourceRefType, getResourceType, Resource, ResourceRef } from './models/resource';
 
 @Injectable({
   providedIn: 'root'
```
jasonleenaylor: JSONAPIService needs a good class comment to help developers know the value it provides and how it should be used.

jasonleenaylor: In fact this file is so critical to the success of our Offline capability I think we should consider carefully documenting all the public methods with some guidance on where and how they are expected to be used.

src/SIL.XForge.Scripture/ClientApp/src/xforge-common/optimistic-resource.service.ts
```diff
@@ -1,54 +1,47 @@
...
+import { FindRecordsTerm, RecordIdentity } from '@orbit/data';
 
 import { JSONAPIService } from './jsonapi.service';
 import { LiveQueryObservable } from './live-query-observable';
-import { Resource, ResourceAttributes } from './models/resource';
-import { identity, record } from './resource-utils';
+import { Resource, ResourceRef } from './models/resource';
 
-export class OptimisticResourceService<TResource extends Resource, TAttrs extends ResourceAttributes> {
+export class OptimisticResourceService<TResource extends Resource> {
```
jasonleenaylor: Class comment please

src/SIL.XForge.Scripture/ClientApp/src/app/core/sfproject.service.ts
ddaspit: We decided that we would use observables for queries and promises for actions and transforms. Observables feel like an unnecessary complication for actions.

PR #402, opened 2018-10-18
Home Page

papeh:
* allow logging in with new home page
* different landing pages for users who are logged in vs. out
* log (in|out), sign up, and change password buttons in a toolbar at the top

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/402)


ddaspit:
I began to review your PR for the Home Page and realized that the code wasn't structured quite the way I expected. Mainly, I expected to see all of the "dev project" stuff in a separate module and the home page component in the app module. Also, the "dev project" represents our main SF app, which requires that the user is logged in in order to use it. I threw together what I am trying to describe in a new branch "feature/homepageWithModule" that is based on your branch. The key thing to look for is the new `DevProjectModule` in the `ClientApp/src/app/dev-project" folder. It would probably be a good idea to move the `HomeComponent` into a separate module as well, so that the app module stays clean.

ddaspit:
Also, we should probably rename "dev-project" to whatever it is going to be when we implement the actual app. I'm not sure what to call it. It really represents the "logged in" area of the application, which includes the "My Projects" view, administration views, and project-specific views. Maybe we could call it something simple like "main".

jasonleenaylor:
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/jsonapi.service.ts, line 65 at r5](https://reviewable.io/reviews/sillsdev/web-languageforge/402#-LPCXcwAAzXmgH0dKfpg:-LPCy2ZcBH34XShniwF8:b961hca) ([raw file](https://github.com/sillsdev/web-languageforge/blob/c27dc03f85be71bd3c49bf04e2d1589576644244/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/jsonapi.service.ts#L65)):*
<details><summary><i>Previously, marksvc wrote…</i></summary><blockquote>

This variable doesn't seem to be checked anywhere. Is it needed?
Maybe it's used by a subclass?
</blockquote></details>

It's checked in the dev-project.component.ts

jasonleenaylor:
___
*[src/SIL.XForge.Scripture/Startup.cs, line 128 at r8](https://reviewable.io/reviews/sillsdev/web-languageforge/402#-LPj55KVBd6gXBfeEwVi:-LPl-cRBCeDl-SxUtojt:bmhipms) ([raw file](https://github.com/sillsdev/web-languageforge/blob/40d8152e99ebd66ff1670ebc483c6c3ec5f7069e/src/SIL.XForge.Scripture/Startup.cs#L128)):*
<details><summary><i>Previously, ddaspit (Damien Daspit) wrote…</i></summary><blockquote>

You can get ride of this line if you move all of the static files to `wwwroot` instead of `sfwwwroot`. You will need to update all of the urls that reference static files, since the path won't begin with `/sf` anymore.
</blockquote></details>

We did it this way intentionally to avoid any potential for future filename conflicts. If we have wwwroot here and in the Identity project and re-use a filename won't it cause an issue?

PR #418, opened 2018-11-01
Add realtime service

ddaspit:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/418)


src/SIL.XForge.Scripture/ClientApp/typings/sharedb.d.ts
```diff
@@ -0,0 +1,81 @@
+declare module 'sharedb/lib/client' {
```
megahirt: @ddaspit I assume you created this typings file because no one exists on npm or elsewhere?  Since you went to the work to create it, it would be good to consider putting this in a separate repo for npm.  The typings files seem to be bleeding edge stuff that some exist and others don't (yet).  And some are lagging behind current versions of libraries, etc.

ddaspit: There is a `@types/sharedb` module for sharedb. Unfortunately, it has a lot of errors, so I didn't bother to use it. I just made my own.

PR #432, opened 2018-11-06
Update to Angular 7

ddaspit:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/432)


megahirt:
If only all the Angular upgrades were this easy :)

PR #436, opened 2018-11-07
Fix unsupported filter exception on e-mail check

jasonleenaylor:
* String.Equals isn't supported by the LINQ Translator

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/436)


RaymondLuong3:
:lgtm:

PR #439, opened 2018-11-08
Fix: SF-53,61 and 62 - System Administration module

samdoss:
* Included the features in system administration module
 * List the users, Add the users, update the users,
   delete the user and change password feature.


src/SIL.XForge.Scripture/ClientApp/src/app/app.module.ts
```diff
@@ -1,33 +1,29 @@
...
-  MatCheckboxModule,
-  MatFormFieldModule,
-  MatInputModule,
-  MatOptionModule,
-  MatProgressBarModule,
-  MatProgressSpinnerModule,
-  MatSelectModule,
-  MatSnackBarModule
-} from '@angular/material';
+import { MatButtonModule, MatCardModule, MatCheckboxModule, MatDialogModule, MatFormFieldModule,
```
jasonleenaylor: please change these back to one per line to help prevent merge conflicts in future commits.

src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/user.ts
```diff
@@ -1,13 +1,15 @@
 import { Resource, ResourceRef } from './resource';
 
-export abstract class User extends Resource {
+export class User extends Resource {
```
jasonleenaylor: I think we want to keep User as an abstract class.

PR #443, opened 2018-11-09
SF-67: Increase Password Reset Security

papeh:
Require username and token to match at all stages

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/443)


irahopkinson:
I don't think that adding the username adds any security. The email must contain both the key and the username and the page request over the wire must also contain both. I'm struggling to see a scenario where a malicious user could get one without the other.

Given the problem outlined above in the code review of not being able to guarantee that a username is even set (we may scrape the username altogether), I'm closing this PR.

PR #444, opened 2018-11-09
Fix: SF-86 Ensure email matches are case-insensitive

RaymondLuong3:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/444)


irahopkinson:
@ddaspit is fixing this a different way that will work properly with Mongo and now that PR#441 is in `sf-develop`. Closing this PR.

PR #449, opened 2018-11-13
vagrant: refreshDeps after ansible

marksvc:
* Fixes running refreshDeps.
* Add a note to more clearly alert user of provisioning problems, and
what to do if so.
* Only pull up to a certain point in the repo, in case a later commit
breaks running the provision script.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/449)

### DIFF
--- a/deploy/vagrant_xenial_gui/Vagrantfile
+++ b/deploy/vagrant_xenial_gui/Vagrantfile
@@ -20,6 +20,17 @@ Vagrant.configure("2") do |config|
 
     function tryharderto() { i=0; until "$@"; do ((++i > 3)) && false; echo >&2 Retrying $i; sleep 2m; done }
 
+    # Canary
+    tee ~/Desktop/development-tools/warning-not-provisioned.txt <<END
+    The presence of this file means that the vagrant virtual machine did not
+    finish provisioning correctly.
+    The best way to get to a working machine state is to backup any data that
+    you want to keep from this virtual machine, and then destroy and re-create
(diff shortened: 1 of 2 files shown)


irahopkinson:
___
*[deploy/vagrant_xenial_gui/Vagrantfile, line 44 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/449#-LREpS0NDtddlCWfIUGo:-LREpS0NDtddlCWfIUGp:ba9qbyf) ([raw file](https://github.com/sillsdev/web-languageforge/blob/58bc4a1b34af60ddb856be4fa792418a934cc311/deploy/vagrant_xenial_gui/Vagrantfile#L44)):*
> ```
>     # Pull only up to a commit that this vagrant was tested to work with.
>     git fetch && git pull --ff-only --recurse-submodules origin ac4b4157ac35abed54efad8a7e711eac0c025f34
> ```
non-blocking: I'm concerned that this will just hide further problems (hey, the provisioning worked) but when they get to the latest commit things might not. Feel free to merge tho if you think this fixes more problems than it might cause.

marksvc:
My assumption is that more meaningfully upgrading and testing a new vagrant machine would be using the base-box creation Vagrantfile instead of this one. But I will updated this PR to not hold the pull back to a specific commit.

marksvc:
Okay, I see, the concern is that someone would be modifying the vagrant by editing just this Vagrantfile, but not know to change this line, and have unintended results. Got it.

PR #456, opened 2018-11-15
Fix TSLint settings

irahopkinson:
- set editor ruler to 120 characters
- remove failing rules


jasonleenaylor:
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/notice/detail-snack-bar.component.ts, line 11 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/456#-LRMyKPL7I3-cTEknglm:-LROSTN_77gPqRxnZEmO:bcpy980) ([raw file](https://github.com/sillsdev/web-languageforge/blob/d9640b15c080e22284d7e45ea181a8799eb5e085/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/notice/detail-snack-bar.component.ts#L11)):*
<details><summary><i>Previously, jasonleenaylor (Jason Naylor) wrote…</i></summary><blockquote>

That's kind of an ugly wrap. I'd probably adjust that later to break between the arguments
</blockquote></details>

Or is this a preferred style?

PR #483, opened 2018-12-04
Bug: SF-110 The selected role should save

samdoss:
* Fixed the issue.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/483)


ddaspit:
This PR will fix the issue, but it does not address the root cause of the problem. This issue occurs because the project component does not correctly update the `role` property in the cached project user object. I will submit a PR that fixes the root cause.

PR #497, opened 2018-12-11
Feature/code generator

jasonleenaylor:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/497)


jasonleenaylor:
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/core/models/sfproject.generated.ts, line 105 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/497#-LT_WZ2v5hAXYrAcPx3y:-LTisuTW5lK4xVz3Kg4Q:b-651tgk) ([raw file](https://github.com/sillsdev/web-languageforge/blob/60263b298fd2f1d8d96cd578aa4ad5e47e205887/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/sfproject.generated.ts#L105)):*
<details><summary><i>Previously, jasonleenaylor (Jason Naylor) wrote…</i></summary><blockquote>

This is a tricky bit. In the C# the reference has an annotation, but that doesn't come through into the json-schema. I'll need to do some thinking.
</blockquote></details>

Ok, found an easy fix I think.

PR #507, opened 2018-12-20
Fix: SF-125 Create verify email component

RaymondLuong3:


RaymondLuong3:
I didn't see if it was possible to create this functionality without making a new component. If there is a simpler approach to this, please let me know.

PR #519, opened 2019-01-09
Fix CodeGenerator issues

ddaspit:
- change SyncJob.State back to a string
- pass type to resource service base classes
- fix/cleanup models
- include Text model when generating code
- remove TYPE field from Resource and ResourceRef
- fix bug when logging into PT in connect project view
- add VS Code task for CodeGenerator
- add VS Code task for running SF backend

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/519)


jasonleenaylor:
___
*[src/SIL.XForge.Scripture/CodeGenerator/Templates/Enum.liquid, line 1 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/519#-LVmjy_p1G1LHOvT5AbF:-LVmjy_p1G1LHOvT5AbG:bhxn9jx) ([raw file](https://github.com/sillsdev/web-languageforge/blob/697e4210a846eb05dc95be423de3cf1b820f00ed/src/SIL.XForge.Scripture/CodeGenerator/Templates/Enum.liquid#L1)):*
> ```Liquid
> {% if HasDescription -%}
> ```

I wouldn't delete this in case we do want to generate enums in the future. Then we wouldn't have to figure out how to get it back.

PR #520, opened 2019-01-10
Add project settings page [SF-105]

jasonleenaylor:
* add route to settings page for each project
FIXME: search for project by id

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/520)


jasonleenaylor:
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/project-settings/project-settings.component.scss, line 8 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/520#-LWA96A3AwlF6-LbTEcm:-LWD2twW5oyL-n1duk0R:b-mg5k7e) ([raw file](https://github.com/sillsdev/web-languageforge/blob/770a3f6a094b459b9ebcbd3acf0c07d6338afd94/src/SIL.XForge.Scripture/ClientApp/src/app/project-settings/project-settings.component.scss#L8)):*

> ```
>   padding-left: 32px;
>   padding-right: 32px;
> ```
These probably won't work well in the responsive view.
</blockquote></details>

Yes, however I haven't figured out how to get this to look even half-way decent in any view without heading to the css rules so I started here. What I really want is to tie it to the placement of the check-box label. It can't be a part of it without messing up the rest of the layout so it needs to be outside of the form-field but respond in lock-step with the check-box label. Ideas?

jasonleenaylor:
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/project-settings/project-settings.component.html, line 2 at r4](https://reviewable.io/reviews/sillsdev/web-languageforge/520#-LWPnJo48OCeqBr6Vo2A:-LWRqfPN30ZeDfOgFUlB:b-qnlc3a) ([raw file](https://github.com/sillsdev/web-languageforge/blob/8425868ad52d5002c9c71582790d2e13316e1ee5/src/SIL.XForge.Scripture/ClientApp/src/app/project-settings/project-settings.component.html#L2)):*

> ```
> mdc-card
> ```
I'm unsure about the use of a card here. I think Material uses elevation for a specific purpose which we aren't doing here. https://material.io/design/environment/surfaces.html#material-environment
</blockquote></details>

I see your point. We have no plans to move the cards around. I'm open to suggestions as what I should do as an alternative. Maybe we can ask Nigel for a suggestion if you don't have one.

PR #521, opened 2019-01-11
SF-76: External sign-up component

ddaspit:


jasonleenaylor:
I didn't find anything to complain about yet other then the name 'External' I understand it after reading through the code but it didn't give me a good idea about what the functionality  was. I'm not sure what a better name is just yet.

For instance, just the class name 'ExternalService' tells me what? If you can find a good name great, otherwise some class/interface level comments could help.

PR #549, opened 2019-02-13
update Vagrantfile to include /var/lib/xforge

gtryus:
This dependency has recently been added.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/549)


megahirt:
@gtryus thanks for this.  I will let Ira review and merge because I don't know enough about the process to include this change.  Looks like a positive contribution.

PR #559, opened 2019-02-22
SF-61: Let the Admin know if an email address is taken

papeh:
- Show a Snackbar notification of any error thrown when creating or updating a user in the SysAdmin User Entry component
- Use constants and `@example.com` in UserServiceTests.cs

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/559)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/system-administration/sa-user-entry.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/system-administration/sa-user-entry.component.spec.ts
@@ -192,6 +192,16 @@ describe('System Administration User Entry Component', () => {
       verify(env.mockedUserService.onlineUpdateAttributes(anything(), anything())).never();
     }));
 
+    it('should report any errors', fakeAsync(() => {
+      const env = new TestUserEntryComponent();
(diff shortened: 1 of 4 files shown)


papeh:
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/system-administration/sa-user-entry.component.ts, line 137 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/559#-LZNt7KDAfMMer5dqvdt:-LZahihCEaaNm-OzaSeq:bua12ei) ([raw file](https://github.com/sillsdev/web-languageforge/blob/acafa69e73bd8159ef559156ac9e85359eff6008/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/system-administration/sa-user-entry.component.ts#L137)):*

> ```
> catch (e)
> ```
Only catch the duplicate email error, re-throw anything else. Look in `sign-up.componet.ts` how to get the email `Conflict` result. Here and below.
</blockquote></details>

Thanks for the pointer. It looks like the sa-user-entry, my-account, and sign-up components are using different API's. `identity.service` (used by `sign-up`) returns a nice result; `user.service` (used by `sa-user-entry` and `my-account` throws. So it looks like we'll need to change the `UserService` API to return a friendly failure result.

I'm also happy to avoid the catch-all in `my-account.component.ts:223` in a separate PR if we feel that's a priority right now.

PR #565, opened 2019-02-26
Update doc, script

marksvc:
The Angular MDC URI might look a little funny, but I find it slightly confusing to find the API docs if I instead am pointed to https://trimox.github.io/angular-mdc-web/ . Throwing a developer right into one of the API docs for a control may make it more intuitive to find the rest of the documentation (it is for me).

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/565)


marksvc:
(answering in slack)

PR #587, opened 2019-03-12
Feature/SF-145

nigel-wells:
- Added answer panel to checking project
- Adjusted font size increase/decrease button size
- Added angular-split

image

I do note that there is still a performance issue with the quill editor being resized. It works fine when there is no wrapping set - you can test this by adding `white-space: nowrap;` to the `.ql-editor p` to see what I mean. That said, I'm not sure if it is the wrapping causing the issue (probably not) or the height required for the quill editor (my current suspicion). If it is a height factor then it will be very interesting to see how it copes with a full chapter of text loaded in.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/587)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/package-lock.json
+++ b/src/SIL.XForge.Scripture/ClientApp/package-lock.json
@@ -1504,6 +1504,14 @@
         "tslib": "^1.9.0"
       }
     },
+    "angular-split": {
+      "version": "2.0.1",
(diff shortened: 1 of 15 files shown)


nigel-wells:
`this.splitContainerElementHeight` will only ever be called when that element is available - it is shown as soon as the page loads and calculations are only needed once a question has been clicked on.

`this.activeQuestion` is currently used in `checking-component.html` for some `*ngIf` statements and for passing through to the answer component.

PR #592, opened 2019-03-16
Reduce audio recording gain from 10 to 1 (LFv1)

Nateowami:
This commit reduces the audio recording gain from 10 to 1, which also eliminates the need to implement clipping.

Work on this was delayed a bit when I discovered an audio recording related bug and attempted to fix it. The audio recorder persists when switching between dictionary entries. That bug has proved to be somewhat more difficult to repair than I anticipated, and I am still working on it.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/592)


Nateowami:
@megahirt Looking through the examples and source code for mic-recorder-to-mp3, it looks like it does not do anything with the gain or trying to account for volume.

I'm guessing recording hardware doesn't make a huge difference to audio volume, though I could be wrong (I just haven't seen any indication that it does make a difference). Distance from the microphone would probably make have an effect.

PR #605, opened 2019-03-21
Fix missing comment indicator in LF entry list (LFv1)

Nateowami:
This issue was caused by `this` referring to the wrong object (possibly starting in 6345e3f4ed27b74c54b51bdbe1395e30557f7bc9). The view calls a function that is a property of the component, but the function is actually defined in the comment service (in the component we have `getEntryCommentCount = this.commentService.getEntryCommentCount`). When AngularJS calls the function it uses the wrong context (most likely the context of the component). When the function references `this.comments` it is undefined.

Because this type of bug is so easy to create I think we should consider whether all methods should be written using arrow functions so `this` will have lexical scope. It's not the default TypeScript syntax, but it is safer and actually more "correct."

The two main ways of writing methods in TypeScript using arrow function are:
``` typescript
public functionA = () => {}
public functionB: () => void = () => {}
```

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/605)

### DIFF
--- a/src/angular-app/bellows/core/offline/lexicon-comments.service.ts
+++ b/src/angular-app/bellows/core/offline/lexicon-comments.service.ts
@@ -108,7 +108,7 @@ export class LexiconCommentService {
   }
 
   getEntryCommentCount(entryId: string): number {
-    if (this.comments == null || this.comments.counts.byEntry[entryId] == null) {
+    if (this.comments.counts.byEntry[entryId] == null) {
       return 0;
     }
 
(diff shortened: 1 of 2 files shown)


megahirt:
[LF-183 Orange comment indicator no longer showing in compact and full list view](https://trello.com/c/RySZZvSI/341-lf-183-orange-comment-indicator-no-longer-showing-in-compact-and-full-list-view)

Nateowami:
Will do. I think it only needs to be changed in one place.

Nateowami:
@megahirt I've made the change and amended the commit. Rather than using an actual arrow function I used the `bind()` method to wrap the function. I assume that's OK. It's functionally equivalent, as it wraps the function internally. It's slightly shorter and doesn't require enumerating the arguments and their types.

PR #610, opened 2019-03-25
Add translate overview page

RaymondLuong3:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/610)


irahopkinson:
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.ts, line 63 at r3](https://reviewable.io/reviews/sillsdev/web-languageforge/610#-Lb2Bh_y4SwDpq_Sr5z_:-Lb2Bh_y4SwDpq_Sr5za:b-nfsilg) ([raw file](https://github.com/sillsdev/web-languageforge/blob/edd00180c5baebd7eb358bfccd1084cb96ee0918/src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.ts#L63)):*
> ```
> this.calculateProgress().then(() => {
> ```
Don't need to use a `then`. The method is async so just `await this.calculateProgress();`.

PR #611, opened 2019-03-26
SF-229 Accommodate data in-progress

marksvc:
SF-229 will benefit from a similar fix in project.component.ts, but here is the first part.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/611)


marksvc:
I'm not certain that this is the *right* solution. I did not yet discover why this problem situation can arise in the first place.

PR #612, opened 2019-03-26
SF-229 Early fetch project from server

marksvc:
A possible solution. See discussion in chat.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/612)


marksvc:
I don't think I understand why the build failed :). Based on the log, I assume it was random?

PR #618, opened 2019-03-28
Ship Language Forge v1.6.7

megahirt:
This PR is to help me see the changes that will ship with Language Forge v1.5.10. I will merge this myself after adequate testing.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/618)


megahirt:
Tested the following on QA:

- Confirmed audio gain is now at an acceptable level
- User admin page is fixed with username field

I have struggled to get working E2E tests running (they don't run well enough in my VM) so I will continue working toward that, but for now am fine to ship without E2E test confirmation.  We've waited long enough for these fixes.

PR #622, opened 2019-03-28
SF-132 Simplify DateTime to object, for successful Unset to MongoRepo…

marksvc:
…sitory

Add Exception to MemoryRepository builder to catch these problems in
tests.
More information in SF-132.

---
So in MemoryUpdateBuild.cs I have a rather limited list, which could trigger problems as soon as someone uses it for a new purpose (though that might be good, depending on how Mongo would behave). I did not investigate why String works and DateTime does not. But certainly DateTime was causing problems. 
An alternate solution might be to modify MongoUpdateBuilder.cs to conform arguments to Unset to contain acceptable data, but then that adds divergence between Mongo and Memory repositories.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/622)

### DIFF
--- a/src/SIL.XForge.Identity/Controllers/IdentityRpcController.cs
+++ b/src/SIL.XForge.Identity/Controllers/IdentityRpcController.cs
@@ -22,6 +22,8 @@ using SIL.XForge.Utils;
 
 namespace SIL.XForge.Identity.Controllers
 {
+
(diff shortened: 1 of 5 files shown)


irahopkinson:
This PR has been around for a month with no movement. @marksvc is this PR and branch still relevant?

irahopkinson:
not fixing on this identity server.

PR #625, opened 2019-03-29
SF-234 Add log in to Paratext button

irahopkinson:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/625)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.html
@@ -30,7 +30,19 @@
               Collaborate with other translators working on the same text and use translation suggestions.
             </p>
           </div>
-          <ng-container *ngIf="translate">
+          <div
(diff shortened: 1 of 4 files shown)


jasonleenaylor:
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.ts, line 42 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/625#-Lb8Rl1H0RGuEc40u5d4:-Lb8Rl1H0RGuEc40u5d5:b-j9g871) ([raw file](https://github.com/sillsdev/web-languageforge/blob/7cc1a714eb63277efb8f0c511061372e5bc42e34/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.ts#L42)):*
> ```XML
>   /** Elements in this component and their states. */
>   private controlStates = new Map<string, ElementState>();
>   private isFirstLoad: boolean = true;
> ```

believe it or not I think a comment on this boolean would be helpful. I had to read through a lot of code to be certain what the 'first' load actually meant.

PR #646, opened 2019-04-04
[LFv1] .htaccess mod_headers to allow proper service worker scoping

megahirt:
This is a small change that adds the following to the .htaccess file.

`Header Set Service-Worker-allowed "/"`       

This change resolves the console errors we are seeing on live currently

image

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/646)

### DIFF
--- a/src/.htaccess
+++ b/src/.htaccess
@@ -6,3 +6,9 @@
     RewriteCond %{REQUEST_FILENAME} !-f
     RewriteRule ^ index.php [QSA,L]
 </IfModule>
+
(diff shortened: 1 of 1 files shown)


megahirt:
After chatting with @irahopkinson about this change, we've decided to not implement this via an .htaccess file and to instead make the change in the apache conf file on the server itself.  These changes will (in theory) be reflected on our new Ansible scripts, forthcoming.

Closing without merging.

PR #647, opened 2019-04-04
[SFv1] Add project-level font size and RTL script options.  This completes the RTL fix from earlier.

megahirt:
This adds two new project properties which are editable by managers on the project properties tab:

image

The font size property controls the font size for all texts in the project, and controls the font size for the left-hand text column (not user-selected quotes).

The RTL property specifically sets the dir="rtl" for the text columns in both the text view and the question/comment view, and also sets dir="rtl" for the scripture quotation divs when user's include a quote in their answer.

image

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/647)


megahirt:
> Totally a nit pick, but shouldn't it be "all texts use" or "all text uses"?

Yes, good call.  I changed it to "texts" in both places that I saw.

PR #658, opened 2019-04-13
Release microphone when done recording (LFv1)

Nateowami:
My original plan was to hang on to the microphone until the recording component was closed. However, after experimenting a bit I found that hanging onto it after the recording stops has no advantages, and was even slightly more complicated to implement.

At present the audio recording indicators of Chrome and Firefox (only tested browsers) only show up when the user is actually recording an audio segment.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/658)


Nateowami:
That is correct.

PR #660, opened 2019-04-15
SF-233 Handle changing mother translation

marksvc:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/660)


marksvc:
I hadn't changed my test to use a MemoryRepository instead of a mock. And I've set various class methods as virtual or public so I can access them from my test. In FW we used to use InternalsVisibleTo to reveal non-public information to tests if helpful, tho it doesn't look like that's been being done in SF.

marksvc:
This is only part of what needs done for SF-233.

PR #665, opened 2019-04-20
Add migration to fix double language tags on comment context guid (LFv1)

Nateowami:
This migration takes comments with a redundant context guid:
```
sense#05f3fb35-e855-47fc-bef7-478f2a7c6dae example#dc996a25-9eab-4684-8fdc-ddadc2e49a5e sentence.en.en
```
and removes one of the language tags:
```
sense#05f3fb35-e855-47fc-bef7-478f2a7c6dae example#dc996a25-9eab-4684-8fdc-ddadc2e49a5e sentence.en
```
I based it off of an existing migration that changed the context guids to use the language tag rather than the language abbreviation.

The reason I added on to another migration is because they use a lot of the same boilerplate, and they're both idempotent. It was an easy way of making sure this was going to work without reimplementing a lot of stuff again. It also makes it easier for reviewers to see exactly what I've added. I'm guessing I should split this out into its own migration? If so I will probably do it by just copying the migration and removing the parts that only apply to the original migration.

Running this migration has the desired effect of making all comments show again, but only after IndexedDB is refreshed, which doesn't happen automatically. In fact, the only way I was able to get it to refresh the data was by clearing IndexedDB in the developer tools. Do we have a way of forcing the data to refresh?

This is a companion pull request to #664.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/665)

### DIFF
--- a/src/Api/Library/Shared/Script/Migration/FixCommentGuid.php
+++ b/src/Api/Library/Shared/Script/Migration/FixCommentGuid.php
@@ -26,7 +26,8 @@ class FixCommentGuid
         $projectList = new ProjectListModel();
         $projectList->read();
         $projectCount = 0;
-        $fixCommentCount = 0;
+        $fixDoubleInputSystemTagCount = 0;
+        $changeLanguageAbbreviationToTagCount = 0;
         foreach ($projectList->entries as $projectParams) { // foreach existing project
             $projectId = $projectParams['id'];
             $project = ProjectModel::getById($projectId);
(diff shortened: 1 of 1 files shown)


megahirt:
Yes, we have a way of forcing a client refresh of the Indexeddb.  We
increment the version number of the client db.  I'm on a phone this weekend
so I don't know in which file the version number is located.  Probably
offline cache or something similar.

Great job on these PRs!  I will review on Monday.  I'm very encouraged by
your progress.

If the migration script accomplishes the task without hurting anything,
then I will accept it as is, even if it does unnecessary processing for a
previous migration.

Kind regards,

Chris

megahirt:
[LF-197 Migration Script](https://trello.com/c/qxNaP6ys/352-lf-197-migration-script)

megahirt:
oops.  Upon further inspection, it looks like with the change to localforage (not indexeddb specific) we lost the ability to drop the DB on version dump.  If we want that feature back, we're going to have to code it.  But I'm thinking that all is not lost.  The migration script should change the modified date of the comment, and so that should get picked up and the update should get pulled down to the client automatically.

Nateowami:
I just ran the migration script to check, and you are indeed correct that `dateModified` is updated when the migration runs. However, the comments were not automatically updated for me (I had to clear IndexedDB). I think it may take a bit of testing to figure out why. Do you know where the code is that triggers the update of client-side data?

PR #679, opened 2019-04-26
Enable sourcemaps in production build

Nateowami:
I believe this is all that is required to enable sourcemaps in production, assuming the command for building the TypeScript for production is `gulp build-webpack --applicationName languageforge --doNoCompression false`.

In addition to helping us locate the source of bugs, I'm guessing it will also help Bugsnag to correctly group errors, which it is currently doing quite poorly.

When the sources are uploaded to production the .map files need to go along with them. I don't see them listed in either upload-exclude.txt or upload-include.txt, which I am guessing are used in the deploy process.

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/679)

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -142,7 +142,7 @@ module.exports = function (env) {
       'process.env.NODE_ENV': JSON.stringify('production')
     }));
     webpackConfig.plugins.push(new UglifyJsPlugin({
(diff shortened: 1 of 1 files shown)


Nateowami:
[Enable source maps on Bugsnag](https://trello.com/c/FlH9eN91/360-enable-source-maps-on-bugsnag)

PR #680, opened 2019-04-27
Fix console error that occurs when going offline

Nateowami:
I first noticed this error locally and then also recognized it as one of the errors on Bugsnag. (It was causing `Cannot set property 'textContent' of null`).

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/680)


megahirt:
[Cannot set property 'textContent' of null (console error when going offline)](https://trello.com/c/8XNud8Mq/357-cannot-set-property-textcontent-of-null-console-error-when-going-offline)

PR #683, opened 2019-04-30
Fix Angular unit tests on CI build

ddaspit:
- turn off source maps by default
- upgrade Karma and Jasmine to the latest
- set random seed on CI build

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/683)


src/SIL.XForge.Scripture/ClientApp/angular.json
```diff
@@ -106,7 +106,8 @@
               },
               "src/styles.scss"
             ],
-            "assets": ["src/favicon.ico", "src/assets", "src/manifest.json"]
+            "assets": ["src/favicon.ico", "src/assets", "src/manifest.json"],
+            "sourceMap": false
```
megahirt: Is there a reason you turn off sourceMaps by default?  Can they do any harm?  We had them off for the longest time in xForge v1 for production and just turned them on to aid in debugging.  I didn't see any harm in turning them...

PR #689, opened 2019-05-17
add HelpHero JS snippet

megahirt:
This adds support for the HelpHero tutorial walk throughs described at https://app.helphero.co/install?type=async-html

This is the HTML snippet method.  If we had a single page app for the entire site, then we would use the npm install method.  We could still use npm install and link locally, however linking to their CDN is a good first step.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/689)

### DIFF
--- a/src/Site/Controller/Base.php
+++ b/src/Site/Controller/Base.php
@@ -162,6 +162,7 @@ class Base
                 $this->data['isAdmin'] = SystemRoles::hasRight($this->_user->role, Domain::USERS + Operation::CREATE);
             }
             $this->data['userName'] = $this->_user->username;
+            $this->data['userId'] = $this->_userId;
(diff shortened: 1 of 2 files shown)


megahirt:
[install HelpHero JS snippet for SFv1](https://trello.com/c/xddlHKvp/363-install-helphero-js-snippet)

Nateowami:
Is this intended to be functional, or just adding the boilerplate code? I've tried it out and don't see anything happening, and looking at the docs it looks like more setup than this is required. I'm just a bit confused about what this PR is attempting to change.

megahirt:
@Nateowami thanks for that.  This is just adding the required JS snippet so that when it hits scriptureforge.org the help tutorial that I've already written will be active.  The only way to see it before it's live is to use their Chrome extension and sign in to helphero as a collaborator.

PR #691, opened 2019-05-29
Fix message when changing role of invited user

Nateowami:
Fixes the message shown when changing an invited user's role on a project.

Old message: "undefined's role was changed to observer_with_comment"

New message: "user@example.com's role was changed to Observer with comment."

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/691)


Nateowami:
[Fix label of notice when changing role of invited user in user management](https://trello.com/c/AZEBAdUL/356-fix-label-of-notice-when-changing-role-of-invited-user-in-user-management)

PR #699, opened 2019-05-31
Always return a value in processEditorDto

rmunn:
Fixes LF-274

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/699)

### DIFF
--- a/src/angular-app/bellows/core/offline/editor-data.service.ts
+++ b/src/angular-app/bellows/core/offline/editor-data.service.ts
@@ -229,6 +229,9 @@ export class EditorDataService {
       this.commentService.updateGlobalCommentCounts();
       deferred.resolve(result);
       return deferred.promise;
+    } else {  // if (result.ok)
+      deferred.reject(result);
+      return deferred.promise;
     }
   }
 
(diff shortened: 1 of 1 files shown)


rmunn:
> I'm assuming this fixes the problem (sounds like it should). Since you want this method to always return, then delete the `else` and move the 2 lines outside the `if`. That would make it clearer your intention to always return the promise and would also not duplicate code.

I considered that, and decided that deleting the `else` would make it *less* clear. E.g., I think that this:

```
if (someCondition) {
    ...
    deferred.resolve();
    return deferred.promise;
} else {
    deferred.reject();
    return deferred.promise;
}
```

clearly communicates that if the condition fails, the promise is rejected. If the condition is long, as it is here, then a comment on the `else` block helps remind you what's being tested, and thus it's completely obvious *why* the promise is being rejected here. Whereas this:

```
if (someCondition) {
    ...
    deferred.resolve();
    return deferred.promise;
}
deferred.reject();
return deferred.promise;
```

never actually makes it explicit why the promise is being rejected. You have to read back up a few lines and notice that the `if` condition leads to a `return`, and *then* you understand that the condition is "If you reach this point, the condition failed".

In other words, I've come to dislike treating `return` as having side effects. Perhaps that's an artifact of having spent quite a bit of time writing F# code, where you're *not allowed* to break out of a function early, and you therefore *have* to write `if ... else` blocks like the one I wrote here.

How about this? I think the following would satisfy both your code preferences and mine:

```
if (someCondition) {
    ...
    deferred.resolve();
} else {
    deferred.reject();
}
return deferred.promise;
```

In other words, move the `return` out of the `if...else` and then there's only a single `return` point in the function, making it totally obvious that the function will return a promise no matter what.

PR #701, opened 2019-06-03
Fix blank lexicon entries on first load

rmunn:
This fixes a race condition where the project data might finish loading before the project config is loaded, which results in it looking empty because the lexicon-entry renderer depends on the project config being populated. Now we load the initial project data only after the config is completely loaded, which should ensure that it's always ready to render as soon as it loads.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/701)


rmunn:
Didn't realize I had PR #699 checked out when I started work on this one, but that's a related bug so it should probably be merged first anyway.

rmunn:
Rebased onto master so that it will merge cleanly.

PR #705, opened 2019-06-07
Ignore Javascript errors from outdated browsers

rmunn:
We have one device with an outdated browser (Chrome Mobile 30) that's sending a lot of Bugsnag reports about Javascript features that that browser doesn't implement (e.g., the Array.includes function). These reports aren't actually bugs and we won't do anything about them, so we'll ignore them and not send them to Bugsnag.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/705)

### DIFF
--- a/src/angular-app/bellows/core/exception-handling.service.ts
+++ b/src/angular-app/bellows/core/exception-handling.service.ts
@@ -56,6 +56,15 @@ export class ExceptionHandlingService {
       return;
     }
 
+    // Silence errors coming from out-of-date browser versions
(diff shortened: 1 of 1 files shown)


rmunn:
Closing this PR pending further discussion on Slack: I think the right thing might be to tweak our Typescript compilation options so that older browsers get served Javascript they can understand. So this probably isn't the right approach, and shouldn't be merged until we decide on an approach.

megahirt:
Agreed - I think there is probably a better way than adding browser
specific code exceptions.  Thanks.

PR #706, opened 2019-06-10
Fix activity log not showing in Safari

rmunn:
Safari parses timestamps strictly according to [RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6), which requires a colon in the timezone (`"+07:00"`, not `"+0700"`). The default ISO8601 format codes in PHP use `"+0700"` format, so we need to change how we encode dates in JSON so that Safari (and other strictly-compliant browsers, like Safari Mobile) accept the dates the server returns.

Tested in BrowserStack and confirmed to work: with this change, the activity log shows up properly in Safari.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/706)

### DIFF
--- a/src/Api/Model/Shared/Mapper/JsonEncoder.php
+++ b/src/Api/Model/Shared/Mapper/JsonEncoder.php
@@ -189,7 +189,9 @@ class JsonEncoder
      */
     public function encodeDateTime($model)
     {
-        return $model->format(\DateTime::ISO8601);
+        // Can't use DateTime::ISO8601 here since it returns timezones like "+0700" (format code "O") and we need "+07:00" (format code "P")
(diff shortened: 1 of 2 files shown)


rmunn:
Darn. Looking at the `php-jiffy` code, it looked like it would be able to parse timestamps in `+07:00` format, but apparently it can't. I'll have to submit a bug report to their issue tracker to fix that, since that's the format required by RFC 3339 (and by the Safari browser), but this bugfix isn't going to be as simple as I thought.

rmunn:
I spoke too soon. `php-jiffy` can parse that format just fine, as long as I get the generating code right.

megahirt:
Squashed and merged

PR #707, opened 2019-06-10
Report bugs to Bugsnag even if username is null

rmunn:
Should fix LF-222, and possibly reveal some bug reports that we'd been missing before (because when `this.metadata` was non-null but `this.metadata.userName` was null, the original exception was swallowed).

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/707)

### DIFF
--- a/src/angular-app/bellows/core/exception-handling.service.ts
+++ b/src/angular-app/bellows/core/exception-handling.service.ts
@@ -51,7 +51,7 @@ export class ExceptionHandlingService {
   }
 
   private notifyBugsnag(exception: Error, cause?: string) {
-    if (this.metadata != null && this.metadata.userName.startsWith('test_runner_')) {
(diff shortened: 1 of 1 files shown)


megahirt:
[/public/signup Uncaught TypeError: Cannot read property 'startsWith' of undefined](https://trello.com/c/ZKvm3Jhx/353-public-signup-uncaught-typeerror-cannot-read-property-startswith-of-undefined)

rmunn:
When we do merge this, I suggest a "Squash and merge" so we don't end up with the commit-and-revert-to-original in the Git history. The revert went back to the state of the original commit, so this can be merged down to a single commit without losing any valuable history whatsoever.

megahirt:
I did a squash and merge to reduce commit churn.

PR #711, opened 2019-06-17
Fix input systems configuration

rmunn:
No model change needed with this logic, only viewmodel changes. In the viewmodel, `hasCustomInputSystemsOverride` becomes an `InputSystemSettings` type so that it has a set of role and group booleans instead of being a single boolean, and it is used to keep track of whether the user has ever changed any of the checkboxes. If so, an override is automatically assumed and set as such in the config. But there's also a Reset button which lets the user go back to a "No override" status.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/711)

### DIFF
--- a/src/angular-app/languageforge/lexicon/settings/configuration/configuration-fields.component.html
+++ b/src/angular-app/languageforge/lexicon/settings/configuration/configuration-fields.component.html
@@ -74,32 +74,32 @@
                 <td class="text-center align-middle">
                     <input class="position-static observer-checkbox"
                            data-ng-model="inputSystem.observer"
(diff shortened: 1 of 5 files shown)


rmunn:
@irahopkinson I'm committing my work on the E2E tests on the [`wip/fix-e2e-input-systems` branch](https://github.com/sillsdev/web-languageforge/tree/wip/fix-e2e-input-systems). I plan to squash it down to a single commit before I merge, which I'll then put in a branch named `fix-e2e-input-systems` without the `wip/` tag. If you feel like following my work as I do it, you can look there, but there's no need if you're focused on something else.

rmunn:
E2E fixes are now in https://github.com/sillsdev/web-languageforge/pull/716.

PR #714, opened 2019-06-18
Ubuntu install script sets group write permissions on src/assets

Andrewdt97:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/714)

### DIFF
--- a/installer/ubuntu1604Installer.sh
+++ b/installer/ubuntu1604Installer.sh
@@ -119,6 +119,9 @@ cd scripts/tools
 sudo php FactoryReset.php run || exit
 cd ../..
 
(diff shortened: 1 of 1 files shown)


installer/ubuntu1604Installer.sh
```diff
@@ -119,6 +119,9 @@ cd scripts/tools
 sudo php FactoryReset.php run || exit
 cd ../..
 
+echo "Set file permissions"
+sudo chmod g+w src/assets/lexicon/
```
Andrewdt97: Actually, is there a `-R` required here that was missed.

PR #717, opened 2019-06-26
Attempt to fix "empty items in project view" bug

rmunn:
I haven't been able to reproduce the "initial load of dictionary list view shows empty list" issue since https://github.com/sillsdev/web-languageforge/pull/701 was merged, but this is one thing that *might* be causing it, and there's no reason for not resetting the visible items every time `sortAndFilterEntries` is called. So this is worth a try.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/717)

### DIFF
--- a/src/angular-app/bellows/core/offline/editor-data.service.ts
+++ b/src/angular-app/bellows/core/offline/editor-data.service.ts
@@ -385,7 +385,7 @@ export class EditorDataService {
             deferred.resolve(newResult);
           });
         } else {
-          this.sortAndFilterEntries(false).then(() => {
+          this.sortAndFilterEntries(true).then(() => {
             deferred.resolve(result);
           });
         }


rmunn:
I deployed this by hand to QA and it did not fix the bug. This is still worth merging, though, because there's no reason to call `sortAndFilterEntries(false)`.

megahirt:
I cannot merge this at this point because there's not enough information for me to determine if your statement 

> "there's no reason to call sortAndFilterEntries(false)"

 is correct or not.

Additionally @feliks-montez tested your commit specifically and it did not fix the problem.

I will leave this PR open for 2 more weeks to allow ample time for an explanation.  Otherwise I will close it.

megahirt:
Closing PR due to inactivity and previous comment

PR #727, opened 2019-07-10
Disable systemd /tmp isolation for dev machines

feliks-montez:
# Overview
Ubuntu 18.04 systemd uses /tmp isolation by default, which means every process gets its own private /tmp folder. Since our php e2e tests run in two processes (a setup script and then the actual tests), the actual tests cannot access the needed tempfiles when /tmp isolation is enabled, causing tests--like file upload tests--to always fail.

Disabling this feature is less secure, but since we only ever need to run these tests on dev machines and never on the production box, this is a safe solution.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/727)

### DIFF
--- a/deploy/playbook_bionic.yml
+++ b/deploy/playbook_bionic.yml
@@ -137,6 +137,18 @@
           value: '60M'
         - property: 'upload_max_filesize'
           value: '60M'
+    - name: copy apache2 service config from /lib to /etc for modification
+      copy:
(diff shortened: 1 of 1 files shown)


feliks-montez:
[Get E2E tests running on 18.04 dev machines](https://trello.com/c/SRqo2IWy/372-get-e2e-tests-running-on-1804-dev-machines)

PR #730, opened 2019-07-11
Run PHP unit tests at end of ansible install script

feliks-montez:
# Overview
Wouldn't it be convenient if, after installing your dev env, it automatically checked to make sure everything worked correctly? I added a `gulp test-php` shell command to the end of the ansible script to accomplish exactly that.

We recently discovered that running the php unit tests for the first time as the regular user creates a tmp directory with the wrong permissions. To fix this, I have made the php unit tests run as the www-data user. This adjustment should make sure the issue never happens.

I also added a configuration change in `ansible.cfg` so that errors (like from the PHP unit tests) get pretty printed instead of all lumped together with newlines escaped as `\n`'s.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/730)


feliks-montez:
[ansible script should run `gulp test-php` as a final step after refreshDeps.sh](https://trello.com/c/4w3MXlgs/389-ansible-script-should-run-gulp-test-php-as-a-final-step-after-refreshdepssh)

feliks-montez:
[Make php tests run as user www-data](https://trello.com/c/m4NCRJTi/393-make-php-tests-run-as-user-www-data)

PR #734, opened 2019-07-11
Change to use languageforge.localhost

megahirt:
The `.localhost` TLD is recognized by browsers and systems for localhost use while our use of `.local` is not.  This change across the entire code-base changes all references from `.local` to `.localhost`

Once this PR is merged, ***developers may need to run ansible and FactoryReset.php again*** before the site is operational again.

The PR also removes references to:
dev.languageforge.local
dev.scriptureforge.local
demo.*

as these are no longer in use.

The site will now be accessed at http://languageforge.localhost

All PHP Tests are passing and all but the 3 E2E tests that are known to fail are passing.

image

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/734)

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -6,6 +6,7 @@
     "msjsdiag.debugger-for-chrome",
     "EditorConfig.editorconfig",
     "felixfbecker.php-pack",
-    "ms-vscode.vscode-typescript-tslint-plugin"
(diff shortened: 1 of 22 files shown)


megahirt:
[use languageforge.localhost instead of languageforge.local](https://trello.com/c/1oZLug6a/387-use-languageforgelocalhost-instead-of-languageforgelocal)

PR #737, opened 2019-07-12
Skip PECL Mongo compile if already installed

feliks-montez:
# Overview
When running the Ansible install script after it has already been initially run, it still builds PECL Mongo, which takes an exorbitant amount of time to complete. This PR adds a check that will skip the PECL Mongo install if it is already present.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/737)

### DIFF
--- a/deploy/playbook_bionic.yml
+++ b/deploy/playbook_bionic.yml
@@ -279,10 +279,15 @@
         group: www-data
         mode: 0755
 
+    - name: check for PECL Mongo Extension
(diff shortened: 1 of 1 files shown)


feliks-montez:
[Reduce wait time for Install PECL Mongo in Ansible install script](https://trello.com/c/jXGgd3O2/392-reduce-wait-time-for-install-pecl-mongo-in-ansible-install-script)

PR #746, opened 2019-07-25
Tech Support Role

Andrewdt97:
**Problem:**
In order to troubleshoot, devs are often added to projects in order to solve issues. When accessing projects, admins are added to those projects as Managers. This can create confusion.

**Solution:**
To solve this problem a Tech Support user was implemented that is effectively a clone of the manager role. It is implemented throughout the front and back ends. On the user management app, a Tech Support user is the only one who can change their own role, it is locked to anyone else. Further, there is no way for a user to assign another user to Tech Support. Currently the only way to become Tech Support is to click the "+ Tech Support" button as an admin from the projects app or attempt to access a project you are not of member of whilst being logged-in as an admin. The role is defined within the `RoleBase` class, which means it exists in SF as well, though it lacks permissions there.  There is no current support for the front-end knowing the admin status of a user which is why an admin cannot change the role of other Tech Support users.

**Tests:**
E2E and PHPUnit tests were written and all tests are passing. Being relatively new to the LF system, a manual test from the reviewer that ensures a Tech Support user can do everything a Manager can may provide useful.

**Screenshots:**
image
*Tech Support user looking at the User Management app.*

___

image
*User Management app page from the project owner's perspective.*

[Trello Card](https://trello.com/c/nXZeK5E7/382-techsupport-project-role)

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/746)

### DIFF
--- a/src/Api/Library/Shared/SilexSessionHelper.php
+++ b/src/Api/Library/Shared/SilexSessionHelper.php
@@ -70,7 +70,7 @@ class SilexSessionHelper
             // Add an admin to the project if they are not already a member
             if ($user->role == SystemRoles::SYSTEM_ADMIN) {
                 if (!$project->userIsMember($userId)) {
-                    $project->addUser($userId, ProjectRoles::MANAGER);
+                    $project->addUser($userId, ProjectRoles::TECH_SUPPORT);
                     $projectId = $project->write();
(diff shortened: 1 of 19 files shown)


megahirt:
Thanks Andrew!  Can you paste in a screenshot?

PR #753, opened 2019-08-02
Lexicon search/filter

Nateowami:
The improved lexicon search/filter is mostly complete and all e2e tests are passing.

Description | Before | After
-------|------|-----
When all entries are being shown | Screenshot from 2019-08-02 15-03-17 | Screenshot from 2019-08-02 15-01-36
When searching for entries | Screenshot from 2019-08-02 15-04-55 | Screenshot from 2019-08-02 15-07-49
Filtering options || Screenshot from 2019-08-02 15-09-57

Notes:
- This is not yet complete. The UI in particular is not yet what we decided on. I wanted to get the substantive changes completed first and not get too far behind master.
- The functionality on the editor list page is identical and the UI almost identical.
- The "Default" sort option changes meaning based on whether or not the search box contains search terms. When there are no search terms, the label is "Default (Word)", which sorts the entries alphabetically by lexeme. When there is a search term, they are sorted by relevance and then alphabetically by lexeme. The label in this case changes to "Default (Relevance)".
- The "Reset Filter" button resets the filter text, any filter options, and the sort order.
- While some code was reused from the typeahead, I found that the code for sorting by relevance was incorrect, so it was rewritten. When sorting by relevance, entries with a lexeme that matches the search term at the very beginning come first, then those that match anywhere in the lexeme, then those that match at the beginning of a gloss, then those that match anywhere in a gloss, then all other entries with matches anywhere. I'm not at all confident that this is the correct solution, but it's what the typeahead was trying to do. Even if this is seen as an approximately correct sort order, I think the citation form ought to be included.
- It's worth noting that the filter code searches all string properties of an entry, except those on a blacklist. This includes part of speech and semantic domains (though they would have to be searched by number).
- The search term is persisted in the URL.
- @megahirt and I discussed wording above the list of entries. We didn't come to an exact conclusion on what it should be, but decided it should take up less space than currently, and it should be able to handle five-digit numbers without wrapping text.

I've merged master into my feature branch and resolved conflicts.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/753)

### DIFF
--- a/rune2e.sh
+++ b/rune2e.sh
@@ -27,8 +27,8 @@ attach_debugger () {
 
   PID=""
   while [ "$PID" = "" ]; do
-    PID=$(pgrep -f protractor/built/cli.js)
     sleep 0.25;
+    PID=$(pgrep -f protractor/built/cli.js)
   done
 
   # See https://nodejs.org/en/docs/guides/debugging-getting-started/
--- a/src/angular-app/bellows/core/offline/editor-data.service.ts
+++ b/src/angular-app/bellows/core/offline/editor-data.service.ts
@@ -619,6 +619,7 @@ export class EditorDataService {
   }
 
   private sortList(config: LexiconConfig, list: LexEntry[]): any[] {
+    const reverse = this.entryListModifiers.sortReverse;
     if (this.entryListModifiers.sortBy.value === 'default' && this.entryListModifiers.filterText() !== '') {
 
       // each of the five slots contain results of varying relevance:
@@ -654,14 +655,15 @@ export class EditorDataService {
         prioritizedResults[4].push(entry);
       }
 
+      if (reverse) prioritizedResults.reverse();
       // sort the individual lists and then flatten the results
-      return prioritizedResults.map(section => this.sortListAlphabetically(config, section))
+      return prioritizedResults.map(section => this.sortListAlphabetically(config, section, reverse))
                                 .reduce((main, sub) => main.concat(sub), []);
 
-    } else return this.sortListAlphabetically(config, list);
+    } else return this.sortListAlphabetically(config, list, reverse);
   }
 
-  private sortListAlphabetically(config: LexiconConfig, list: LexEntry[]): LexEntry[] {
+  private sortListAlphabetically(config: LexiconConfig, list: LexEntry[], reverse: boolean): LexEntry[] {
     const inputSystem = this.getInputSystemForSort(config);
     const compare = ('Intl' in window) ? Intl.Collator(inputSystem).compare : (a: string, b: string) => a < b ? -1 : 1;
 
@@ -670,7 +672,7 @@ export class EditorDataService {
       value: this.getSortableValue(config, entry)
     }));
 
-    mapped.sort((a: any, b: any) => compare(a.value, b.value) * (this.entryListModifiers.sortReverse ? -1 : 1));
+    mapped.sort((a: any, b: any) => compare(a.value, b.value) * (reverse ? -1 : 1));
 
     return mapped.map((el: any) => list[el.index]);
   }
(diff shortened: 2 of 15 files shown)


Nateowami:
@megahirt One thing I forgot to address was your request to make the options panel not be closable if settings are not at their default. I'm not sure exactly what that would look like. Just do nothing when the user clicks to close the options? That would feel like a bug to me. Perhaps that's not exactly what you were thinking of.

The only other idea I had was to show an indicator if options have been closed while some options are still active:
Screen Shot 2019-08-05 at 12 19 48

This could be a chance to address the use of localStorage to remember whether the options panel was open or closed:
``` ts
this.$window.localStorage.getItem('viewFilter') === 'false');
```
We could instead choose to open it on page load only if the options are not at their default values.

Nateowami:
- I've added the button to clear the search text
- The filter and sort options are now side-by-side only on larger screens, and only on the editor list page. The rest of the time filter options are above sort options. This is done with a media query and `flex-direction`
- The `+` button is slightly narrower now, making it a square. In the previous screenshots it's a bit wide for its height. 

Screen Shot 2019-08-05 at 14 52 47

megahirt:
[Improved lexicon search/filter](https://trello.com/c/rfww2MTy/376-improved-lexicon-search-filter)

Nateowami:
@megahirt I've made most of the changes, but I haven't implemented the semantic domain search yet. Unfortunately I ran into some permission issues today with the PHP tests, made them far worse, and ended up blowing away my whole repo. Fortunately the new one was up and running pretty quickly. I also need to make sure it's skipping input systems the user isn't allowed to access. I think as of right now that is not the case.

Nateowami:
@megahirt I was really hoping this would be finished today, but there are still one or two outstanding issues. Semantic domains are now handled correctly and I'm quite happy with how that works. 

The remaining issues are with input systems and hidden fields. I'm trying to make sure it doesn't search fields the user isn't supposed to see, or search input systems the user doesn't have access to. I thought I had mostly figured out how to do this, using the config as a reflection of each entry (i.e. you can iterate the config to find out what fields _should_ be displayed, and simultaneously iterate the entry to find that data that is actually present).

Right now I am seeing some discrepancies in my test project. I think this is due to three sources of information that have to be reconciled: Input systems and fields that _exist_, input systems and fields that are _supposed_ to exist (a subset of those that do exist), and input systems and fields that the user has _permission_ to access. So, for example. if I disable a French input for glosses, any data in that field still exists and is shown in the UI. However, the filter is not finding it, because it's using the list of input systems that are supposed to exist for that field.

I've had some cases though where the filter was searching data that was not displayed in the view. When I tried to look into some of this further, I noticed that the config information being used in the view was different from what the filter was getting when I stopped it in the debugger. The list of input systems had length 7 in the debugger, and 6 in the view. The extra input system was the exact one where the data I wasn't supposed to be searching was contained. While I need to look into this further, I'm suspicious that two versions of the config exist, perhaps one of them being loaded fresh from the server, and the other being cached locally. Unfortunately I don't have much figured out yet.

megahirt:
>While I need to look into this further, I'm suspicious that two versions of the config exist, perhaps one of them being loaded fresh from the server, and the other being cached locally. Unfortunately I don't have much figured out yet.

AFAIK there is only one config and it is never cached locally.  It is "cached" in memory in the config service I think but that is just to prevent unnecessary network requests.

I feel like the unified config view + user specific permissions for a certain input system or field makes for a very complex model, and we've done a poor job and keeping a good data model around this so I'm not surprised that what you're trying to do is hard.

Keep persevering, but also know when it's time to say it's good enough.  I will accept a search that mostly works and mostly respects the user/field permissions as best as you can get it acknowledging the fact that the underlying model may be overly complex and needs refactoring anyway.

megahirt:
Also, we need to rebase this on latest master before merging.

Nateowami:
I just discovered the comment and activity ui was broken (despite e2e tests passing). After a quick bisect I see I did this in `editor-entry.view.html` about 27 commits ago:
``` diff
                 <div class="word-definition-title">
-                    <h3>Entry Preview</h3>
                     <dc-rendered global-config="$ctrl.lecConfig" config="$ctrl.lecConfig.entry" model="$ctrl.currentEntry" option-lists="$ctrl.lecOptionLists"></dc-rendered>
-                </div>
```
The closing div tag should not have been removed. This only took a few minutes to find, but probably would have taken me hours without `git bisect`. I've just force-pushed a fix.

As far as I am aware, this PR is now mergeable.

PR #754, opened 2019-08-05
Fix one reason for "Cannot read property 'data' of undefined" error

rmunn:
One of our `null` tests was one line *after* where it needed to be.

Should fix https://jira.sil.org/browse/LF-318

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/754)


megahirt:
[Cannot read property "data" of undefined](https://trello.com/c/3CYE9k2w/423-cannot-read-property-data-of-undefined)

PR #758, opened 2019-08-07
Display messages from url in angular

Andrewdt97:
There is currently no way for the back-end to pass messages to angular's `notice.service`. This PR creates `noticeService.checkUrlForNotices()` which is designed to be called from the `onInit()` of a controller and be passed `$location.search()`. It displays base64 decoded strings from the URL query parameters: `errorMessage`, `infoMessage`, and `successMessage`. The following screenshot shows functionality, but is taken from another branch where the method is implemented.

The function takes an `any` argument because `$location.search()` returns a basic JSON object. Is it worth it to parse that into a type to before passing into `checkUrlForNotices`?

image

Developer's Note: This seemed like a separate piece from what I have been working on, so I put it in it's own PR. If you would prefer to just lump it in with the rest of my back end work for link-sharing, I can.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/758)

### DIFF
--- a/src/angular-app/bellows/apps/projects/projects-app.component.ts
+++ b/src/angular-app/bellows/apps/projects/projects-app.component.ts
@@ -60,6 +60,8 @@ export class ProjectsAppController implements angular.IController {
         this.helpHeroService.anonymous();
       }
     });
+
+    this.notice.checkUrlForNotices();
   }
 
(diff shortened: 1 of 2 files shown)


Andrewdt97:
This is included in #759

PR #762, opened 2019-08-13
Consistent, translatable login buttons for Google and Facebook

rmunn:
The login buttons for Google and Facebook now have a consistent look and are compliant with each company's brand-use guidelines (e.g., we use the right shade of blue for each button). As a bonus, we also create the buttons with HTML and CSS instead of PNG images, which means their text can be translated.

Also, I noticed that Facebook avatar URLs were being incorrectly handled in the previous Facebook OAuth code, so I've fixed that. We now produce correct URLs for people's Facebook profile picture, and use that as our default avatar if someone signs up via Facebook OAuth (just like we do for Google OAuth).

Screenshot:

login-with-facebook

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/762)

### DIFF
--- a/src/Site/OAuth/FacebookOAuth.php
+++ b/src/Site/OAuth/FacebookOAuth.php
@@ -24,8 +24,22 @@ class FacebookOAuth extends OAuthBase
 
     public function getFullSizeAvatarUrl(string $avatarUrl)
     {
-        // TODO: Learn what Facebook avatar URLs look like
-        return $avatarUrl;
+        // Facebook avatar URL returned from OAuth looks like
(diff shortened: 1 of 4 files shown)


Nateowami:
@rmunn Were the images added to the repo? They're not loading, and I'm not finding them anywhere in the repo either.

Screenshot from 2019-08-14 10-14-34

megahirt:
@rmunn oops it looks like you didn't check in the svg files referenced e.g. /Site/views/shared/image/google-logo.svg and I missed that in the review.  Can you please add svg files in a separate PR?

Nateowami:
@rmunn I also just noticed the sign up page still only has the Google icon, and uses the image instead of the new login buttons.

PR #765, opened 2019-08-15
Display warning message for Internet Explorer

rmunn:
We have seen many Javascript errors in Bugsnag caused by IE 11 not being up-to-date with modern Javascript features. `Array.includes` is one of them, which was recently fixed by https://github.com/sillsdev/web-languageforge/pull/755, but there are others: IE 11 doesn't support `String.prototype.normalize()`, for example. We have two choices: we could add polyfills for all the different things that IE 11 doesn't support, which would be chasing a moving target since we never test with IE. Or, since we have already decided that we don't support IE (since Edge is available), we can just inform the user that Internet Explorer is not supported and ask them to use a better browser.

This PR makes the latter choice. We should probably have a discussion about whether that's the right choice before merging it. We can also discuss the wording of the error message we want to present.

To test this on a Linux box where Internet Explorer isn't available, you can edit `Base.php` and change the `isIE` regex to look for the strings "Chrome" or "Firefox" instead of "MSIE" or "Trident". This will let you experience what an IE-using user would see.

Sample of what the warning looks like below; note there is no "close" box. This is intentional, so that the warning shows up on every Angular page until they switch to a better browser:

ie-warning

And how the warning looks on the login page:

ie-warning

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/765)

### DIFF
--- a/src/Site/Controller/Base.php
+++ b/src/Site/Controller/Base.php
@@ -188,11 +188,7 @@ class Base
     protected function isIE(string $userAgent)
     {
         // Internet Explorer 11 does not report itself as MSIE but does have Trident in the user agent string
-        if (strpos($userAgent, 'MSIE') || strpos($userAgent, 'Trident')) {
-            return true;
-        } else {
-            return false;
-        }
+        return (bool)(strpos($userAgent, 'MSIE') || strpos($userAgent, 'Trident'));
     }
 
     protected function getThemePath($theme = "") {
--- a/src/Site/Controller/Base.php
+++ b/src/Site/Controller/Base.php
@@ -27,6 +27,7 @@ class Base
         $this->data['version'] = VERSION;
         $this->data['useMinifiedJs'] = USE_MINIFIED_JS;
         $this->data['http_host'] = $_SERVER['HTTP_HOST'];
+        $this->data['isIE'] = $this->isIE($_SERVER['HTTP_USER_AGENT'] ?? '');
 
         $this->data['jsFiles'] = [];
         $this->data['jsNotMinifiedFiles'] = [];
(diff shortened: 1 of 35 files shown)


rmunn:
This would fix https://jira.sil.org/browse/LF-301

rmunn:
BTW, I got the idea of looking for MSIE or Trident from https://stackoverflow.com/questions/19999388/check-if-user-is-using-ie/22551342#22551342, and I checked http://www.useragentstring.com/pages/useragentstring.php?name=Internet+Explorer to verify that those two text tokens would identify all versions of IE but not Edge or other browsers.

Nateowami:
I definitely support this change.

I'm a little bit reluctant to recommend Edge. A couple years ago the lexicon editor had serious performance issues in Edge, to the point of being almost unusable. I tried to track down the issue but was ultimately unsuccessful. I just tried it in a VM and there didn't seem to be any trouble. Has anyone else tested in Edge anytime recently? I'm guessing it's probably OK now.

rmunn:
My thinking about recommending Edge was that anyone using Internet Explorer also has Edge already installed on their system without needing to go download anything, which might be important for users in Internet-poor areas. We could easily reorder the recommendations list to list Chrome first, then Firefox, then Edge. (Or even list Firefox first because it has Graphite, so people won't encounter font issues.)

Nateowami:
> anyone using Internet Explorer also has Edge already installed on their system

Really? I wouldn't have guessed. I know many people in China still use XP with antique versions of IE, but I doubt that's any significant part of our user base, so it's probably not very significant. I wouldn't be surprised if other areas of the world were similar though.

rmunn:
@megahirt Since this is a pretty important change, I'd like to get approval from the project lead before I merge this. Some points to consider:

- Should there be a close button on this so that the user can dismiss the message?
- If the user dismisses the message, how long should we wait before pestering him again? One hour? 24 hours?
- Currently I've used red for the message, matching what we use for errors. Should I change it to a warning (yellow) message instead?

PR #770, opened 2019-08-16
Use user-specified fonts on lexicon editor page

Nateowami:
This PR applies the user-specified fonts to elements that render lexical entries. They are already applied to inputs for the lexical editor, but there are some places where unrendered characters can still occur:

The methodology is slightly hackish, but I think it really is a reasonable solution. Rather than aplly a font face by setting `style="..."` on each element that may contain lexical data, a `user-fonts` class is put on all elements that may need a custom font, and then a `<style>` element is added that applies a font face to all `.user-fonts` elements.

The main alternative I can think of is applying the style directly to each element. This is a little more complicated, as font metadata has to be passed to each part of the application that handles rendering of lexical data.

Our current default font is `-apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol"`.

The font family this would apply to each `.user-fonts` element is `<that default font>, "user font for input system 1", "user font for input system 2"`, etc. This way most characters are still rendered with the same fonts, regardless of user settings, but it falls back to the user specified font when the other fonts cannot render it.

- Input systems with no custom font are skipped
- If a syntactically invalid font name is supplied, it is ignored
- I'm fairly confident this successfully avoids allowing CSS injection
- E2e tests are passing (except for errors caused by missing images on the login page, but those errors can be worked around, and then everything passes).

There's a companion pull request (#771) that is a cherry-pick of this commit to `lf-live`, as requested by @megahirt. Comments and discussion should be on this pull request, unless they directly relate to the cherry-pick to `lf-live`.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/770)

### DIFF
--- a/src/angular-app/languageforge/lexicon/editor/editor-entry.view.html
+++ b/src/angular-app/languageforge/lexicon/editor/editor-entry.view.html
@@ -77,7 +77,7 @@
                         <div class="col d-none d-md-block">
                             <div id="compactEntryListContainer" class="lexiconItemListContainer" data-pui-when-scrolled="$ctrl.show.more()">
                                 <div class="list-group">
-                                    <div id="entryId_{{entry.id}}" class="list-group-item lexiconListItemCompact"
+                                    <div id="entryId_{{entry.id}}" class="list-group-item lexiconListItemCompact user-fonts"
                                          data-ng-class="{selected: entry.id == $ctrl.currentEntry.id, listItemHasComment: $ctrl.getEntryCommentCount(entry.id) > 0}"
                                          title="{{$ctrl.getCompactItemListOverlay(entry)}}"
                                          data-ng-repeat="entry in $ctrl.visibleEntries track by entry.id" data-ng-click="$ctrl.editEntry(entry.id)">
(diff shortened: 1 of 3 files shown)


Nateowami:
@megahirt Sorry I've taken so long to get back to this.

You're right this isn't the ideal solution. I pursued it because I thought it was more robust than just assuming that what `getPrimaryListItemForDispaly()` returns corresponds to the first input system, which is what we'd discussed doing when we talked. It would also respect the user's font choice on the entry list view, and the entry preview, which we'd planned to leave alone at this time.

That said, I don't think implementing full-fledged font support will be that difficult. I think I would move some of the logic that is currently in the view into the controller, to have more control over the process.

> I think this PR is a replacement of the dynamic style="" method which is partially implemented, however I don't see code being removed for that piece. I would have expected this PR to remove the style="" method and use the class-based method instead, but I"m not seeing those deletions.

I didn't look closely into how we were wrapping things with HTML tags for styling. It appears to me that some changes had been made, so that what is now displayed on the page is not wrapped with HTML tags anymore. HTML tags *were* showing up in hover text though, where it doesn't belong. I'll be sure to look into this more when I rework this PR. Here's an example screenshot:
Screenshot_from_2019-08-16_16-17-45


> - continue to specify the exact font string for an input system input box or display

As far as I know, it's already working for inputs; it's just the display that needs work.
> - prevent CSS injection

This is not really difficult, especially with the abilities AngularJS gives us to specify conditional style properties. it was only really an issue when dynamically injecting a `<style>` tag into the page.
> - provide the simplest change that can get the user going again with displaying their language

I think doing it "right" is actually the simplest way, and I don't think it will be that complicated. I haven't looked at it closely, but I think it can be done without any drastic changes or introducing a lot of new complexity.

Nateowami:
Closing as this is superseded by #777.

PR #772, opened 2019-08-19
Fix bug where entryListModifiers.sortBy is null

rmunn:
It is often happening that there's an invalid `sortBy` parameter in the URL, and our code currently assumes that the URL parameter will always be found in the sort options; when that wasn't happening, the `findSelectedFilter` function was returning null or undefined. Now we check the result first, and only assign it to `entryListModifiers.sortBy` if it is valid.

This passes E2E tests once PR #768 has been merged, and should fix LF-258, LF-259, and LF-283.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/772)

### DIFF
--- a/src/angular-app/bellows/core/offline/editor-data.service.ts
+++ b/src/angular-app/bellows/core/offline/editor-data.service.ts
@@ -19,26 +19,36 @@ import {CommentsOfflineCacheService} from './comments-offline-cache.service';
 import {EditorOfflineCacheService} from './editor-offline-cache.service';
 import {LexiconCommentService} from './lexicon-comments.service';
 
-class FilterOption {
-  label?: string;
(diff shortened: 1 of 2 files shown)


rmunn:
After rebasing onto current master (commit b6953fcd2), E2E tests are failing with:

> Element `<a id="settingsBtn" class="btn btn-std dropdown-toggle" title="Settings">...</a>` is not clickable at point (48, 70). Other element would receive the click: `<div class="btn-group dropdown" id="settings-dropdown-button">...</div>`

It looks like we just need to update our `projects.page.ts` definition of what the settings button should look like. I'll open a new PR for that, to keep the good practice of having each PR be about one subject.

PR #773, opened 2019-08-19
Update PHP live config to have Facebook client ID and secret

Nateowami:
The lines needed in the config for Facebook login were added in 65fdc3b03eec4d28e679e08fff16edf4ebec549b, but only in `config.php`, not `config.php.live`. The result was that when the e2e tests finished running, `config.php` had been modified.

@rmunn, should they also be added to `config.php.fortest`?

Wouldn't it be "more correct" for them to be named `config.live.php` and `config.fortest.php`? At the very least editors would (probably) automatically detect it as PHP, whereas right now VS Code renders them as plain text files. I'm guessing the reason `config.php.fortest` ends with `fortest` is because `.test` would have caused the file to be treated incorrectly.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/773)

### DIFF
--- a/src/config.php.fortest
+++ b/src/config.php.fortest
@@ -70,6 +70,14 @@ if (! defined('GOOGLE_CLIENT_SECRET')) {
     define('GOOGLE_CLIENT_SECRET', 'googleClientSecret');
 }
 
+if (! defined('FACEBOOK_CLIENT_ID')) {
+    define('FACEBOOK_CLIENT_ID', 'facebookClientId');
+}
+
+if (! defined('FACEBOOK_CLIENT_SECRET')) {
+    define('FACEBOOK_CLIENT_SECRET', 'facebookClientSecret');
+}
+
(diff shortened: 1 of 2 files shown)


Nateowami:
@rmunn What about adding those lines to `config.php.fortest`? Are they needed there too, or only in the live version?

rmunn:
@Nateowami Probably a good idea to add them to `config.php.fortest` as well, now that I think about it. That way all the config files will be similar. We won't necessarily benefit from having them in `.fortest` because there's not really a good way to do E2E testing of Facebook OAuth (AFAIUI, Facebook doesn't allow "bot" accounts and wouldn't appreciate having someone's real account be automatically logged into by an automated system). But if that changes in the future, having the appropriate config lines already in `config.php.fortest` will make it easier.

Nateowami:
@rmunn Consistency seems like a good idea. I've amended and force-pushed with the lines added to the test config.

PR #777, opened 2019-09-14
Custom user fonts on editor page entry list

Nateowami:
- User-specified fonts applied to entry list on editor page
- Word/lexeme now truncated with ellipsis (this could be controversial, but I think it's better than the current behavior of wrapping and pushing the second line out of the way).
- String values are not set to "[Empty]" if their use is anything other than immediately being displayed in the UI. This means if a value is empty, we pass around the empty string, and only convert it to "[Empty]" for display in the UI.

Screenshot from 2019-09-13 22-21-56

This does not change how `dc-rendered` works, which is responsible for both the entry preview, and the entry list on the list page. I've started some work on that and made good progress. It was clear, however, that it would take significantly more work, so I wanted to get this PR out there as a start. Also, @megahirt, if I recall, when we last about this you were suggesting that we may want to leave the rendered entry alone for now, and the important part was the list on the editor page. I don't think it will take a huge amount of time, but didn't want to attack it before you had a chance to comment.

Tip: Viewing commits one at a time will probably make it significantly easier to understand than trying to take everything in at once.

E2e tests are passing.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/777)

### DIFF
--- a/src/angular-app/bellows/core/offline/editor-data.service.ts
+++ b/src/angular-app/bellows/core/offline/editor-data.service.ts
@@ -311,11 +311,7 @@ export class EditorDataService {
         );
     }
 
-    if (!meaning) {
-      return '[Empty]';
-    }
-
-    return meaning;
+    return meaning || '';
   }
 
   getSortableValue = (config: any, entry: any): string => {
@@ -385,11 +381,7 @@ export class EditorDataService {
       }
     }
 
-    if (!sortableValue) {
-      return '[Empty]';
-    }
-
-    return sortableValue;
+    return sortableValue || '';
   }
 
   private doFullRefresh(offset: number = 0): angular.IPromise<any> {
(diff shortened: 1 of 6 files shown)


Nateowami:
[Display lexical data with user-specified font](https://trello.com/c/6Kul0MSd/434-display-lexical-data-with-user-specified-font)

Nateowami:
@megahirt Do you want a PR that cherry-picks this to live as well? I believe this PR is ready to merge.

megahirt:
No worries.  I'm going to cherry pick or rebase this change onto live and
then test it there.  Thanks.

Warm regards,

Christopher Hirt
Team Manager - Thailand
SIL Language Software Development
chris_hirt@sil.org

megahirt:
I take that back :(  The cherry-pick on live has merge conflicts, and since
we don't have a shipping QA box to test this out on, I'm flying blind in a
sense and don't feel qualified to resolve the conflicts and push to live.
If you can do a rebase/cherry-pick and are fairly confident that it will
work with the code from back in July, then I'll merge it.  Otherwise, this
will have to wait until I have time to get bionic working for QA and Live.

Warm regards,

Christopher Hirt
Team Manager - Thailand
SIL Language Software Development
chris_hirt@sil.org

PR #778, opened 2019-09-18
Fix LF-372: parse inline LIFT ranges right

rmunn:
LiftImport needs to be able to cope with a `<range>` element that doesn't have an `href` attribute, but defines the range inline. Our current code had an assumption that there will always be an `href` attribute, but the LIFT spec (as defined by the RelaxNG schema) [lists `href` as optional](https://github.com/sillsdev/libpalaso/blob/18f9433cdf2808bc0f362eef9ae38e6c8706c8ca/SIL.Lift/Validation/lift-0.13.rng#L537). The code already knows how to parse `<range>` elements defined inline, we just need to make sure we don't assume the presence of an optional attribute.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/778)

### DIFF
--- a/src/Api/Model/Languageforge/Lexicon/Import/LiftDecoder.php
+++ b/src/Api/Model/Languageforge/Lexicon/Import/LiftDecoder.php
@@ -382,7 +382,9 @@ class LiftDecoder
             throw new \Exception("'" . $sxeNode->getName() . "' is not a gloss");
         }
         $inputSystemTag = \Normalizer::normalize((string) $sxeNode['lang']);
-        $multiText->form($inputSystemTag, (string) $sxeNode->{'text'});
+        // Some LIFT files can have multiple <gloss> elements which are supposed to be separated by a semicolon-space
+        // pair, so we use appendForm() here instead of form(). 2019-09 RM
+        $multiText->appendForm($inputSystemTag, (string) $sxeNode->{'text'}, '; ');
 
         $this->project->addInputSystem($inputSystemTag);
         if (isset($inputSystems)) {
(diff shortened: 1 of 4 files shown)


rmunn:
No unit tests yet, which is why this PR is still WIP. I'll add unit tests to exercise this bug (and verify no regressions) tomorrow.

rmunn:
Unit tests added. Note that one of these two tests currently FAILS, because our LIFT import code doesn't currently handle the following situation (taken from a real file produced by Lexique Pro):

```xml
  <sense id="foo; bar_12345678-90ab-cdef-1234-567890abcdef">
    <grammatical-info value="n"/>
    <gloss lang="en"><text>foo</text></gloss>
    <gloss lang="en"><text>bar</text></gloss>
  </sense>
```

It's clear from the sense ID that this is supposed to result in a gloss of `foo; bar`, but our LiftImport.php code currently results in "last one wins", e.g. `bar` for the above. So this is going to cause user confusion when they import a file like this and see most of their glosses missing.

The LIFT spec has a similar example:

```xml
<entry id="db24e454-307d-4939-a517-7852376185c2">
  <lexical-unit>
    <form lang="und-Latn"><text>utan</text></form>
  </lexical-unit>
  <sense id="ae081346-aa0a-4e79-86db-1ebcf94d5a53">
    <grammatical-info value="n"/>
    <trait name="semantic-domain" value="Nplant"/>
    <gloss lang="en"><text>veg</text></gloss>
    <gloss lang="id"><text>sayur</text></gloss>
    <gloss lang="id"><text>jamu</text></gloss>
    <!-- Reversal and Definition omitted -->
  </sense>
</entry>
```

The corresponding SFM data for that example is:

```
\ge veg
\gn sayur ; jamu
```

So it seems plain that the multiple glosses are intended to be separated by semicolons (whether there should be a space before the semicolons is unclear since it's not consistent between these two examples), but our code currently doesn't do that.

Note that the actual *text* of the LIFT spec is silent on how to handle multiple glosses in a single sense, but the example in the spec, and a case of actual usage in Lexique Pro, suggest that semicolon separation is what would be expected.

PR #784, opened 2019-10-01
Project insights

Nateowami:
Ready for review.

~~This isn't complete yet, but I think it's ready to start being reviewed.~~ There are some attributes that are common to SF and LF projects, and some that are specific to the project type.

~~At this point I've implemented the parts that are needed for SF (generic data, and SF-specific data), but not LF-specific data.~~

There are likely bugs and/or obvious things that need to be changed. ~~I haven't reviewed it yet.~~

~~Here's what it currently looks like on my local SF install:~~ (Table has been removed in favor of CSV download).

Screen Shot 2019-09-30 at 22 10 20

TODO:
- ~~Implement LF-specific data~~
- ~~Look over it for obvious mistakes~~
- ~~Convert it to a CSV download~~

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/784)


Nateowami:
I've made some more changes and implemented the parts for LF, except for number of entries with audio. That turns out to be very difficult to do without loading every single lexical entry on the server into memory and traversing it. I think for hundreds of projects this would be prohibitively slow (based on extrapolations of response times on my machine). I do have a solution in mind that involves dynamically creating a very complex* query for Mongo to execute (Complex doesn't necessarily mean slow, and I think it would still be quite robust), but I didn't think it was important enough to hold up the stats page for SF. I think generating a CSV file is of greater importance, so I plan to focus my efforts there first.

Nateowami:
I believe the PR is ready for final review.

Here's example output for LF: [languageforge_project_insights_2019-10-02T15_33_46.740Z.zip](https://github.com/sillsdev/web-languageforge/files/3682861/languageforge_project_insights_2019-10-02T15_33_46.740Z.zip)

PR #797, opened 2020-03-16
Fix bug where editing input systems could fail

rmunn:
In some cases, field.inputSystems could end up with at least one undefined value in field-unified-view-model.ts after many user edits. This change should fix that issue.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/797)


rmunn:
@megahirt I'm pretty sure the null gets there due to the issue I fixed in https://github.com/sillsdev/web-languageforge/pull/798 - if there's an optionlist with an empty `items` array, the MongoDB PHP code sees an empty array and removes it from the database to save room. Then the Javascript code loads an `undefined` value instead of an empty array in the picklist control's `items` array, which leads to the picklist result being undefined -- and that result is what's getting in here.

So once #798 is merged too, this should not reoccur. But it's still a good idea to have null checks.

PR #799, opened 2020-06-04
Use API token for new Language Depot API

rmunn:
It does no harm to send an Authorization: Bearer token to an API endpoint that's not expecting it, so we can put this into place in LF first and then add the API token requirement to the Language Depot API.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/799)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -961,6 +961,11 @@ gulp.task('build-productionConfig', function () {
     facebookClientSecret = 'facebookClientSecret';
   }
 
+  var languageDepotApiToken = process.env.LANGUAGE_DEPOT_API_TOKEN;
+  if (languageDepotApiToken === undefined) {
+    languageDepotApiToken = 'languageDepotApiToken';
+  }
+
   var gatherWordsClientId = process.env.GATHERWORDS_CLIENT_ID;
   if (gatherWordsClientId === undefined) {
     gatherWordsClientId = 'gatherWordsClientId';
@@ -1003,6 +1008,10 @@ gulp.task('build-productionConfig', function () {
       demand: false,
       default: facebookClientSecret,
       type: 'string' })
+    .option('languageDepotApiToken', {
+      demand: false,
+      default: languageDepotApiToken,
+      type: 'string' })
     .option('gatherWordsClientId', {
       demand: false,
       default: gatherWordsClientId,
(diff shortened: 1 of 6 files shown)


rmunn:
Weird; the PHP unit tests are warning me of an undefined `LANGUAGE_DEPOT_API_TOKEN` constant, but the PR defines it in config.php, config.php.live, and config.php.fortest. Are those not included when PHP runs the unit tests? (Same behavior when I run the tests locally as when TeamCity runs them, so it's not a difference in the TeamCity environment).

rmunn:
Looks like that did the trick: LF unit tests are all passing on TeamCity. As soon as the rest of the required status checks come in, I'll merge this PR.

megahirt:
@rmunn please finish out this PR.  Looks like you nearly got it across the finish line.  I see the TC build failed on unit tests.  Was this environmental?  Please reply with your plan for this branch in this PR.  Thanks.

rmunn:
I re-ran the TeamCity build and this time the unit tests all passed, and the two failures looked environmental (failed to find an expected file), so I think this is ready to merge. I'll merge it tomorrow when I have a chance to actually look at the code.

rmunn (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/799)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[test/php/TestConfig.php, line 33 at r3](https://reviewable.io/reviews/sillsdev/web-languageforge/799#-M9Rv0Qi69wnaSnNIcpD:-MQSSRNCBM26YIXZtr1f:b-1rr2my) ([raw file](https://github.com/sillsdev/web-languageforge/blob/a3013b06f0a022d4c236a4bad32a7ae2b28ea7cc/test/php/TestConfig.php#L33)):*

> ```
> define('LANGUAGE_DEPOT_API_TOKEN', 'not-a-secret');
> ```
This feels wrong since the other secrets aren't needed here. Is there something different about this one? However, since this is test code, it doesn't need to block this PR.
</blockquote></details>

This is months since I apparently missed this comment when it was originally written, but if the tests are going to talk to Language Depot at all (e.g. if they're testing the new LD API in any way) then they're going to need the API token to be available. For PHP unit tests, you're right that it's not needed; I think I added it because I was just reflexively adding this to all the *Config.php files. I'll remove it from this one.

rmunn:
Ended up not deleting the LANGUAGE_DEPOT_API_TOKEN definition from TestConfig.php after all, despite what I said to Ira yesterday, because I noticed that I had specifically added it in the most recent commit on that branch. (And that's *all* that that commit did, which means that it was necessary because something had broken without it.) It's been long enough since I did that that I don't remember *what* unit test needed the token, but I'm pretty sure I added it because one of the unit tests actually needed to talk to the LD API and therefore needed the token.

PR #800, opened 2020-07-21
Ansible script to deploy on Ubuntu 20.04 (Focal)

rmunn:
Two unit tests fail on PHP 7.4, so this PR might need to be followed with some adjustments to make the unit tests work properly in PHP 7.4. Marking this as a draft PR until I figure out what changed in PHP 7.4 that causes these failures.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/800)

### DIFF
--- a/deploy/hosts
+++ b/deploy/hosts
@@ -1,5 +1,5 @@
 [all]
-localhost ansible_connection=local
+localhost ansible_connection=local ansible_python_interpreter=/usr/bin/python3
 
 [build_agents]
 ba_xenial_web_s2_138 ansible_ssh_host=lf-dev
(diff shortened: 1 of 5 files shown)


rmunn:
Commit https://github.com/sillsdev/web-languageforge/pull/800/commits/580516fec8607144032598d85e519625ace9ba59 makes all PHP unit tests pass on Focal with PHP 7.4, but there's still more work to do. Loading up the login page in a browser yields deprecation warnings about "Using array_key_exists() on objects is deprecated. Use isset() or property_exists() instead." We'll want to fix those deprecations, too, before we call this PR ready to merge.

rmunn:
Some of the deprecated `array_key_exists()` calls are in vendor code, e.g. Twig, which needs to be updated. But some are in our own code. This will not be a small patch.

rmunn:
This PR should be **squashed and merged** into a single commit once I've addressed the review comments. I don't intend for my PHP 7.4 changes to make their way into the master branch.

PR #801, opened 2020-07-30
Make user list robust against missing projects

rmunn:
When the site admin tries to list users, if there is even one user who was in a deleted project but somehow that project was not removed from their membership list, the user list fails to load. This change makes it just ignore that missing project when loading the user list.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/801)


rmunn:
Found this by accident while working on the AngularJS -> Angular upgrade, so this is a quick-and-dirty fix. A proper fix would probably involve a database migration script to handle this case, and that's not a high priority for me right now.

PR #808, opened 2021-01-08
Fix blank project page for Tech Support role

rmunn:
Many projects don't have anything configured for the Tech Support role, since it was newly added, so those projects end up throwing an exception in the filter() call in removeDisabledConfigFields. And since the filter didn't return true, all fields end up being stripped out. The Tech Support role is supposed to have the same rights as a project manager, so if it's looked for and not found, we give manager rights.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/808)

### DIFF
--- a/src/angular-app/languageforge/lexicon/core/lexicon-config.service.ts
+++ b/src/angular-app/languageforge/lexicon/core/lexicon-config.service.ts
@@ -1,6 +1,7 @@
 import * as angular from 'angular';
 
 import {SessionService} from '../../../bellows/core/session.service';
+import {ProjectRoles} from '../../../bellows/shared/model/project.model';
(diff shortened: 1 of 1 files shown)


rmunn:
This is a bugfix that I apparently never created a PR for, but that still LGTM. It's not a bug that users will encounter, but if you're logged in with an account that has the Tech Support role on a project, there's some UI annoyance that this PR fixes.

megahirt (approved):
Great!  Thanks for this contribution.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/808#-:-MRIGBZkCXLv8fb-Am__:boi96pd)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #809, opened 2021-01-08
Input system tag, not name, in comment.regarding

rmunn:
This is a revival of the **bugfix/regarding-inputsystem-should-be-tag** branch, which had suffered code rot over the years and which I'm about to delete. Some of its fixes had been applied since it was created, but two fixes in that branch had never gotten applied, so I'm reviving them here.

Before merging this PR, I need to investigate the state of comments in the Mongo database on live, to see how many of them (if any) suffered from this bug where the input system *name* ("English") was inserted in the `comment.regarding` field, where the *tag* ("en") should be. The effect of that bug would be that the link between the comment and the thing it's commenting about gets lost and the UI can't render the comment bubble in the right place (e.g., it'll try to render on the "English" part of the Notes field, but the Notes field has an "en" part, which doesn't match). If there are any places where this bug actually showed up, I'll have to write a migration script to apply this fix to the database as well as to the UI.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/809)

### DIFF
--- a/src/angular-app/bellows/core/offline/lexicon-comments.service.ts
+++ b/src/angular-app/bellows/core/offline/lexicon-comments.service.ts
@@ -54,7 +54,7 @@ export class LexiconCommentService {
               contextGuid = comment.contextGuid;
             } else {
               contextGuid = comment.regarding.field +
-                (comment.regarding.inputSystemAbbreviation ? '.' + comment.regarding.inputSystemAbbreviation : '');
+                (comment.regarding.inputSystem ? '.' + comment.regarding.inputSystem : '');
               if (fieldConfig != null && fieldConfig.type === 'multioptionlist') {
                 contextGuid += '#' + comment.regarding.fieldValue;
               }
--- a/src/angular-app/languageforge/lexicon/editor/comment/lex-comments-view.component.ts
+++ b/src/angular-app/languageforge/lexicon/editor/comment/lex-comments-view.component.ts
@@ -46,7 +46,7 @@ export class LexCommentsViewController implements angular.IController {
       this.newComment.contextGuid = contextGuid;
       if (inputSystemTag) {
         this.newComment.regarding.fieldValue = LexCommentsViewController.getFieldValue(model, inputSystemTag);
-        this.newComment.regarding.inputSystem = this.control.config.inputSystems[inputSystemTag].languageName;
+        this.newComment.regarding.inputSystem = inputSystemTag;
         this.newComment.regarding.inputSystemAbbreviation =
           this.control.config.inputSystems[inputSystemTag].abbreviation;
       } else if (multioptionValue) {


rmunn:
Original commit from the **bugfix/regarding-inputsystem-should-be-tag** branch, so that we don't have to go digging for it once the branch is deleted. Having it here should allow us to verify that all of its fixes got into master.

```diff
commit e4bba926e2319d755efe9d0ebfa0e200a4c3ee6f
Author: Robin Munn <rmunn@pobox.com>
Date:   Tue Feb 27 15:27:56 2018 +0700

    Input system tag, not name, in comment.regarding
    
    We've had this bug for several years, so a migration script may be
    required. But some parts of the code assume that a comment's "regarding"
    object will have inputSystem be the *tag* ("en"), and yet it's being set
    to the *name* ("English"). We never use it to display the name to the
    user (that's what inputSystemAbbreviation is for), so this changes it to
    actually be the tag that the rest of the code is expecting.
    
    
    Former-commit-id: 569d22264e67776f4702e9e8ccedee1153a845bf

diff --git a/src/angular-app/bellows/core/offline/lexicon-comments.service.ts b/src/angular-app/bellows/core/offline/lexicon-comments.service.ts
index 3b3c451b8..2ded769b3 100644
--- a/src/angular-app/bellows/core/offline/lexicon-comments.service.ts
+++ b/src/angular-app/bellows/core/offline/lexicon-comments.service.ts
@@ -85,7 +85,7 @@ export class LexiconCommentService {
               contextGuid = comment.contextGuid;
             } else {
               contextGuid = comment.regarding.field +
-                (comment.regarding.inputSystemAbbreviation ? '.' + comment.regarding.inputSystemAbbreviation : '');
+                (comment.regarding.inputSystem ? '.' + comment.regarding.inputSystem : '');
               if (fieldConfig.type === 'multioptionlist') {
                 contextGuid += '#' + comment.regarding.fieldValue;
               }
diff --git a/src/angular-app/bellows/directive/palaso.ui.comments.comment-bubble.js b/src/angular-app/bellows/directive/palaso.ui.comments.comment-bubble.js
index da2e8b82b..a9147f4eb 100644
--- a/src/angular-app/bellows/directive/palaso.ui.comments.comment-bubble.js
+++ b/src/angular-app/bellows/directive/palaso.ui.comments.comment-bubble.js
@@ -100,8 +100,8 @@ angular.module('palaso.ui.comments')
               $scope.contextGuid += '#' + $scope.multiOptionValue;
             }
 
-            $scope.contextGuid += ($scope.inputSystem.abbreviation ? '.' +
-              $scope.inputSystem.abbreviation : '');
+            $scope.contextGuid += ($scope.inputSystem.tag ? '.' +
+              $scope.inputSystem.tag : '');
             if ($scope.contextGuid.indexOf('undefined')  === -1) {
               $scope.active = true;
             }
diff --git a/src/angular-app/bellows/directive/palaso.ui.comments.lex-comments-view.js b/src/angular-app/bellows/directive/palaso.ui.comments.lex-comments-view.js
index 9b086d42c..ecf449be8 100644
--- a/src/angular-app/bellows/directive/palaso.ui.comments.lex-comments-view.js
+++ b/src/angular-app/bellows/directive/palaso.ui.comments.lex-comments-view.js
@@ -42,7 +42,7 @@ angular.module('palaso.ui.comments')
               if (inputSystem) {
                 $scope.newComment.regarding.fieldValue = getFieldValue(model, inputSystem);
                 $scope.newComment.regarding.inputSystem =
-                  $scope.config.inputSystems[inputSystem].languageName;
+                  $scope.config.inputSystems[inputSystem].tag;
                 $scope.newComment.regarding.inputSystemAbbreviation =
                   $scope.config.inputSystems[inputSystem].abbreviation;
               } else if (multioptionValue) {
```

megahirt (approved):
I approve of this PR.  This is a good contribution.  Can you describe a bit more of what we are looking for in the Mongo database to determine if a migration is necessary?  For example, look for any comment where the regarding.inputSystem does not match an valid bcp47 tag?

Also, what is the implication of merging this without knowing whether data needs to be migrated?  What will users see if "English" is in the inputsystem attribute?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/809#-:-MRIH0qo0tuvRwFcJJx1:b-ve161m)* status: :shipit: complete! all files reviewed, all discussions resolved

megahirt (commented):
Let's discuss before merging...

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/809#-:-MRII8r31WKGxfIeQ4kz:b-b6izes)* status: :shipit: complete! all files reviewed, all discussions resolved

rmunn (commented):
Yes, checking for comments where `regarding.inputSystem` does not match a valid tag would be a good start, as well as comments where `regarding.inputSystem` does match the *name* of an input system found in the project.

I'll double-check what would happen if the data isn't migrated, but I suspect what would happen is that an error message would show up in the browser console (invisible to the user) and the comment's "Regarding" feature (the part that's supposed to show the data in the field that the comment is about) would just be invisible, i.e. the user would see the comment divorced from its context. Depending on the previous comment conversation, that could mean very little as the context would be obvious, or it could mean the person reading the comment would call his colleague and say "I don't know which part of the entry you were talking about here. Do you mean the sense, or the definition?" And the work would get done slower than it would have if we had managed to display the right context.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/809#-:-MRIhc6l9XgYY4fDATgK:b-6fxq8d)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rmunn)

rmunn (commented):
After further code review and looking through Mongo, I've concluded that my initial approach to this bug was a little mistaken, and there's a better approach that will save work. We'll still need to run a migration, but the migration won't require nearly as much work.

The comment "regarding" field has two properties, inputSystem and inputSystemAbbreviation. (Plus some other properties that aren't relevant to this issue). The contextGuid for the comment, which is what is used to link it up to its data source, is built from the inputSystemAbbreviation. The inputSystemAbbreviation, in turn, is being set from inputSystems[tag].abbreviation, which seems quite reasonable but is wrong. That should be set from the tag, because the abbreviation is user-editable and will probably be localized into the user's native language.

My original proposal was to set the inputSystem value from the tag, but I now realize that that would create a massive change in the database. Whereas the inputSystemAbbreviation is correct for everyone using the English UI, and for *most* other people (I'm sure there are some where it's wrong, but I have yet to find one in my manual search). So I need to revise this PR and not change the inputSystem value after all.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/809#-:-MRIxKZ74pjnNziB4FUf:b-l6ynbq)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rmunn)

megahirt:
Good call.  I didn't catch that!

Kind regards,

Chris

rmunn:
Setting PR to draft status so it doesn't get merged by mistake until I've changed the approach used.

megahirt:
@rmunn is there uncommitted work that you've done that includes the new approach you talk about in your PR?  I'd like to see this finished up sooner rather than later.

I am also going to rebase this branch on master to kick off the new unit/e2e tests

rmunn:
Closing for now because I'm working on other things. I'll reopen this when I'm working on it again.

PR #810, opened 2021-01-08
Ignore Javascript errors from outdated browsers

rmunn:
Draft PR until I've checked Bugsnag; I'll either mark it ready for review or close it depending on what I find. See below for details.

We have (or had in the past) one user with device with an outdated browser (Chrome Mobile 30), which was sending a lot of Bugsnag reports about Javascript features that that browser doesn't implement because it's so old (e.g., the Array.includes function). These reports aren't actually bugs and we won't do anything about them, so we should ignore them and not send them to Bugsnag to reduce how much Bugsnap spam we're receiving. **If,** that is, that same user is still using that device with the old browser. Once I've had a chance to look through our Bugsnag reports, I'll either mark this PR as ready for review if that device is still spamming us with Bugsnag reports, or I'll close this PR if that device has gone away and we no longer have to deal with Chrome Mobile 30 any longer (which would be great news).

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/810)

### DIFF
--- a/src/angular-app/bellows/core/exception-handling.service.ts
+++ b/src/angular-app/bellows/core/exception-handling.service.ts
@@ -56,6 +56,15 @@ export class ExceptionHandlingService {
       return;
     }
 
+    // Silence errors coming from out-of-date browser versions
+    // So far we've only seen errors from Chrome version 30, but if other old
+    // browser versions trigger errors, they can be added below as well
+    const userAgent = navigator.userAgent;
+    if (userAgent.indexOf('Chrome/30') >= 0) {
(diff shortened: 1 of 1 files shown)


megahirt:
@rmunn I did some digging on this and found #705 which is a PR for this branch.  We discussed it and we decided that this code isn't necessary in favor of tweaking our TS compilation.  I don't know if that will ever happen at this point, but it's a moot point since we haven't gotten this error in ages.  Thanks for following up.

I'm closing this PR and will also delete the remote branch per our decision.

PR #811, opened 2021-01-08
Handle rearranged items (senses, examples, etc.) in Mongo

rmunn:
WIP which was waiting until we would upgrade the live server to Mongo 3.6, because [the positional-update operators](https://docs.mongodb.com/manual/reference/operator/update/positional-filtered/) (new in Mongo 3.6) would be necessary for the rest of the work. That's happened by now, so work can resume on this feature at some point.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/811)


megahirt:
This branch (and all the others) need to be rebased.

I see now that this is unfinished work, so a WIP PR would be appropriate, except that I have no idea when we could schedule finishing this work.  So I will mark this branch in our document as unfinished work in progress but close the PR since it's not "almost finished".

PR #814, opened 2021-01-13
feature/dockerize

megahirt:
A draft PR with no intention to merge as means for Billy and I to discuss the code...asynchronously.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/814)


Dockerfile
```diff
@@ -1,29 +1,39 @@
...
+# Extensions we know LF requires/desires:
+# curl - already provided in image
+# mbstring - already provided in image
+# xdebug
+# intl
+# gd
+# mongodb
+#
+# see https://github.com/mlocati/docker-php-extension-installer
+COPY --from=mlocati/php-extension-installer /usr/bin/install-php-extensions /usr/local/bin/
```
longrunningprocess, line 12: slick!  Is this simply a different way of addressing extensions?  Here's what I did for mongo...

```yaml
# install mongo extension and enable it
RUN pecl install mongodb && \
    docker-php-ext-enable mongodb
```

obviously there's more than one way to skin a cat and these may both work so it may be moot, I'm asking for my own knowledge.

Dockerfile
```diff
@@ -1,29 +1,39 @@
...
+RUN install-php-extensions gd xdebug mongodb curl intl
```
longrunningprocess, line 13: it would be premature at this point but let's add a TODO: in here so we remember to come back and optimize these "layers", i.e., we'll want to do things like this in one line once we get past the "make it work" phase and begin optimization and housekeeping.

Dockerfile
```diff
@@ -1,29 +1,39 @@
...
+#   - postfix
+#   - git
+#
+#
+#
 
-# troubleshooting latest failure...
-# `docker-compose run --rm lf bash` will get you into container
-# `ansible-playbook playbook_bionic.yml --limit localhost` (with or without -K not relevant afaik at this point but bash-ing into the container will provide the opportunity to run with the -K (just enter at prompt))
+# it feels like these should run in separate containers:
```
longrunningprocess, line 36: conventionally, yes..., theoretically, if node is only for bundling, we could do it as part of the container start-up (or even the image build) but we'd lose (for now) the HMR stuff.

And, as you thought, there may be some interesting ways of running the ui build in a separate container but getting the dist in the right place within the php container via volumes.

Dockerfile
```diff
@@ -1,29 +1,39 @@
...
 
-# time to run the ansible scripts
-WORKDIR /data/deploy
 
-# TODO: need to make a distinction between environment setup commands and those commands meant for running the services.
-# TODO: left off the -K otherwise this command becomes interactive (password requested)
-# RUN ansible-playbook playbook_bionic.yml --limit localhost
+# apt packages to investigate further
+#   - default-jre-headless
+#   - lfmerge
```
longrunningprocess, line 28: this one we still has my curiosity piqued...we may be able to run it in a separate container and use some form of `rsh/rexec` from the php container to the lfmerge container for issuing those commands against another "volume-shared" file system.  Docker sets up a private network between the containers so this should be possible as far as I can think without trying it.

Ideally we'd have a daemon-like process listening for some protocol to issue the commands appropriately.

Dockerfile
```diff
@@ -1,29 +1,39 @@
...
+#
+#
+#
 
-# troubleshooting latest failure...
-# `docker-compose run --rm lf bash` will get you into container
-# `ansible-playbook playbook_bionic.yml --limit localhost` (with or without -K not relevant afaik at this point but bash-ing into the container will provide the opportunity to run with the -K (just enter at prompt))
+# it feels like these should run in separate containers:
+# nodejs
+# postfix (for sending email)
```
longrunningprocess, line 38: absolutely...we'd want to investigate the existing postfix images out there but a quick search and https://hub.docker.com/r/juanluisbaptiste/postfix already looks promising.

docker/lfmerge/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM ubuntu:bionic
+
+RUN apt update && apt install curl \
+&& curl -L http://linux.lsdev.sil.org/downloads/sil-testing.gpg > /etc/apt/trusted.gpg.d/sil-testing.gpg \
+&& echo "deb http://linux.lsdev.sil.org/ubuntu bionic main" > /etc/apt/sources.list.d/linux-lsdev-sil-org.list \
+&& apt update
+
+RUN apt install -y lfmerge
```
longrunningprocess, line 8: I was curious about the `apt` usage I see...I'm used to seeing `apt-get`...I'm "guessing" they are probably the same but one is meant to be interactive while another is meant be used programmatically?

longrunningprocess (commented):
short of a `Dockerfile` being present in the app's root directory, there's no specific convention when there are multiple...what you did is fine.  I took a similar approach in that little docker-sandbox repo of mine and just had top-level directories for each "app".

Traditionally, each app would be in its own repo as a separate, independent codebase that would publish an image to Docker Hub to then be pulled in as needed.  This is why you typically won't see multiple Dockerfile's in a single project.  It's certainly possible but it is a "code smell" that reflects high coupling...certainly not meant as a criticism of this project, obviously we're in the throws of a POC here.

megahirt (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/814)* status: 0 of 10 files reviewed, 6 unresolved discussions (waiting on @longrunningprocess)

___
*[docker/lfmerge/Dockerfile, line 8 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/814#-MQvrkC02HITVv4gKpzN-r2-8:-MR4Czm69Bl8ksOP_SUP:b5yaie) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2a33ff6e3e41f49f31f83dd1b99469a0356db9f2/docker/lfmerge/Dockerfile#L8)):*

I was curious about the `apt` usage I see...I'm used to seeing `apt-get`...I'm "guessing" they are probably the same but one is meant to be interactive while another is meant be used programmatically?
</blockquote></details>

looks like we should always use apt-get in scripts
___
*[Dockerfile, line 12 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/814#-MQvrkC02HITVv4gKpzI-r2-12:-MR4D6_92PjRFmH7QdlX:b-ugv6ai) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2a33ff6e3e41f49f31f83dd1b99469a0356db9f2/Dockerfile#L12)):*

slick!  Is this simply a different way of addressing extensions?  Here's what I did for mongo...

```yaml
# install mongo extension and enable it
RUN pecl install mongodb && \
    docker-php-ext-enable mongodb
```

obviously there's more than one way to skin a cat and these may both work so it may be moot, I'm asking for my own knowledge.
</blockquote></details>

I suspect using the php-extension-installer isn't buying us much which this short list of extensions.  Maybe the gd extension since that requires an additional apt package.  The extension installer takes care of installing any apt packages dependencies whereas the provided docker-php-ext-enable does not
___
*[Dockerfile, line 13 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/814#-MQvrkC02HITVv4gKpzI-r2-13:-MR4Dd1SD70-EFcDyc5c:bp7vbga) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2a33ff6e3e41f49f31f83dd1b99469a0356db9f2/Dockerfile#L13)):*

it would be premature at this point but let's add a TODO: in here (or capture an action somewhere) so we remember to come back and optimize these "layers", i.e., we'll want to do things like this in one line once we get past the "make it work" phase and begin optimization and housekeeping.
</blockquote></details>

sounds good.
___
*[Dockerfile, line 28 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/814#-MQvrkC02HITVv4gKpzI-r2-28:-MR4DhIcAqcLQJ3vk1gE:bb7e6n6) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2a33ff6e3e41f49f31f83dd1b99469a0356db9f2/Dockerfile#L28)):*

this one we still has my curiosity piqued...we may be able to run it in a separate container and use some form of `rsh/rexec` from the php container to the lfmerge container for issuing those commands against another "volume-shared" file system.  Docker sets up a private network between the containers so this should be possible as far as I can think without trying it.

Ideally we'd have a daemon-like process listening for some protocol to issue the commands appropriately, e.g., lfmerge with a REST frontend or something :-D, but that could be over-engineered if simply issuing remote commands between containers will work.
</blockquote></details>

yes - I've split out lfmerge into it's own container
___
*[Dockerfile, line 36 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/814#-MQvrkC02HITVv4gKpzI-r2-36:-MR4DlfM8RyPskJdOPmh:b-ejuuju) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2a33ff6e3e41f49f31f83dd1b99469a0356db9f2/Dockerfile#L36)):*

conventionally, yes..., theoretically, if node is only for bundling, we could do it as part of the container start-up (or even the image build) but we'd lose (for now) the HMR stuff.

And, as you thought, there may be some interesting ways of running the ui build in a separate container but getting the dist in the right place within the php container via volumes.
</blockquote></details>

I think we're going to try and get LF running without any postfix or email service for now...
___
*[Dockerfile, line 38 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/814#-MQvrkC02HITVv4gKpzI-r2-38:-MR4DrAnEJWYT-u4SKXG:b-mvv2sx) ([raw file](https://github.com/sillsdev/web-languageforge/blob/2a33ff6e3e41f49f31f83dd1b99469a0356db9f2/Dockerfile#L38)):*

absolutely...we'd want to investigate the existing postfix images out there but a quick search and https://hub.docker.com/r/juanluisbaptiste/postfix already looks promising.
</blockquote></details>

yes, thanks.

megahirt:
I've renamed my feature/dockerize-chris branch which clobbers feature/dockerize and squashed commits, and I'm hoping @longrunningprocess is ok with that :)

PR #831, opened 2021-02-23
Add Docker images for Language Depot API

rmunn:
<del>WIP. At the moment these Docker images will build correctly, but I don't yet have the right infrastructure in place to have the Language Forge images talk to them. I want to create a hosts entry in the lf-api container that will tell it that "admin.languagedepot.org" is the ld-api container, and have lf-api trust the certificate (generated by Caddy?) that ld-api will return. Once that's in place, this will be ready to merge.</del>

**Update:** Now working and ready for review.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/831)

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -153,6 +153,7 @@ services:
       - mail
     environment:
       - WAIT_HOSTS=db:27017, mail:25
+      - "LDAPI_BASE_URL=http://ld-api:3000/api/v2/"
     command: sh -c "/wait && /run.sh"
     volumes:
       # share dist folder between ui and api containers
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -32,6 +32,7 @@ services:
     environment:
       - WAIT_HOSTS=db:27017, mail:25
       - "LDAPI_BASE_URL=http://ld-api:3000/api/v2/"
+      # LDAPI_BASE_URL hostname and port need to match container name and port of the ld-api container (defined further down)
     # TODO: Make a run.sh that sets umask and then runs apache2-foreground
     command: sh -c "/wait && umask 0002 && apache2-foreground"
     volumes:
@@ -98,10 +99,10 @@ services:
     build:
       context: ld-sql
     image: ld-sql
-    ports:
-      # Exposed to host so you can run MySQL admin tools if you want
-      # Note that this will *fail* if you have MySQL running on your host, as port 3306 will already be taken
-      - 3306:3306
+    # To access the MySQL database via localhost:3306 on your dev machine (e.g., in VS Code), uncomment the "ports" config below
+    # Note that if you're running MySQL on your dev machine already, change the first number to something else, like 3307 and access localhost:3307
+    # ports:
+    #   - 3306:3306
     env_file: ld-api/ld.env
     container_name: ld-sql
 
@@ -111,10 +112,10 @@ services:
     image: ld-api
     depends_on:
       - ld-sql
-    ports:
-      - 3000:3000
     # env_file should be same one given to ld-sql image
     env_file: ld-api/ld.env
+    environment:
+      - PORT=3000
     container_name: ld-api
 
   e2e-ui:
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -28,8 +28,10 @@ services:
     depends_on:
       - db
       - mail
+      - ld-api
     environment:
       - WAIT_HOSTS=db:27017, mail:25
+      - "LDAPI_BASE_URL=http://ld-api:3000/api/v2/"
     # TODO: Make a run.sh that sets umask and then runs apache2-foreground
     command: sh -c "/wait && umask 0002 && apache2-foreground"
     volumes:
@@ -91,6 +93,29 @@ services:
       - /dev/shm:/dev/shm
     ports:
       - "5900:5900"
+
+  ld-sql:
+    build:
+      context: ld-sql
+    image: ld-sql
+    ports:
+      # Exposed to host so you can run MySQL admin tools if you want
+      # Note that this will *fail* if you have MySQL running on your host, as port 3306 will already be taken
+      - 3306:3306
+    env_file: ld-api/ld.env
+    container_name: ld-sql
+
+  ld-api:
+    build:
+      context: ld-api
+    image: ld-api
+    depends_on:
+      - ld-sql
+    ports:
+      - 3000:3000
+    # env_file should be same one given to ld-sql image
+    env_file: ld-api/ld.env
+    container_name: ld-api
 
   e2e-ui:
     build:
--- a/docker/ld-api/Dockerfile
+++ b/docker/ld-api/Dockerfile
@@ -1,0 +1,38 @@
+# Should be built with `docker build . --tag ld-api`
+# Then run with something like `docker run --init --env-file=.env --network=host ld-api:latest`
+#
+# This assumes a .env file that looks something like this:
+# PORT=3000
+# MYSQL_HOST=localhost
+# MYSQL_USER=mysqlusername
+# MYSQL_PASSWORD=mysqlpassword
+# MYSQL_DATABASE=testldapi
+# MYSQL_DATABASE_PRIVATE=testldapipvt
+
+# Step 1: Check out source
+FROM alpine AS git
+RUN apk --update add git && \
+    rm -rf /var/lib/apt/lists/* && \
+    rm /var/cache/apk/*
+RUN git clone --depth=1 https://github.com/sillsdev/web-languagedepot-api/ git
+
+# Step 2: Build app
+FROM node:14.15.4-alpine AS builder
+WORKDIR /app
+COPY --from=git /git/package*.json ./
+RUN npm ci
+COPY --from=git /git/static ./static/
+COPY --from=git /git/svelte.config.js /git/snowpack.config.js /git/tsconfig.json ./
+COPY --from=git /git/src ./src/
+RUN npm run build && npm run adapt
+
+# Step 3: Create minimal container with running app
+FROM node:14.15.4-alpine
+WORKDIR /app
+COPY --from=git /git/static assets
+COPY --from=git /git/package*.json ./
+RUN npm ci
+COPY --from=builder app/build ./
+EXPOSE 3000
+USER node
+ENTRYPOINT ["node", "index.js"]
--- a/docker/ld-api/Dockerfile
+++ b/docker/ld-api/Dockerfile
@@ -1,6 +1,3 @@
-# Should be built with `docker build . --tag ld-api`
-# Then run with something like `docker run --init --env-file=.env --network=host ld-api:latest`
-#
 # This assumes a .env file that looks something like this:
 # PORT=3000
 # MYSQL_HOST=localhost
@@ -9,30 +6,23 @@
 # MYSQL_DATABASE=testldapi
 # MYSQL_DATABASE_PRIVATE=testldapipvt
 
-# Step 1: Check out source
-FROM alpine AS git
-RUN apk --update add git && \
-    rm -rf /var/lib/apt/lists/* && \
-    rm /var/cache/apk/*
-RUN git clone --depth=1 https://github.com/sillsdev/web-languagedepot-api/ git
-
-# Step 2: Build app
+# Step 1: Build app
 FROM node:14.15.4-alpine AS builder
-WORKDIR /app
-COPY --from=git /git/package*.json ./
+WORKDIR /git
+ADD https://github.com/sillsdev/web-languagedepot-api/tarball/nodejs ./
+# Github tarball has top-level directory USER-REPO-HASH, and HASH is unpredictable ahead of time, so we rename it
+RUN tar xvzf nodejs && rm nodejs && mv sillsdev-web-languagedepot-api-* ldapi
+WORKDIR /git/ldapi
 RUN npm ci
-COPY --from=git /git/static ./static/
-COPY --from=git /git/svelte.config.js /git/snowpack.config.js /git/tsconfig.json ./
-COPY --from=git /git/src ./src/
 RUN npm run build && npm run adapt
 
-# Step 3: Create minimal container with running app
+# Step 2: Create minimal container with running app
 FROM node:14.15.4-alpine
 WORKDIR /app
-COPY --from=git /git/static assets
-COPY --from=git /git/package*.json ./
+COPY --from=builder /git/ldapi/static assets
+COPY --from=builder /git/ldapi/package*.json ./
 RUN npm ci
-COPY --from=builder app/build ./
+COPY --from=builder /git/ldapi/build ./
 EXPOSE 3000
 USER node
 ENTRYPOINT ["node", "index.js"]
--- a/docker/ld-api/ld.env
+++ b/docker/ld-api/ld.env
@@ -1,0 +1,6 @@
+MYSQL_HOST=ld-sql
+MYSQL_DATABASE=testldapi
+MYSQL_DATABASE_PRIVATE=testldapipvt
+MYSQL_USER=redmine
+MYSQL_PASSWORD=not-a-secret
+MYSQL_ROOT_PASSWORD=also-not-a-secret
--- a/docker/ld-sql/Dockerfile
+++ b/docker/ld-sql/Dockerfile
@@ -1,0 +1,2 @@
+FROM mysql:5.7
+COPY testlanguagedepot.sql /docker-entrypoint-initdb.d/
--- a/docker/ld-sql/testlanguagedepot.sql
+++ b/docker/ld-sql/testlanguagedepot.sql
@@ -518,7 +518,7 @@ CREATE TABLE `email_addresses` (
 LOCK TABLES `email_addresses` WRITE;
 /*!40000 ALTER TABLE `email_addresses` DISABLE KEYS */;
 INSERT INTO `email_addresses` (`id`, `user_id`, `address`, `is_default`, `notify`, `created_on`, `updated_on`) VALUES
-(1, 1,     'admin@example.net',          1, 1, '2009-07-22 06:32:07', '2009-07-23 08:45:37'),
+(1, 1,     'admin@example.com',          1, 1, '2009-07-22 06:32:07', '2009-07-23 08:45:37'),
 (2, 3,     'king.richard@example.com',   1, 1, '2019-10-04 13:27:30', '2019-10-04 13:27:30'),
 (3, 4,     'princejohn@example.com',     1, 1, '2019-10-04 13:27:30', '2019-10-04 13:27:30'),
 (4, 10,    'manager1@example.net',       1, 1, '2009-07-22 06:32:07', '2009-07-23 08:45:37'),
--- a/docker/ld-sql/testlanguagedepot.sql
+++ b/docker/ld-sql/testlanguagedepot.sql
@@ -1,0 +1,1849 @@
+-- MySQL dump 10.13  Distrib 5.7.27, for Linux (x86_64)
+--
+-- Host: localhost    Database: languagedepot
+-- ------------------------------------------------------
+-- Server version	5.7.27-0ubuntu0.18.04.1
+
+SET SQL_MODE = "NO_AUTO_VALUE_ON_ZERO";
+SET time_zone = "+00:00";
+
+/*!40101 SET @OLD_CHARACTER_SET_CLIENT=@@CHARACTER_SET_CLIENT */;
+/*!40101 SET @OLD_CHARACTER_SET_RESULTS=@@CHARACTER_SET_RESULTS */;
+/*!40101 SET @OLD_COLLATION_CONNECTION=@@COLLATION_CONNECTION */;
+/*!40101 SET NAMES utf8 */;
+/*!40103 SET @OLD_TIME_ZONE=@@TIME_ZONE */;
+/*!40103 SET TIME_ZONE='+00:00' */;
+/*!40014 SET @OLD_UNIQUE_CHECKS=@@UNIQUE_CHECKS, UNIQUE_CHECKS=0 */;
+/*!40014 SET @OLD_FOREIGN_KEY_CHECKS=@@FOREIGN_KEY_CHECKS, FOREIGN_KEY_CHECKS=0 */;
+/*!40101 SET @OLD_SQL_MODE=@@SQL_MODE, SQL_MODE='NO_AUTO_VALUE_ON_ZERO' */;
+/*!40111 SET @OLD_SQL_NOTES=@@SQL_NOTES, SQL_NOTES=0 */;
+
+--
+-- Table structure for table `ar_internal_metadata`
+--
+
+DROP TABLE IF EXISTS `ar_internal_metadata`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `ar_internal_metadata` (
+  `key` varchar(255) NOT NULL,
+  `value` varchar(255) DEFAULT NULL,
+  `created_at` datetime NOT NULL,
+  `updated_at` datetime NOT NULL,
+  PRIMARY KEY (`key`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+/*!40101 SET character_set_client = @saved_cs_client */;
+
+--
+-- Dumping data for table `ar_internal_metadata`
+--
+
+LOCK TABLES `ar_internal_metadata` WRITE;
+/*!40000 ALTER TABLE `ar_internal_metadata` DISABLE KEYS */;
+INSERT INTO `ar_internal_metadata` (`key`, `value`, `created_at`, `updated_at`) VALUES
+('environment','production_languagedepot','2019-10-04 14:08:08','2019-10-04 14:08:08');
+/*!40000 ALTER TABLE `ar_internal_metadata` ENABLE KEYS */;
+UNLOCK TABLES;
+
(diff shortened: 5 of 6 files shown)


megahirt:
@rmunn what needs to be done with this branch in order for it to be ready for review?

rmunn:
@megahirt Thanks to some help from @longrunningprocess yesterday, I managed to get this working. The key was realizing that it was the URL for the Language Depot API is really config, not code. It's okay to have different config in the dev environment vs. in production, we just don't want different *code*. But I can have one URL in the dev environment, which will point to the hostname created by docker-compose, and a different URL in the QA and production environments, and that's fine since it's all the same code being run. So I didn't need to do trickery with /etc/hosts files or Caddy config to have the name "admin.languagedepot.org" map to a different IP address in the dev environment; all I needed was to pass in the appropriate base URL via an environment variable, which is standard Docker practice.

So this is ready to review and merge now. I've tested it locally and it gets rid of the "Oh. Error." popups.

megahirt:
This one failing E2E test looks suspicious: **Explore Site Admin Page**
```
Browser Console JS Error: 
http://e2e/dist/vendor.bundle.js?v=9.9.9 45084:40 "Error: data.slice is not a function (reported to Bugsnag)"
```
It's helpful to include an explicit number of failing tests in the PR, since the status report doesn't declare how many failing tests.  This PR has **42 E2E failures**, as reported by TC.  Steadily decreasing!!!
image

megahirt (changes requested):
All the code looks good to me, however there is a suspicious E2E test that I'd like you to investigate before approval.  Otherwise, this looks good to me.

> Reviewed 5 of 6 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-:-MUvTflPATunr_ym7Qni:byozw11)* status: all files reviewed, 2 unresolved discussions (waiting on @rmunn)

___
*[docker/docker-compose.yml, line 114 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-MUvEJ0q4yyLPj73vZ-X:-MUvEJ0q4yyLPj73vZ-Y:b17fzc4) ([raw file](https://github.com/sillsdev/web-languageforge/blob/261674af7a0d3e71aa043e16082fa9928c8f2dbe/docker/docker-compose.yml#L114)):*
> ```YAML
>       - ld-sql
>     ports:
>       - 3000:3000
> ```

After a very helpful in-office discussion around maintaining development vs. production configuration in our docker-compose file, my current thinking is that docker-compose will remain a developer-centric service definition that serves our development workflow.  This includes exposed ports to the host, debug images (I'm looking at you, selenium), volume maps to source code, etc.

Our production docker configuration will necessarily be written in kubernetes format (YAML?) and define what services actually run in production.

I think our strategy is to keep our Dockerfiles free of developer-centric statements such that we are running the exact same docker containers on developer machines and build agents as we are running in production.  And that's how we achieve 95% sameness between our development and production environments.  The remaining 5% is basically the differences that we have between our docker-compose and Kubernetes service definitions.

Looking forward to getting Kubernetes config going in a future PR.
(I will re-post this in Slack)
___
*[docker/ld-api/Dockerfile, line 29 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-MUvOIaU1IePy6ybz3sI:-MUvOIaU1IePy6ybz3sJ:b-ud5wvs) ([raw file](https://github.com/sillsdev/web-languageforge/blob/261674af7a0d3e71aa043e16082fa9928c8f2dbe/docker/ld-api/Dockerfile#L29)):*
> ```Dockerfile
> RUN npm run build && npm run adapt
> 
> # Step 3: Create minimal container with running app
> ```

This is really good.  I learned a lot from your Dockerfile, much of which I presume is best practice (using small containers like alpine, multi-stage builds, locking to specific versions for production, etc).

@rmunn and I talked in the office and we decided that it's appropriate to check-in this ld-api file into web-languageforge, even though this is copy/paste from the ld-api repo itself.  Future work includes publishing an ld-api Docker image somewhere publicly and then referencing that image in our docker-compose file.  That way, this code isn't duplicated.  For now this gets us up and running.
___
*[src/Api/Service/Ldapi.php, line 77 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-MUvT-cX0oCsmjsg30cU:-MUvT-cX0oCsmjsg30cV:b-s2828o) ([raw file](https://github.com/sillsdev/web-languageforge/blob/261674af7a0d3e71aa043e16082fa9928c8f2dbe/src/Api/Service/Ldapi.php#L77)):*
> ```
>         $baseUrl = getenv('LDAPI_BASE_URL');
>         if (!$baseUrl) { $baseUrl = ; }
>         // Note that getenv('LDAPI_BASE_URL') || 'default' fails because getenv can return boolean, so PHP evaluates that line to 1 instead of the string value of getenv
> ```

Perhaps this would work?
```
$baseUrl = getenv('LDAPI_BASE_URL') ?: 'https://admin.languagedepot.org/api/v2/';
```

From https://www.php.net/manual/en/language.operators.comparison.php (search for "Ternary Operator"):
>The expression (expr1) ? (expr2) : (expr3) evaluates to expr2 if expr1 evaluates to true, and expr3 if expr1 evaluates to false.
It is possible to leave out the middle part of the ternary operator. Expression expr1 ?: expr3 returns expr1 if expr1 evaluates to true, and expr3 otherwise.

rmunn:
The E2E failure is being caused by an error coming back from the LD API server: "No route found for /api/v2/projects" (because the site-admin page tries to load projects from LD API now as well as from LF, hence why this failure is new). And the reason for those failures was because I added an environment variable for the LF API to know where to find the LD API, but I forgot to add it to the `e2e-api` definition. So all the LD API calls were failing, such as the `/api/v2/projects` call that gets made on the siteadmin page. Which wasn't causing any visible problems on the page, but was causing console logging of errors, which in turn caused the E2E failures.

Running locally, I can prove that this fixes the suspicious E2E test that you mentioned, @megahirt. The TC build is showing 43 errors, including 2 new ones, but my local build only has 30 E2E errors. And when I look at the 2 new errors on TC, they appear to be unrelated. I believe this is ready to merge.

longrunningprocess:
nice!

longrunningprocess:
> Our production docker configuration will necessarily be written in kubernetes format (YAML?) and define what services actually run in production.
> 
> I think our strategy is to keep our Dockerfiles free of developer-centric statements such that we are running the exact same docker containers on developer machines and build agents as we are running in production. And that's how we achieve 95% sameness between our development and production environments. The remaining 5% is basically the differences that we have between our docker-compose and Kubernetes service definitions.

absolutely!  Seems like a grok...I know we had hoped to begin talking through deployment but this is a key point, the images are the app with clean api's, the dc.yml, kub def, aws task defintion, etc.,  are the configs for each runtime environment.  This is an excellent realization to come to on our own. 🧠

rmunn:
@megahirt Want to approve this in Reviewable? I believe I've addressed all the issues you raised.

rmunn:
Just rebased this PR on top of current master. At this time, current master builds with 18 failing E2E tests. (15 failing E2E tests once #860 is merged). With this PR, we go down to 17 failing E2E tests (which I expect to be 14 once #860 is merged). The one test that is fixed is the "create: new empty project" test in lexicon-new-project.e2e-spec.ts. Before this PR, that failed with "Failed to load resource: the server responded with a status of 403 (Forbidden)" on the URL `http://e2e/upload/lf-lexicon/import-zip`. I'm unclear why that test passed, because I didn't think its failure was due to an LD API call not having a server to respond to. So it's possible that this PR does not fix any currently broken tests — but it also doesn't break any existing tests.

rmunn:
And now that #860 is merged, I've rebased again. This time I expect to see 14 failing tests.

rmunn:
:heavy_check_mark: All tests pass.

rmunn (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/831)* status: 3 of 6 files reviewed, all discussions resolved (waiting on @megahirt)

___
*[src/Api/Service/Ldapi.php, line 77 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-MUvT-cX0oCsmjsg30cU:-MVLPgw54DOltbAfGj49:b-kff1bt) ([raw file](https://github.com/sillsdev/web-languageforge/blob/261674af7a0d3e71aa043e16082fa9928c8f2dbe/src/Api/Service/Ldapi.php#L77)):*

> ```
>         $baseUrl = getenv('LDAPI_BASE_URL');
>         if (!$baseUrl) { $baseUrl = ; }
>         // Note that getenv('LDAPI_BASE_URL') || 'default' fails because getenv can return boolean, so PHP evaluates that line to 1 instead of the string value of getenv
> ```

Perhaps this would work?
```
$baseUrl = getenv('LDAPI_BASE_URL') ?: 'https://admin.languagedepot.org/api/v2/';
```

From https://www.php.net/manual/en/language.operators.comparison.php (search for "Ternary Operator"):
>The expression (expr1) ? (expr2) : (expr3) evaluates to expr2 if expr1 evaluates to true, and expr3 if expr1 evaluates to false.
It is possible to leave out the middle part of the ternary operator. Expression expr1 ?: expr3 returns expr1 if expr1 evaluates to true, and expr3 otherwise.
</blockquote></details>

Done in latest commit (8285212fb).

longrunningprocess (dismissed):
> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/831)* status: 4 of 6 files reviewed, 8 unresolved discussions (waiting on @megahirt and @rmunn)

___
*[docker/docker-compose.yml, line 34 at r4](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-MVMYUxT7MBr3UL066w3:-MVMYUxT7MBr3UL066w4:b-tgv6hn) ([raw file](https://github.com/sillsdev/web-languageforge/blob/8285212fbeb70bfb097322b229da3237d0dad0a8/docker/docker-compose.yml#L34)):*
> ```YAML
>     environment:
>       - WAIT_HOSTS=db:27017, mail:25
>       - "LDAPI_BASE_URL=http://ld-api:3000/api/v2/"
> ```

I'd be curious if simply referencing `http://ld-api/api/v2/` would work?  There are places where we can rely on docker internal networking to work as defined, i.e., in some places you don't need to make explicit reference to the port...
___
*[docker/docker-compose.yml, line 107 at r4](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-MVMXOlpCJui7Iov_1Yi:-MVMXOlpCJui7Iov_1Yj:b-xkixsc) ([raw file](https://github.com/sillsdev/web-languageforge/blob/8285212fbeb70bfb097322b229da3237d0dad0a8/docker/docker-compose.yml#L107)):*
> ```YAML
>     #   - 3306:3306
>     env_file: ld-api/ld.env
>     container_name: ld-sql
> ```

we've sort of ended up with a common convention with these, mind moving this under `image` to follow the pattern?
___
*[docker/docker-compose.yml, line 118 at r4](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-MVMXkfBAX1pW-Lu1FMW:-MVMXkfBAX1pW-Lu1FMX:bl5cdgv) ([raw file](https://github.com/sillsdev/web-languageforge/blob/8285212fbeb70bfb097322b229da3237d0dad0a8/docker/docker-compose.yml#L118)):*
> ```YAML
>     env_file: ld-api/ld.env
>     environment:
>       - PORT=3000
> ```

`PORT` has me asking, which port...for this container, for another container, etc?  Mind if we provide a little more info on this var name to help understand it better?  For example, if it's for the api container, maybe something like `LF_API_PORT` or something.
___
*[docker/docker-compose.yml, line 119 at r4](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-MVMXYNT7ziwo80z7k0q:-MVMXYNT7ziwo80z7k0r:b-e4wcb1) ([raw file](https://github.com/sillsdev/web-languageforge/blob/8285212fbeb70bfb097322b229da3237d0dad0a8/docker/docker-compose.yml#L119)):*
> ```YAML
>     environment:
>       - PORT=3000
>     container_name: ld-api
> ```

move under `image` to follow convention that's begun emerging in this file.
___
*[docker/docker-compose.yml, line 156 at r4](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-MVMZ32_AqY6ClKWWzJl:-MVMZ32_AqY6ClKWWzJm:b6fw20q) ([raw file](https://github.com/sillsdev/web-languageforge/blob/8285212fbeb70bfb097322b229da3237d0dad0a8/docker/docker-compose.yml#L156)):*
> ```YAML
>     environment:
>       - WAIT_HOSTS=db:27017, mail:25
>       - "LDAPI_BASE_URL=http://ld-api:3000/api/v2/"
> ```

can we get away with `http://ld-api/api/v2/` here?
___
*[docker/ld-api/Dockerfile, line 26 at r4](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-MVMZNcu3DFTE-2PyYn1:-MVMZNcw6e2b4ofP8dvB:b-cyw66h) ([raw file](https://github.com/sillsdev/web-languageforge/blob/8285212fbeb70bfb097322b229da3237d0dad0a8/docker/ld-api/Dockerfile#L26)):*
> ```Dockerfile
> RUN npm ci
> COPY --from=builder /git/ldapi/build ./
> EXPOSE 3000
> ```

REALLY nice Dockerfile!

does this have to be the same as the `PORT` env var?
___
*[docker/ld-api/ld.env, line 1 at r4](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-MVMZvCw-RY9LN87ykDA:-MVMZvCw-RY9LN87ykDB:b4f4cy) ([raw file](https://github.com/sillsdev/web-languageforge/blob/8285212fbeb70bfb097322b229da3237d0dad0a8/docker/ld-api/ld.env#L1)):*
> ```Shell
> MYSQL_HOST=ld-sql
> ```

A typical pattern for something like this would be to provide a `.env.example` and then provide either a script or instructions to `cp .env.example .env` and populate the local `.env` with appropriate values.  Most importantly though is to provide a `.gitignore` here that ignores the `.env` with real values so we don't ever have someone accidentally commit real secrets.
___
*[docker/ld-sql/Dockerfile, line 2 at r4](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-MVMa3MbBRnhLRFh_L5V:-MVMa3Me4eZbe_M4OBwt:b-kltjjm) ([raw file](https://github.com/sillsdev/web-languageforge/blob/8285212fbeb70bfb097322b229da3237d0dad0a8/docker/ld-sql/Dockerfile#L2)):*
> ```Dockerfile
> FROM mysql:5.7
> COPY testlanguagedepot.sql /docker-entrypoint-initdb.d/
> ```

slick!
___
*[src/Api/Service/Ldapi.php, line 77 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-MUvT-cX0oCsmjsg30cU:-MVMaH9z7bQWJ8fYA5fM:b-3otyst) ([raw file](https://github.com/sillsdev/web-languageforge/blob/261674af7a0d3e71aa043e16082fa9928c8f2dbe/src/Api/Service/Ldapi.php#L77)):*

Done in latest commit (8285212fb).
</blockquote></details>

I much prefer this technique, thank you!

rmunn:
There's one E2E test failing now, which I've determined is flaky and will be fixed by https://github.com/sillsdev/web-languageforge/pull/866.

rmunn (commented):
> Dismissed @longrunningprocess from 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/831)* status: 4 of 6 files reviewed, 1 unresolved discussion (waiting on @longrunningprocess and @megahirt)

___
*[docker/docker-compose.yml, line 34 at r4](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-MVMYUxT7MBr3UL066w3:-MVULuTrAQn9py0jzBXT:b4lxzh8) ([raw file](https://github.com/sillsdev/web-languageforge/blob/8285212fbeb70bfb097322b229da3237d0dad0a8/docker/docker-compose.yml#L34)):*

I'd be curious if simply referencing `http://ld-api/api/v2/` would work?  There are places where we can rely on docker internal networking to work as defined, i.e., in some places you don't need to make explicit reference to the port...
</blockquote></details>

I don't think that will work without having the ld-api server process run on port 80, and since it runs non-privileged it can't listen on port 80. So to do that I'd have to set up a caddy reverse proxy, and I don't think that's worth the effort. I'll try it at some point, but I think I don't want to put that in this PR; I'll do a separate PR to tweak the port allocation.
___
*[docker/docker-compose.yml, line 107 at r4](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-MVMXOlpCJui7Iov_1Yi:-MVUN7M42HynTGCyhajG:b-896fix) ([raw file](https://github.com/sillsdev/web-languageforge/blob/8285212fbeb70bfb097322b229da3237d0dad0a8/docker/docker-compose.yml#L107)):*

we've sort of ended up with a common convention with these, mind moving this under `image` to follow the pattern?
</blockquote></details>

Done.
___
*[docker/docker-compose.yml, line 118 at r4](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-MVMXkfBAX1pW-Lu1FMW:-MVUMHwg1ipMeJKEDd00:bmggx7o) ([raw file](https://github.com/sillsdev/web-languageforge/blob/8285212fbeb70bfb097322b229da3237d0dad0a8/docker/docker-compose.yml#L118)):*

`PORT` has me asking, which port...for this container, for another container, etc?  Mind if we provide a little more info on this var name to help understand it better?  For example, if it's for the api container, maybe something like `LF_API_PORT` or something.
</blockquote></details>

The name `PORT` is required by the underlying library I'm using and I can't change the name. I could write a wrapper shell script into the Dockerfile that sets `PORT=$LD_API_PORT` or something, but I think a comment in the docker-compose.yml file is a better approach.
___
*[docker/docker-compose.yml, line 119 at r4](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-MVMXYNT7ziwo80z7k0q:-MVUN8QQDsmfbuCDAHX8:b-896fix) ([raw file](https://github.com/sillsdev/web-languageforge/blob/8285212fbeb70bfb097322b229da3237d0dad0a8/docker/docker-compose.yml#L119)):*

move under `image` to follow convention that's begun emerging in this file.
</blockquote></details>

Done.
___
*[docker/docker-compose.yml, line 156 at r4](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-MVMZ32_AqY6ClKWWzJl:-MVUORyj4n7geWgsv9df:b-kkex5y) ([raw file](https://github.com/sillsdev/web-languageforge/blob/8285212fbeb70bfb097322b229da3237d0dad0a8/docker/docker-compose.yml#L156)):*

can we get away with `http://ld-api/api/v2/` here?
</blockquote></details>

I think I'd have to set up a caddy reverse proxy to do that; I don't think Docker will automatically send the traffic to port 3000 if I omit the port here. I'll test that, but in a separate PR; this one is working, and I don't want to make changes to the working PR. It's been two weeks already; I want to get it merged and then make style tweaks later.
___
*[docker/ld-api/Dockerfile, line 26 at r4](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-MVMZNcu3DFTE-2PyYn1:-MVUOl629sTRHBPNJXJM:b3w613z) ([raw file](https://github.com/sillsdev/web-languageforge/blob/8285212fbeb70bfb097322b229da3237d0dad0a8/docker/ld-api/Dockerfile#L26)):*

REALLY nice Dockerfile!

does this have to be the same as the `PORT` env var?
</blockquote></details>

Yes. Actually, since the Node server will run on whatever port is specified in `$PORT`, this should either be `EXPOSE $PORT` or else moved into the docker-compose.yml file as `expose: 3000`. But I'll do that in a later PR.
___
*[docker/ld-api/ld.env, line 1 at r4](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-MVMZvCw-RY9LN87ykDA:-MVUQAQm0hz6nylEa-ce:b-ky426u) ([raw file](https://github.com/sillsdev/web-languageforge/blob/8285212fbeb70bfb097322b229da3237d0dad0a8/docker/ld-api/ld.env#L1)):*

A typical pattern for something like this would be to provide a `.env.example` and then provide either a script or instructions to `cp .env.example .env` and populate the local `.env` with appropriate values.  Most importantly though is to provide a `.gitignore` here that ignores the `.env` with real values so we don't ever have someone accidentally commit real secrets.
</blockquote></details>

I plan for the real values to be set via Kubenetes config files in our deployment setup, so the ld.env files here are never intended to become `.env` files that the app actually looks at. I can rename this file to `ld.env.testing` if you like, but I'd prefer to do that in a separate PR so that this one doesn't drag out even longer.

When you talk about a script to `cp .env.example .env`, do you mean in the Makefile before the `docker build` step? Or do you mean in the Dockerfile? If it's the latter, the `mysql` image doesn't actually use the `.env` pattern. The reason I'm putting these in a `.env` is to have a single source of truth for the `ld-sql` and `ld-api` containers to share the same MySQL username, password, and DB name(s).
___
*[docker/ld-sql/Dockerfile, line 2 at r4](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-MVMa3MbBRnhLRFh_L5V:-MVUPqlk5dEaO59e-Qx9:bb31ru5) ([raw file](https://github.com/sillsdev/web-languageforge/blob/8285212fbeb70bfb097322b229da3237d0dad0a8/docker/ld-sql/Dockerfile#L2)):*

slick!
</blockquote></details>

Can't take credit for this myself. The entrypoint script for the `mysql` Docker image uses these environment variables at startup time to create a database, set up its permissions, and populate it from the `*.sql` files in `/docker-entrypoint-initdb.d/` when the container runs — but ONLY if the MySQL data directory is empty. Since I'm defining these container images for the purpose of testing, I'm not defining any volumes, which means that the data directory will be ephemeral and so the SQL script that populates the test database will run each time the container starts up.

megahirt (approved):
:lgtm: This looks good to me.  Learned some things too!

> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-:-MVUTcfh6SsNBTck98uN:bgp29pq)* status: all files reviewed, 1 unresolved discussion (waiting on @longrunningprocess)

megahirt (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/831)* status: all files reviewed, 1 unresolved discussion (waiting on @longrunningprocess)

rmunn (commented):
> Dismissed @longrunningprocess from a discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/831)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rmunn)

longrunningprocess (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/831)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[docker/docker-compose.yml, line 34 at r4](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-MVMYUxT7MBr3UL066w3:-MVWXS-k9OCQFSXrRMLh:b36jnd1) ([raw file](https://github.com/sillsdev/web-languageforge/blob/8285212fbeb70bfb097322b229da3237d0dad0a8/docker/docker-compose.yml#L34)):*

I don't think that will work without having the ld-api server process run on port 80, and since it runs non-privileged it can't listen on port 80. So to do that I'd have to set up a caddy reverse proxy, and I don't think that's worth the effort. I'll try it at some point, but I think I don't want to put that in this PR; I'll do a separate PR to tweak the port allocation.
</blockquote></details>

no need to jump through any hoops, I just know there is a lot of intelligence in the Docker network and there have been times I thought I need to reference the port explicitly only to learn that by the using the service alias all of that was understood.  I bring this up because it eliminates a "maintenance" should the port ever change.  Like I said, just curious but what you're saying certainly seems intuitive.
___
*[docker/docker-compose.yml, line 118 at r4](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-MVMXkfBAX1pW-Lu1FMW:-MVWYJgx2D_21bse-Ogw:bpprjws) ([raw file](https://github.com/sillsdev/web-languageforge/blob/8285212fbeb70bfb097322b229da3237d0dad0a8/docker/docker-compose.yml#L118)):*

The name `PORT` is required by the underlying library I'm using and I can't change the name. I could write a wrapper shell script into the Dockerfile that sets `PORT=$LD_API_PORT` or something, but I think a comment in the docker-compose.yml file is a better approach.
</blockquote></details>

seems like overkill, a comment would be fine, thank you
___
*[docker/ld-api/ld.env, line 1 at r4](https://reviewable.io/reviews/sillsdev/web-languageforge/831#-MVMZvCw-RY9LN87ykDA:-MVWYSg09xRdjsWoSQRu:bbfwdbw) ([raw file](https://github.com/sillsdev/web-languageforge/blob/8285212fbeb70bfb097322b229da3237d0dad0a8/docker/ld-api/ld.env#L1)):*

I plan for the real values to be set via Kubenetes config files in our deployment setup, so the ld.env files here are never intended to become `.env` files that the app actually looks at. I can rename this file to `ld.env.testing` if you like, but I'd prefer to do that in a separate PR so that this one doesn't drag out even longer.

When you talk about a script to `cp .env.example .env`, do you mean in the Makefile before the `docker build` step? Or do you mean in the Dockerfile? If it's the latter, the `mysql` image doesn't actually use the `.env` pattern. The reason I'm putting these in a `.env` is to have a single source of truth for the `ld-sql` and `ld-api` containers to share the same MySQL username, password, and DB name(s).
</blockquote></details>

gotcha, that makes sense, thank you.

PR #837, opened 2021-02-25
Fix reset password part of E2E traversal test

rmunn:
The failure on the reset password part of the E2E traversal test is because the traversal test is designed to be run first, before anything else, but the test locator logic is running tests in subdirectories before it runs tests in the parent directory, so that the test/app/bellows/bellows-traversal.e2e-spec.ts test is running *after* the test/app/bellows/reset-forgotten-password.e2e-spec.ts test. The bellows-traversal test needs to run first. Moving the reset-forgotten-password test into the parent directory fixes this issue.

This doesn't fix all the E2E tests, but we at least have fewer failures after this PR than before the PR. :-)

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/837)

### DIFF
--- a/test/app/bellows/reset-forgotten-password.e2e-spec.ts
+++ b/test/app/bellows/reset-forgotten-password.e2e-spec.ts
@@ -1,12 +1,12 @@
 import {browser, ExpectedConditions} from 'protractor';
 
-import {BellowsForgotPasswordPage} from '../shared/forgot-password.page';
-import {BellowsLoginPage} from '../shared/login.page';
-import {PageHeader} from '../shared/page-header.element';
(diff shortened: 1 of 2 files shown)


megahirt (approved):
I assume this only fixes the issue because *b* comes before *r* in the alphabet...

A different, more robust solution might have been to explicitly call the traversal tests in the list of specs to run which is defined in the gulpfile.  I'll accept this though for expediency.  Thank you.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/837#-:-MUU3CIyBKJPydQ6xiKt:bfobtr0)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #846, opened 2021-03-01
Fix calling /api/v2/roles in LD API

rmunn:
Roles API call doesn't need authorization and shouldn't fail if there's no username available.

I haven't been able to prove that this works yet, due to multiple unrelated errors in E2E tests. **Update:** I think it's working now. Ready to be reviewed and merged.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/846)

### DIFF
--- a/docker/e2e-api/run.sh
+++ b/docker/e2e-api/run.sh
@@ -2,4 +2,6 @@
 
 cd /var/www/test/app
 php setupTestEnvironment.php localhost
+chmod -R g+w /var/www/html/assets /var/www/html/cache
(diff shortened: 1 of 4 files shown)


rmunn:
With commit https://github.com/sillsdev/web-languageforge/pull/846/commits/daf0b907a17095ba3f13fb2c77ad6bbc3590f53f I think this is now ready for review; it seems to work on my machine.

rmunn:
Closing in favor of #849, which includes all these fixes. That way there's just one outstanding E2E fix PR instead of multiple ones, which should hopefully be less confusing.

PR #847, opened 2021-03-01
Fix most async problems in E2E tests

rmunn:
This fixes many (hopefully, all) E2E test failures caused by race conditions, e.g. a FooPage.get() call at the end of one test that isn't being awaited, so the next test starts before the get() call has actually completed and then the next test fails because the browser isn't yet on that page.

There are still quite a few E2E test failures remaining, but I believe those are all, or mostly, real failures.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/847)

### DIFF
--- a/test/app/bellows/project-settings.e2e-spec.ts
+++ b/test/app/bellows/project-settings.e2e-spec.ts
@@ -11,17 +11,17 @@ describe('Bellows E2E Project Settings app', () => {
   const projectsPage = new ProjectsPage();
   const settingsPage = new BellowsProjectSettingsPage();
 
-  it('Normal user cannot access projectSettings to a project of which the user is a member', () => {
-    loginPage.loginAsMember();
-    projectsPage.get();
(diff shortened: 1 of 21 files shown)


megahirt (approved):
wow, this is a massive amount of insertions...how long did this take you?  Would have taken me 2 hours of Ctrl-V or the like.  Thanks for this.  :lgtm: 

> Reviewed 21 of 21 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/847#-:-MUkdpOjAtKcRSmCHYBl:bwr2bj7)* status: :shipit: complete! all files reviewed, all discussions resolved

rmunn:
I was able to use VS Code's multiple-cursors feature to speed up a lot of the process, so it only took me a couple of hours *counting* all the time I spent waiting for E2E test runs to finish.

PR #849, opened 2021-03-02
Fix more async problems in E2E tests

rmunn:
Three fixes included:

- Fix LD API call that was causing "Oh. Error." messages during most tests, and several test failures
- Fix unsightly RPC errors in Chrome if it hangs around after you cancel E2E test execution (caused by the offline check trying to look up rights in the current project when the current project was a test project that had been deleted at the end of the test)
- Fix a few utility helper functions that weren't actually async even though they looked async (e.g., findEntryByLexeme, which I renamed to clickEntryByLexeme because then it can be called as `await editorPage.browse.clickEntryByLexeme('foo')` instead of `await (await editorPage.browse.findEntryByLexeme('foo')).click()`).

This PR supersedes #846 as it includes all of #846's fixes. I decided it was less confusing to put all the fixes into a single PR rather than have multiple PRs that need to be merged in before you can start working on real E2E errors.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/849)

### DIFF
--- a/src/Api/Model/Shared/Command/LdapiCommands.php
+++ b/src/Api/Model/Shared/Command/LdapiCommands.php
@@ -68,8 +68,8 @@ class LdapiCommands
         return ($result === 'Manager');
     }
 
-    public static function getAllRoles($languageDepotUsername) {
-        return Ldapi::call($languageDepotUsername, 'get', self::ROLES_BASE_URL . self::URL_PART_GET_ALL);
+    public static function getAllRoles() {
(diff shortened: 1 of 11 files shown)


megahirt (approved):
:lgtm: really good stuff.  Thanks Robin.

> Reviewed 11 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/849#-:-MUqowfz-GfiHTAZRhr7:b7afwtg)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #852, opened 2021-03-02
Chore/simplify docker cleanup and get some more helpful docs in place

longrunningprocess:
php tests are all still passing
⚠️  e2e tests are pretty flaky:
46 failures on TC `master` , 48 on `master` locally
50 failures on TC `chore/simplify-docker-cleanup`, 52 on `chore/simplify-docker-cleanup` locally

so it seems there are 2 extra failures locally than out on TC and it seems I've increased the failures with this PR, I'll keep looking to see if I can determine which failures are new and which 2 fail locally only.  Please let me know if things jump out to you in this PR that could be having some effect on the tests.

I tried figuring out where exactly the specs are being loaded but was getting a little lost in whether they were getting loaded in the webpack config or the gulpfile or somewhere else.  It would be really nice to have a way to limit the tests that are being run for troubleshooting purposes.

I'll just wait on this one until I can either pair or get some more experienced eyes on it because the number of failures changes each run for me locally, no need to keep burning cycles on it when I can be productive elsewhere.  Thanks for any help you can offer.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/852)

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ Other useful resources:
 
 ### Docker ###
 
-1. Install [Docker](https://www.docker.com/get-started)
+1. Install [Docker](https://www.docker.com/get-started) (Linux users will need some additional steps, please visit https://docs.docker.com/compose/install for info on install the engine and compose)
 1. Install [Make](https://www.gnu.org/software/make/).  This is actually optional but simplifies things a bit.
 1. Clone the repo:  `git clone https://github.com/sillsdev/web-languageforge`
 1. `cd web-languageforge/docker`
--- a/README.md
+++ b/README.md
@@ -102,6 +102,26 @@ Other useful resources:
 1. You should see a landing page, click "Login"
 1. Use `admin` and `password` to get in
 
+#### If you would like to run the E2E tests
+
+1. `make e2e`
+1. Individual test results will appear in your terminal but if you'd like to watch them in real-time, simply VNC into the running tests via `localhost:5900`, e.g., Mac OSX users simply `open vnc://localhost:5900` and use `secret` as the password.  Other operating systems may require installing a separate VNC Viewer tool.
+
+#### If you would like to run the unit tests
+
+1. `make tests`
+1. Test results will appear in your terminal
+
+#### Viewing logs
+
+1. `make logs` will show logs from any runnign contianers, results can be filtered by simply grepping, e.g., `make logs | grep lf-ui`
+
+#### Cleanup
+
+1. `make clean` is the most common, it shuts down and cleans up running containers
+1. less commonly, if you need to blow away shared artifacts from previous runs, simply `make clean-volumes`
+1. rarely needed but for a "start from scratch" environment, `make clean-powerwash`.
+
 #### Language Forge Configuration File ####
 
 If you are working with FLEx Send and Receive feature, manually edit the Language Forge config file
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ Other useful resources:
 
 #### If you would like to run the E2E tests
 
-1. `make e2e`
+1. `make build e2e`
 1. Individual test results will appear in your terminal but if you'd like to watch them in real-time, simply VNC into the running tests via `localhost:5900`, e.g., Mac OSX users simply `open vnc://localhost:5900` and use `secret` as the password.  Other operating systems may require installing a separate VNC Viewer tool.
 
 #### If you would like to run the unit tests
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -31,10 +31,10 @@ logs:
 
 .PHONY: clean
 clean:
-	docker-compose down
+	docker-compose down -t 1
 	docker system prune -f
 
 .PHONY: clean-volumes
 clean-volumes:
-	docker-compose down -v
+	docker-compose down -t 1 -v
 	docker system prune -f --volumes
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -31,8 +31,10 @@ logs:
 
 .PHONY: clean
 clean:
-	docker-compose down -v
+	docker-compose down
 	docker system prune -f
 
-  # by default volumes are not pruned.  To prune volumes as well, do:
-  # docker system prune -f --volumes
+.PHONY: clean-volumes
+clean-volumes:
+	docker-compose down -v
+	docker system prune -f --volumes
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -7,7 +7,7 @@ start: wait
 
 .PHONY: e2e
 e2e:
-  docker-compose build ui api
+	docker-compose build ui api
 	docker-compose run e2e-ui
 
 .PHONY: tests
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -3,7 +3,7 @@
 .PHONY: start
 start: wait
   # starts the entire runtime infrastructure
-	docker-compose up -d ssl && docker-compose logs -f ui
+	docker-compose up -d ssl
 
 .PHONY: e2e
 e2e:
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -6,7 +6,7 @@ start: wait
 	docker-compose up -d ssl
 
 .PHONY: e2e
-e2e: build
+e2e:
 	docker-compose run e2e-ui
 
 .PHONY: tests
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -6,7 +6,7 @@ start: wait
 	docker-compose up -d ssl
 
 .PHONY: e2e
-e2e:
+e2e: build
 	docker-compose run e2e-ui
 
 .PHONY: tests
@@ -38,3 +38,7 @@ clean:
 clean-volumes:
 	docker-compose down -t 1 -v
 	docker system prune -f --volumes
+
+.PHONY: clean-powerwash
+clean-powerwash: clean-volumes
+	docker-compose down -t 1 --rmi all
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -27,12 +27,8 @@ db-reset:
 
 .PHONY: clean
 clean:
-	docker-compose kill
-	docker-compose rm -f
+	docker-compose down -v
 	docker system prune -f
-
-    # remove lf-specific volumes
-	docker volume ls -q |grep ^docker_lf_ | xargs --no-run-if-empty docker volume rm
 
   # by default volumes are not pruned.  To prune volumes as well, do:
   # docker system prune -f --volumes
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -25,6 +25,10 @@ wait:
 db-reset:
 	docker-compose run db-reset
 
+.PHONY: logs
+logs:
+	docker-compose logs
+
 .PHONY: clean
 clean:
 	docker-compose down -v
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -7,6 +7,7 @@ start: wait
 
 .PHONY: e2e
 e2e:
+  docker-compose build ui api
 	docker-compose run e2e-ui
 
 .PHONY: tests
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -102,7 +102,6 @@ services:
       - e2e-api
       - mail
       - selenium
-      - ui
     environment:
       - WAIT_HOSTS=db:27017, mail:25, selenium:4444
     command: sh -c "/wait && /run.sh"
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -84,7 +84,9 @@ services:
   selenium:
     # image: selenium/standalone-chrome
     image: selenium/standalone-chrome-debug
+    # TODO: seems would should have a depends_on for the e2e-api
     links:
+      # TODO:  I wonder why this was necessary, I would've thought e2e-api was already available, maybe not?
       - "e2e-api:e2e"
     volumes:
       - /dev/shm:/dev/shm
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -11,7 +11,7 @@ services:
       - api
     volumes:
       # share dist folder between ui and api containers
-      - lf_webpack_dist_folder:/data/src/dist
+      - lf-webpack-dist-folder:/data/src/dist
 
       # for developer convenience
       - ../src:/data/src
@@ -33,10 +33,10 @@ services:
     command: sh -c "/wait && apache2-foreground"
     volumes:
       # share dist folder between ui and api containers
-      - lf_webpack_dist_folder:/var/www/src/dist
+      - lf-webpack-dist-folder:/var/www/src/dist
 
       # persist user data uploads and assets
-      - lf_user_data_assets:/var/www/src/assets
+      - lf-user-data-assets:/var/www/src/assets
 
       # for developer convenience
       - ../src/Api:/var/www/src/Api
@@ -51,8 +51,8 @@ services:
     depends_on:
       - ui
     volumes:
-      - lf_caddy_data:/data
-      - lf_caddy_config:/config
+      - lf-caddy-data:/data
+      - lf-caddy-config:/config
     # https://caddyserver.com/docs/command-line
     command: caddy reverse-proxy --from localhost --to api
     restart: unless-stopped
@@ -71,7 +71,7 @@ services:
       # exposed this to host for admin tools
       - 27017:27017
     volumes:
-      - lf_mongo_data:/data/db
+      - lf-mongo-data:/data/db
     restart: always
 
   # TODO: still needs to be built/run/tested, etc.
@@ -107,7 +107,7 @@ services:
     command: sh -c "/wait && /run.sh"
     volumes:
       # share dist folder between ui and api containers
-      - lf_webpack_dist_folder:/data/src/dist
+      - lf-webpack-dist-folder:/data/src/dist
 
   e2e-api:
     build:
@@ -123,7 +123,7 @@ services:
     command: sh -c "/wait && /run.sh"
     volumes:
       # share dist folder between ui and api containers
-      - lf_webpack_dist_folder:/var/www/src/dist
+      - lf-webpack-dist-folder:/var/www/src/dist
 
   test-api:
     # TODO: get XDebug working with VSCode
@@ -154,8 +154,8 @@ services:
       - db
 
 volumes:
-  lf_caddy_config:
-  lf_caddy_data:
-  lf_mongo_data:
-  lf_user_data_assets:
-  lf_webpack_dist_folder:
+  lf-caddy-config:
+  lf-caddy-data:
+  lf-mongo-data:
+  lf-user-data-assets:
+  lf-webpack-dist-folder:
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -102,6 +102,7 @@ services:
       - e2e-api
       - mail
       - selenium
+      - ui
     environment:
       - WAIT_HOSTS=db:27017, mail:25, selenium:4444
     command: sh -c "/wait && /run.sh"
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -111,6 +111,12 @@ services:
       # share dist folder between ui and api containers
       - lf-webpack-dist-folder:/data/src/dist
 
+      # for developer convenience
+      - ../src:/data/src
+      - ../typings:/data/typings
+      - ../webpack.config.js:/data/webpack.config.js
+      - ../gulpfile.js:/data/gulpfile.js
+
   e2e-api:
     build:
       context: ..


megahirt (changes requested):
Thanks for the helpful PR to add to the README and docker cleanup.  More discussion needed - see comments below.

> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-:-MUqnFqC5oA8MruMqRh3:b-nhzlpw)* status: all files reviewed, 3 unresolved discussions (waiting on @longrunningprocess and @rmunn)

___
*[README.md, line 105 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-MUqnB4aEGi2tVPRIueH:-MUqnB4aEGi2tVPRIueI:buoay4k) ([raw file](https://github.com/sillsdev/web-languageforge/blob/11125dc891c260548a83f371e3479635b88a902f/README.md#L105)):*
> ```Markdown
> 1. Use `admin` and `password` to get in
> 
> #### If you would like to run the E2E tests
> ```

Thanks for adding instructions to the README!!!
___
*[docker/docker-compose.yml, line 105 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-MUqmUR37sk91PxjJfXg:-MUqmUR37sk91PxjJfXh:b-foausn) ([raw file](https://github.com/sillsdev/web-languageforge/blob/11125dc891c260548a83f371e3479635b88a902f/docker/docker-compose.yml#L105)):*
> ```YAML
>       - mail
>       - selenium
>       - ui
> ```

I don't understand why `e2e-ui` depends on `ui`.  I would guess this would cause problems.  They are mutually exclusive containers, not designed to be run at the same time.
___
*[docker/docker-compose.yml, line 158 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-MUqlAMt9H7vr3zo-7v-:-MUqlAMt9H7vr3zo-7v0:bz5m7zh) ([raw file](https://github.com/sillsdev/web-languageforge/blob/11125dc891c260548a83f371e3479635b88a902f/docker/docker-compose.yml#L158)):*
> ```YAML
> 
> volumes:
>   lf-caddy-config:
> ```

why are we going back to kebab case?  Does it look better to you?  Docker uses _ for its volume names, but its a moot point since we don't actually need to identify the volumes directly.  If we're going to change back to kebab case (which I'm fine with), we might as well remove the "lf-" since that was only necessary to identify the LF docker volumes among others you (or the CI agent) have on their system.  The `docker-compose down -v` solution is much preferred
___
*[docker/Makefile, line 6 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-MUqn26-0-e9al6OFEik:-MUqn26-0-e9al6OFEil:bnzctbz) ([raw file](https://github.com/sillsdev/web-languageforge/blob/ce3c3708297a17c38ac7898732e54559b6c29b1b/docker/Makefile#L6)):*
> ```Makefile
> start: wait
>   # starts the entire runtime infrastructure
> 	docker-compose up -d ssl && docker-compose logs -f ui
> ```

thanks for cleaning this up
___
*[docker/Makefile, line 9 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-MUqlqnTEe3sYWe7j9IM:-MUqlqnU3ch05RWIcqRU:b-k808bu) ([raw file](https://github.com/sillsdev/web-languageforge/blob/11125dc891c260548a83f371e3479635b88a902f/docker/Makefile#L9)):*
> ```Makefile
> 
> .PHONY: e2e
> e2e: build
> ```

I guess I'm not a fan of having e2e depend on build.  `build` takes time on my machine, even with most layers cached and the optimizations we've done.  Ideally we should be able to volume map in the src and test folders into the appropriate places so that we can make changes to the tests / code and run tests without triggering a bigger build - would you agree?
___
*[docker/Makefile, line 42 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-MUqkzfF5IB1B8oJm8VS:-MUqkzfF5IB1B8oJm8VT:bh0fxee) ([raw file](https://github.com/sillsdev/web-languageforge/blob/11125dc891c260548a83f371e3479635b88a902f/docker/Makefile#L42)):*
> ```Makefile
> 	docker system prune -f --volumes
> 
> .PHONY: clean-powerwash
> ```

I like having the "power wash" option.  I will use that instead of `make clean` in the TC build step

longrunningprocess (commented):
Thank you, I'll go through those.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-:-MUsJGjb0EHexiJLls4H:b-9xtp92)* status: all files reviewed, 3 unresolved discussions (waiting on @megahirt and @rmunn)

___
*[README.md, line 105 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-MUqnB4aEGi2tVPRIueH:-MUsJKybDm3N2ZRS6LqJ:bnnz312) ([raw file](https://github.com/sillsdev/web-languageforge/blob/11125dc891c260548a83f371e3479635b88a902f/README.md#L105)):*

Thanks for adding instructions to the README!!!
</blockquote></details>

you bet!
___
*[docker/docker-compose.yml, line 105 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-MUqmUR37sk91PxjJfXg:-MUsJNiHDOxC4sQaHCc6:bpusuy6) ([raw file](https://github.com/sillsdev/web-languageforge/blob/11125dc891c260548a83f371e3479635b88a902f/docker/docker-compose.yml#L105)):*

I don't understand why `e2e-ui` depends on `ui`.  I would guess this would cause problems.  They are mutually exclusive containers, not designed to be run at the same time.
</blockquote></details>

my thinking is that even though the ui may have built the image already, the right artifacts from a UI build are not shared through the shared volume until the UI container is started, does that sound right to you?
___
*[docker/docker-compose.yml, line 158 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-MUqlAMt9H7vr3zo-7v-:-MUsJgJh27A2F5_42JXg:b-jjw9ir) ([raw file](https://github.com/sillsdev/web-languageforge/blob/11125dc891c260548a83f371e3479635b88a902f/docker/docker-compose.yml#L158)):*

why are we going back to kebab case?  Does it look better to you?  Docker uses _ for its volume names, but its a moot point since we don't actually need to identify the volumes directly.  If we're going to change back to kebab case (which I'm fine with), we might as well remove the "lf-" since that was only necessary to identify the LF docker volumes among others you (or the CI agent) have on their system.  The `docker-compose down -v` solution is much preferred
</blockquote></details>

:-D yeah, the tl; dr of it is, as a I put in the commit message for this one, I wanted to avoid confusion in the docker-compose file between docker-compose properties which strictly follow the snake_case.

I personally had not ever seen shared volumes with snake_case and your reasoning for going that way made sense to me at the time.  I was curious and investigated this a little further and found that every example I see from the community and the docs always, when separating the name, uses kebab-case (https://docs.docker.com/compose/compose-file/compose-file-v3/#volumes) as well as (https://docs.docker.com/storage/volumes/) and this is consistent with naming conventions for networks.

This is by no means something we need to have contention over, the simple answer here is I am just trying to follow convention but snake-case obviously works so if you'd prefer that style we can revert this.

Separately, the reason for the `lf_` is more of a namespacing intention.  In time you may find you have containers running from other projects or even volumes from other projects.  When you do something like `docker volume ls` or `docker ps` it's going to show you everything it knows of, regardless of which project you are currently in.  You see this today when you do a `docker images`.  That will show you all images it knows of, not just the ones related to project directory you're in.  Now, docker-compose makes its commands directly relevant to where you are because it relies on the `docker-compose.yml` file to do its thing.  The namespacing will make more sense at scale.
___
*[docker/Makefile, line 6 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-MUqn26-0-e9al6OFEik:-MUsUHFO6577uo8FcI0E:bnnz312) ([raw file](https://github.com/sillsdev/web-languageforge/blob/ce3c3708297a17c38ac7898732e54559b6c29b1b/docker/Makefile#L6)):*

thanks for cleaning this up
</blockquote></details>

you bet!
___
*[docker/Makefile, line 9 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-MUqlqnTEe3sYWe7j9IM:-MUsUJBH7x3mx0FVNugJ:b-u7tkql) ([raw file](https://github.com/sillsdev/web-languageforge/blob/11125dc891c260548a83f371e3479635b88a902f/docker/Makefile#L9)):*

I guess I'm not a fan of having e2e depend on build.  `build` takes time on my machine, even with most layers cached and the optimizations we've done.  Ideally we should be able to volume map in the src and test folders into the appropriate places so that we can make changes to the tests / code and run tests without triggering a bigger build - would you agree?
</blockquote></details>

I understand what you're saying about the delay...it's a cost associated with us having to give the `build_context: ..` because this means docker has to take the entire repo's filesystem and parse it for changes which is a bunch in the case of this repo.  We can certainly pull that dependency in the Makefile out but we do have to ensure the build has already run somehow so we can choose to put that in the instructions and hope the dev reads it or we could look at optimizing the build context a little and this step will be lightening fast.  We're just sort of brute-forcing it right now.
___
*[docker/Makefile, line 42 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-MUqkzfF5IB1B8oJm8VS:-MUsVIoT1B3bif3wabEq:byitz99) ([raw file](https://github.com/sillsdev/web-languageforge/blob/11125dc891c260548a83f371e3479635b88a902f/docker/Makefile#L42)):*

I like having the "power wash" option.  I will use that instead of `make clean` in the TC build step
</blockquote></details>

Excellent, I intended to circle back through TC to see how these changes would affect that process.  I wouldn't expect you would need a powerwash in a CI env though because I assume TC is creating a clean env for each run...is that not the case?

longrunningprocess (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/852)* status: all files reviewed, 3 unresolved discussions (waiting on @megahirt and @rmunn)

___
*[docker/docker-compose.yml, line 158 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-MUqlAMt9H7vr3zo-7v-:-MUsYmrq9D02eOH9rbzH:budrm5w) ([raw file](https://github.com/sillsdev/web-languageforge/blob/11125dc891c260548a83f371e3479635b88a902f/docker/docker-compose.yml#L158)):*

:-D yeah, the tl; dr of it is, as a I put in the commit message for this one, I wanted to avoid confusion in the docker-compose file between docker-compose properties which strictly follow the snake_case.

I personally had not ever seen shared volumes with snake_case and your reasoning for going that way made sense to me at the time.  I was curious and investigated this a little further and found that every example I see from the community and the docs always, when separating the name, uses kebab-case (https://docs.docker.com/compose/compose-file/compose-file-v3/#volumes) as well as (https://docs.docker.com/storage/volumes/) and this is consistent with naming conventions for networks.

This is by no means something we need to have contention over, the simple answer here is I am just trying to follow convention but snake-case obviously works so if you'd prefer that style we can revert this.

Separately, the reason for the `lf_` is more of a namespacing intention.  In time you may find you have containers running from other projects or even volumes from other projects.  When you do something like `docker volume ls` or `docker ps` it's going to show you everything it knows of, regardless of which project you are currently in.  You see this today when you do a `docker images`.  That will show you all images it knows of, not just the ones related to project directory you're in.  Now, docker-compose makes its commands directly relevant to where you are because it relies on the `docker-compose.yml` file to do its thing.  The namespacing will make more sense at scale.
</blockquote></details>

Additionally I feel the urge to clarify another point, I did not intend to undo your work because I didn't like it but I can see how that might've seemed that way.  I only reverted this because it was originally changed to make your `grep` work better but when we learned about the `down` we didn't actually need the `grep` approach anymore.  So my thinking was to revert this since the approach changed...as well as the convention thing.  If undoing the snake_case did come across the wrong way I'm really sorry about  that, it wasn't my intention if that helps.

rmunn (commented):
We also need to add something to the README about needing to download docker-compose from the GitHub releases if you're on Ubuntu Bionic or earlier, because the version in Apt is too old. See my detailed comment on one of the Makefile changes.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-:-MUv8AMV6fm5HsZtXzYo:bmlvbee)* status: all files reviewed, 4 unresolved discussions (waiting on @longrunningprocess and @megahirt)

___
*[docker/Makefile, line 34 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-MUv6614C7BQ0JU_5dav:-MUv6614C7BQ0JU_5daw:b-r6vu0) ([raw file](https://github.com/sillsdev/web-languageforge/blob/11125dc891c260548a83f371e3479635b88a902f/docker/Makefile#L34)):*
> ```Makefile
> .PHONY: clean
> clean:
> 	docker-compose down -t 1
> ```

Just so you're aware, this would be a breaking change (although it doesn't look like it) for anyone who installed docker-compose via "apt install docker-compose" on Bionic. The docker-compose version available in Ubuntu Bionic is 1.17.1, and the `-t` option was introduced in a docker-compose version later than 1.17.1. I'd rather avoid making people install docker-compose from GitHub releases if we don't *have* to make that change. However, here there's really little choice about the breaking change, because the reason why the shutdown is a little slow is, I believe, because we're not using the "init" option (see https://docs.docker.com/compose/compose-file/compose-file-v3/#init for details and https://blog.true-kubernetes.com/why-does-my-docker-container-take-10-seconds-to-stop/ for an explanation of why the "init" option fixes the issue).

Thing is, to have the "init" option available you need a docker-compose.yml file of version 3.7 or higher, and version 1.17.1 of the CLI tool only handles compose file format version 3.5 or lower.

So since either way of fixing the shutdown slowness requires people using Bionic to download a docker-compose version from GitHub instead of from Apt, I'm okay with this change. However, I think the README needs to be updated to mention the minimum docker-compose version requirements, and the fact that if you're on Bionic or earlier then the version in Apt is too old.

megahirt:
I can go either way on this.  Mandate a higher version of docker-compose
than what ships on an LTS, or go with lowest common denominator features to
simplify the developer prerequisites.  Thoughts?

megahirt (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/852)* status: all files reviewed, 4 unresolved discussions (waiting on @longrunningprocess)

___
*[docker/docker-compose.yml, line 105 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-MUqmUR37sk91PxjJfXg:-MUu01Dn2hWAUeLoML71:bo9s9yh) ([raw file](https://github.com/sillsdev/web-languageforge/blob/11125dc891c260548a83f371e3479635b88a902f/docker/docker-compose.yml#L105)):*

my thinking is that even though the ui may have built the image already, the right artifacts from a UI build are not shared through the shared volume until the UI container is started, does that sound right to you?
</blockquote></details>

ah I see where you're going.  In that case, shouldn't we just duplicate the developer volume mappings here instead?  Or does that feel anti-DRY to you?

```
      # for developer convenience
      - ../src:/data/src
      - ../typings:/data/typings
      - ../webpack.config.js:/data/webpack.config.js
      - ../gulpfile.js:/data/gulpfile.js
```
___
*[docker/docker-compose.yml, line 158 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-MUqlAMt9H7vr3zo-7v-:-MUu0ote1uopCFOhgyqK:b-3n5e85) ([raw file](https://github.com/sillsdev/web-languageforge/blob/11125dc891c260548a83f371e3479635b88a902f/docker/docker-compose.yml#L158)):*

Additionally I feel the urge to clarify another point, I did not intend to undo your work because I didn't like it but I can see how that might've seemed that way.  I only reverted this because it was originally changed to make your `grep` work better but when we learned about the `down` we didn't actually need the `grep` approach anymore.  So my thinking was to revert this since the approach changed...as well as the convention thing.  If undoing the snake_case did come across the wrong way I'm really sorry about  that, it wasn't my intention if that helps.
</blockquote></details>

Yes to namespacing !

Yes to kebab case conventions !  I wasn't convinced that docker standardized on _ , just what I observed when doing `docker volume ls`

Not offended at all !  Was just trying to understand.  Thanks for the reaching out.
___
*[docker/Makefile, line 9 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-MUqlqnTEe3sYWe7j9IM:-MUu1eml4qPnMC9pHHWV:b-xismle) ([raw file](https://github.com/sillsdev/web-languageforge/blob/11125dc891c260548a83f371e3479635b88a902f/docker/Makefile#L9)):*

I understand what you're saying about the delay...it's a cost associated with us having to give the `build_context: ..` because this means docker has to take the entire repo's filesystem and parse it for changes which is a bunch in the case of this repo.  We can certainly pull that dependency in the Makefile out but we do have to ensure the build has already run somehow so we can choose to put that in the instructions and hope the dev reads it or we could look at optimizing the build context a little and this step will be lightening fast.  We're just sort of brute-forcing it right now.
</blockquote></details>

My preference is to leave it to the dev at this point to `make build` when they see fit.  The meaning and necessity of `make build` is going to change over time.  As you know, it's a rough DX currently having to build in order to see src changes depending upon what you are trying to do.
___
*[docker/Makefile, line 42 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-MUqkzfF5IB1B8oJm8VS:-MUu24Sk2jZRPV8cK7nA:buvkwgp) ([raw file](https://github.com/sillsdev/web-languageforge/blob/11125dc891c260548a83f371e3479635b88a902f/docker/Makefile#L42)):*

Excellent, I intended to circle back through TC to see how these changes would affect that process.  I wouldn't expect you would need a powerwash in a CI env though because I assume TC is creating a clean env for each run...is that not the case?
</blockquote></details>

Well, doing CI with docker is brand new to me.  Currently the Docker build agent is simply an ubuntu bionic machine that has Docker installed.  This means that it's basically like any other machine in that you are responsible for cleaning up between runs.  It takes care of managing the repo checkout, but with regard to Docker volumes,  existing images, etc. it's up to you.  Crucial to the cleanup is deleting volumes after the build is finished (even if the build fails).  I posted what I have currently for the TC configuration in slack.

longrunningprocess (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/852)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @longrunningprocess, @megahirt, and @rmunn)

___
*[docker/docker-compose.yml, line 158 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-MUqlAMt9H7vr3zo-7v-:-MUxZWu7ErUzc8BvJsJk:b-rp24y5) ([raw file](https://github.com/sillsdev/web-languageforge/blob/11125dc891c260548a83f371e3479635b88a902f/docker/docker-compose.yml#L158)):*

Yes to namespacing !

Yes to kebab case conventions !  I wasn't convinced that docker standardized on _ , just what I observed when doing `docker volume ls`

Not offended at all !  Was just trying to understand.  Thanks for the reaching out.
</blockquote></details>

ok, great, sounds we're on the same page :-)
___
*[docker/Makefile, line 9 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-MUqlqnTEe3sYWe7j9IM:-MUx_7zi3Kj36DroMXzR:bhmjv0i) ([raw file](https://github.com/sillsdev/web-languageforge/blob/11125dc891c260548a83f371e3479635b88a902f/docker/Makefile#L9)):*

My preference is to leave it to the dev at this point to `make build` when they see fit.  The meaning and necessity of `make build` is going to change over time.  As you know, it's a rough DX currently having to build in order to see src changes depending upon what you are trying to do.
</blockquote></details>

you got it, committed that change.
___
*[docker/Makefile, line 34 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-MUv6614C7BQ0JU_5dav:-MUxgC7DE0aPpJSrdx-0:bb426a0) ([raw file](https://github.com/sillsdev/web-languageforge/blob/11125dc891c260548a83f371e3479635b88a902f/docker/Makefile#L34)):*

Just so you're aware, this would be a breaking change (although it doesn't look like it) for anyone who installed docker-compose via "apt install docker-compose" on Bionic. The docker-compose version available in Ubuntu Bionic is 1.17.1, and the `-t` option was introduced in a docker-compose version later than 1.17.1. I'd rather avoid making people install docker-compose from GitHub releases if we don't *have* to make that change. However, here there's really little choice about the breaking change, because the reason why the shutdown is a little slow is, I believe, because we're not using the "init" option (see https://docs.docker.com/compose/compose-file/compose-file-v3/#init for details and https://blog.true-kubernetes.com/why-does-my-docker-container-take-10-seconds-to-stop/ for an explanation of why the "init" option fixes the issue).

Thing is, to have the "init" option available you need a docker-compose.yml file of version 3.7 or higher, and version 1.17.1 of the CLI tool only handles compose file format version 3.5 or lower.

So since either way of fixing the shutdown slowness requires people using Bionic to download a docker-compose version from GitHub instead of from Apt, I'm okay with this change. However, I think the README needs to be updated to mention the minimum docker-compose version requirements, and the fact that if you're on Bionic or earlier then the version in Apt is too old.
</blockquote></details>

good idea, I've committed a change to the README that will address the additional installation work required for Linux users

As for the shutdown time, I thought it was directly related to the timeout which defaults to 10s..that's why I put the `-t 1` in place...it resolved the "slowness" when I did that.  I'm not familiar with the init option but if the timeout option is not working for you please let me know, maybe I'll need to learn more about init.

longrunningprocess (commented):
I simply made reference to Docker's doc on the matter, it will age the best over time.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-:-MUxhP2w1A8LdQbjqWdr:b-jc9rok)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @longrunningprocess, @megahirt, and @rmunn)

longrunningprocess (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/852)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @megahirt and @rmunn)

___
*[docker/docker-compose.yml, line 105 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-MUqmUR37sk91PxjJfXg:-MUxhbwn3J_qjUXU8uAl:bgtzfrj) ([raw file](https://github.com/sillsdev/web-languageforge/blob/11125dc891c260548a83f371e3479635b88a902f/docker/docker-compose.yml#L105)):*

ah I see where you're going.  In that case, shouldn't we just duplicate the developer volume mappings here instead?  Or does that feel anti-DRY to you?

```
      # for developer convenience
      - ../src:/data/src
      - ../typings:/data/typings
      - ../webpack.config.js:/data/webpack.config.js
      - ../gulpfile.js:/data/gulpfile.js
```
</blockquote></details>

sharing the right dirs might be a cleaner approach but a few things have to be in place, the directory the UI is filling with build artifacts need to be shared "out" of the container.  If that's the case then yes, dropping a volume on the `e2e-ui` container is a better way to go.  I'll need to research that to confirm.

As for the particular files you referenced, I would think it would only make sense to volume those in *if* the e2e-ui container is making use of them to re-build or something.  I'll need to refresh my memory on what's being done in `e2e-ui` either on build or on run.

megahirt (approved):
:lgtm: thanks for this contribution!

> Reviewed 1 of 15 files at r2, 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/852#-:-MV-2j8s6QAYrrkVFJJk:b-70b6du)* status: all files reviewed, 1 unresolved discussion (waiting on @rmunn)

megahirt (approved):
> Dismissed @rmunn from a discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/852)* status: all files reviewed, 1 unresolved discussion

megahirt (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/852)* status: :shipit: complete! all files reviewed, all discussions resolved

megahirt (approved):
> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/852)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #855, opened 2021-03-05
Fix project config E2E tests

rmunn:
The "Select All" tests were failing because the wrong Angular directive was being used for the "Select All" checkboxes in the field config. The ng-click directive fires before ng-model, so at the time the ng-click directive was firing on a "Select All" checkbox, the model was holding the *old* value for the checkbox, not the *new* value. Result: when you turned a "Select All" box *on*, it would turn all the checkboxes in its row and column *off*.

Replacing ng-click with ng-change on all the Select All checkboxes solved the issue. Since ng-change runs *after* ng-model, by the time the handler runs and updates the state of the checkboxes in the row and column, the model contains the *new* value and the "Select All" feature works like you'd expect it to work.

With this PR, all the project-configuration E2E tests pass, bringing the total number of failing E2E tests down to 17.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/855)

### DIFF
--- a/src/angular-app/languageforge/lexicon/settings/configuration/configuration-fields.component.html
+++ b/src/angular-app/languageforge/lexicon/settings/configuration/configuration-fields.component.html
@@ -31,28 +31,28 @@
                 <td class="text-center align-middle">
                     <input class="position-static" id="input-system-select-all-observer-checkbox"
                            data-ng-model="$ctrl.unifiedViewModel.inputSystems.selectAllColumns.observer"
-                           data-ng-click="$ctrl.selectAllRoleColumn($ctrl.unifiedViewModel.inputSystems.settings, $ctrl.unifiedViewModel.inputSystems.selectAllColumns, 'observer');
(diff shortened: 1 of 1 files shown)


rmunn:
Looks like the reason TeamCity is reporting 47 failures rather than the 17 I get when running locally is because on my local computer, I have the Docker images from #831 in my config. But since #831 isn't merged into master yet, a lot of Language Depot API calls are failing and popping up notices, hence why some tests that would otherwise pass are failing. Because those tests include a check for "expect(loginPage.infoMessages.count()).toBe(1)" and the infoMessages list includes an RPC error from the LD API call. So the test fails even though all code the test is checking for actually works.

Once #831 is merged in, I expect TeamCity to show the same 17 E2E failures that I'm seeing on my local dev box.

megahirt:
TC reports 46 failing tests on this PR - getting lower!  Thanks!

megahirt (approved):
:lgtm: great work Robin - this is super helpful!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/855#-:-MVEdLNM15Y8HjNxHuyU:bxesrhw)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #866, opened 2021-03-10
Fix one flaky E2E test

rmunn:
Race condition between dismissing modal and checking that the item was deleted; solved with `waitForAngular()` which will ensure that the item deletion completes before the next line of the E2E test proceeds.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/866)

### DIFF
--- a/test/app/languageforge/lexicon/editor/editor-entry.e2e-spec.ts
+++ b/test/app/languageforge/lexicon/editor/editor-entry.e2e-spec.ts
@@ -592,8 +592,10 @@ describe('Lexicon E2E Editor List and Entry', () => {
     await editorPage.browse.clickEntryByLexeme(constants.testEntry3.lexeme.th.value);
     await editorPage.edit.actionMenu.click();
     await editorPage.edit.deleteMenuItem.click();
+    await browser.waitForAngular();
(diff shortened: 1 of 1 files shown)


rmunn:
This approach (using `browser.waitForAngular()`) might be good to apply elsewhere in the code: anywhere where we have a `browser.sleep()` with an arbitrary time is probably a candidate for replacing with `waitForAngular()`, which will let that test run deterministically instead of guessing how many milliseconds to wait for.

longrunningprocess (approved):
seems fine to me for now...we'll go back through at some point and decide whether or not this is actually a code smell (maybe don't use modals ;-)) or a test smell (you should be hooking into angular's lifecycle rather just waiting to get lucky on timing)...either way, I think delaying on this would be letting perfect be the enemy of good, this is giving us reliable e2e test results and while that's not a statement of achieving green at all costs, we're not in a place for this project where we need to be caught up in dogma at this time.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/866#-:-MVSS11d0WTC-1lFrkN7:b-c38ue2)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rmunn)

PR #873, opened 2021-03-15
Remove Scripture Forge code and references

megahirt:
This is a large PR that isn't actually reviewable because of the number of files it touches.  This is my latest attempt at removing all traces of Scripture Forge and the DOA app "semdomtrans" from the codebase, with the expressed goal of removing unused code for readability and code comprehension by developers who don't have prior history in the repo.  Plus, it's always great to delete lines of code we no longer need :)  You will see that this PR is overwhelmingly deleting code.

284 files changed is probably not worth sifting through.  The important thing is that PHP and E2E tests pass.  This is a draft PR so that I can have TC run the E2E tests for me.  PHP tests should be passing.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/873)


rmunn (approved):
LGTM. The one failing E2E test is flaky and passed on a second run. I believe the flakiness is caused by hitting the real Language Depot API server during the test, and we'll be able to resolve it by setting up the E2E tests to use the ld-api container (using /api/v2 endpoints) for doing Send/Receive testing. (This will require a bit of work, so it's suitable for a different PR).

I think this should come out of draft status and be merged. :rocket: 

> Reviewed 286 of 286 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/873#-:-MVpFziH5I9UMT1uzg5e:b4bamf6)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @megahirt)

rmunn:
@megahirt I've reviewed & approved this PR. If you want to mark it as "ready for review", you should be able to then merge it immediately. The flaky E2E test passed on the second round, and I have an idea for fixing the lexicon-new-project.e2e-spec.ts tests by having them use the new ld-api container (right now I *think* they're going off and hitting the real admin.languagedepot.org endpoint, which is why they're slow and sometimes timeout before getitng a response). I'll discuss that with you at the standup meeting tomorrow.

megahirt:
All green

image

PR #885, opened 2021-03-19
Move webpack and sass to npm scripts (remove from gulp)

megahirt:
This removes all references to sass and webpack from the gulpfile

In order to accomplish parallel sass and webpack watches, I am using the npm-run-all dev dependency.  The new way to run sass and webpack are:
```
npm run sass
npm run sass:watch
npm run webpack
npm run webpack:watch
npm run webpack:production
```

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/885)

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,6 +1,5 @@
 'use strict';
 
-var BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
 var CopyWebpackPlugin = require('copy-webpack-plugin');
 var path = require('path');
 var LiveReloadPlugin = require('webpack-livereload-plugin');
@@ -13,9 +12,6 @@ var workboxPlugin = require('workbox-webpack-plugin');
 module.exports = function (env) {
   if (env == null) {
     env = {
-      applicationName: 'languageforge',
-      isProduction: false,
-      isAnalyze: false,
       isTest: false
     };
   }
@@ -114,17 +110,17 @@ module.exports = function (env) {
   var serviceWorkerConfig = {
     clientsClaim: true,
     importScripts: [
-      '/service-worker/' + env.applicationName + '/service-worker.js'
+      '/service-worker/languageforge/service-worker.js'
     ],
     include: [] // To be changed once ready to start caching the whole app
   };
 
   // Set up and return a customized Webpack config according to the environment we're in
 
-  webpackConfig.entry.main = './src/angular-app/' + env.applicationName + '/main' +
+  webpackConfig.entry.main = './src/angular-app/languageforge/main' +
     (env.isTest ? '.specs' : '') + '.ts';
 
-  if (env.isProduction) {
+  if (env.NODE_ENV == "production") {
     webpackConfig.plugins.push(new webpack.DefinePlugin({
       'process.env.NODE_ENV': JSON.stringify('production')
     }));
@@ -138,10 +134,6 @@ module.exports = function (env) {
     serviceWorkerConfig.importScripts = ['/service-worker/service-worker-debug.js']
       .concat(serviceWorkerConfig.importScripts);
     serviceWorkerConfig.skipWaiting = true;
-  }
-
-  if (env.isAnalyze) {
-    webpackConfig.plugins.push(new BundleAnalyzerPlugin({ analyzerMode: 'static' }));
   }
 
   if (env.isTest) {
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -120,7 +120,7 @@ module.exports = function (env) {
   webpackConfig.entry.main = './src/angular-app/languageforge/main' +
     (env.isTest ? '.specs' : '') + '.ts';
 
-  if (env.NODE_ENV == "production") {
+  if (process.env.NODE_ENV == "production") {
     webpackConfig.plugins.push(new webpack.DefinePlugin({
       'process.env.NODE_ENV': JSON.stringify('production')
     }));
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ vagrant_ansible_inventory_*
 test/app/protractorConfBrowserStack.js
 artifacts/
 src/userConfig.php
+src/dist
 
 secrets/
 
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -7,10 +7,10 @@
         "label": "webpack",
         "type": "shell",
         "windows": {
-          "command": "bash -c 'gulp webpack-lf'"
+          "command": "bash -c 'npm run webpack'"
         },
         "linux": {
-          "command": "gulp webpack-lf"
+          "command": "npm run webpack"
         },
         "problemMatcher": "$awesometsc",
         "group": {
@@ -26,10 +26,10 @@
         "type": "shell",
         "isBackground": true,
         "windows": {
-          "command": "bash -c 'gulp webpack-lf:watch'"
+          "command": "bash -c 'npm run webpack:watch'"
         },
         "linux": {
-          "command": "gulp webpack-lf:watch"
+          "command": "npm run webpack:watch"
         },
         "presentation": {
           "reveal": "silent"
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -23,7 +23,7 @@ build-test-php: wait
 
 .PHONY: build
 build: wait
-	docker-compose build
+	docker-compose build mail ui-builder app
 
 .PHONY: wait
 wait:
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       dockerfile: docker/ui-builder/Dockerfile
     image: lf-ui-builder
     container_name: lf-ui-builder
-    command: npm run watch-all
+    command: npm run build:watch
     volumes:
       # for developer convenience
       - ../src:/data/src
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       dockerfile: docker/ui-builder/Dockerfile
     image: lf-ui-builder
     container_name: lf-ui-builder
-    command: gulp sass:watch webpack-lf:watch
+    command: npx npm-run-all -p -r sass:watch webpack:watch
     volumes:
       # for developer convenience
       - ../src:/data/src
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -27,7 +27,6 @@ services:
       - db
       - mail
       - ld-api
-      - ui-builder
     environment:
       - WAIT_HOSTS=db:27017, mail:25
       - "LDAPI_BASE_URL=http://ld-api:3000/api/v2/"
@@ -49,12 +48,12 @@ services:
       - 80:80
       - 443:443
     depends_on:
-      - ui-builder
+      - app
     volumes:
       - lf-caddy-data:/data
       - lf-caddy-config:/config
     # https://caddyserver.com/docs/command-line
-    command: caddy reverse-proxy --from localhost --to api
+    command: caddy reverse-proxy --from localhost --to app
     restart: unless-stopped
 
   mail:
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       dockerfile: docker/ui-builder/Dockerfile
     image: lf-ui-builder
     container_name: lf-ui-builder
-    command: npx npm-run-all -p -r sass:watch webpack:watch
+    command: npm run watch-all
     volumes:
       # for developer convenience
       - ../src:/data/src
--- a/docker/ui-builder/Dockerfile
+++ b/docker/ui-builder/Dockerfile
@@ -26,4 +26,4 @@ COPY gulpfile.js webpack.config.js tsconfig.json tslint.json ./
 COPY src ./src/
 
 # artifacts built to /data/src/dist
-RUN gulp sass webpack-lf
+RUN npm run sass && npm run webpack
--- a/docker/ui-builder/Dockerfile
+++ b/docker/ui-builder/Dockerfile
@@ -26,4 +26,4 @@ COPY gulpfile.js webpack.config.js tsconfig.json tslint.json ./
 COPY src ./src/
 
 # artifacts built to /data/src/dist
-RUN npm run sass && npm run webpack
+RUN npm run build
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -27,8 +27,6 @@ var _template = require('lodash.template');
 var Server = require('karma').Server;
 var path = require('path');
 var del = require('del');
-var webpack = require('webpack');
-var webpackConfig = require('./webpack.config.js');
 
 // -------------------------------------
 //   Global Variables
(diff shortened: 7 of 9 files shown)


rmunn (changes requested):
One issue, the rest looks good. I think one case of `env.NODE_ENV` needs to be `process.env.NODE_ENV`.

> Reviewed 7 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/885#-:-MW8aIFp9aDKCu2oKwtU:b-o8turw)* status: all files reviewed, 2 unresolved discussions (waiting on @longrunningprocess and @megahirt)

___
*[webpack.config.js, line 13 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/885#-MW8_DSp1PbNs9T8iM9Y:-MW8_DSp1PbNs9T8iM9Z:bjfo2d8) ([raw file](https://github.com/sillsdev/web-languageforge/blob/578781be7d7a7ecef53703009eca0058df737590/webpack.config.js#L13)):*
> ```JavaScript
> 
> module.exports = function (env) {
>   if (env == null) {
> ```

Is this `if (env == null)` check even useful any more? If it's possible that we could get passed a null `env` value, then it would be useful... but if that's the case then the `if (env.NODE_ENV == 'production')` check would fail later on as there is no `NODE_ENV` value being set here in the default.
___
*[webpack.config.js, line 123 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/885#-MW8_RGG6Z6D5eUUYB9-:-MW8_RGG6Z6D5eUUYB90:b6wvqt2) ([raw file](https://github.com/sillsdev/web-languageforge/blob/578781be7d7a7ecef53703009eca0058df737590/webpack.config.js#L123)):*
> ```JavaScript
>     (env.isTest ? '.specs' : '') + '.ts';
> 
>   if (env.NODE_ENV == "production") {
> ```

Should this be `process.env.NODE_ENV`? Or does Webpack pass in an environment that's guaranteed to include this? From https://webpack.js.org/api/cli/#environment-options it seems like the `env` that we'd receive does *not* have environment variables in it, and this should be `process.env.NODE_ENV`.

megahirt (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/885)* status: all files reviewed, 2 unresolved discussions (waiting on @longrunningprocess and @rmunn)

___
*[webpack.config.js, line 13 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/885#-MW8_DSp1PbNs9T8iM9Y:-MW8ba0qBoo4kOWPCPWm:b-gkiu32) ([raw file](https://github.com/sillsdev/web-languageforge/blob/578781be7d7a7ecef53703009eca0058df737590/webpack.config.js#L13)):*

Is this `if (env == null)` check even useful any more? If it's possible that we could get passed a null `env` value, then it would be useful... but if that's the case then the `if (env.NODE_ENV == 'production')` check would fail later on as there is no `NODE_ENV` value being set here in the default.
</blockquote></details>

per my comment below, I think we still need the if statement.  I'm actually uncertain if we ever use the `test` variable, but there are statements in the webpack config to switch functionality for it.  I'm wondering if those are related to when we had .net core and other karma tests.  We may not be using this part of the config anymore.
___
*[webpack.config.js, line 123 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/885#-MW8_RGG6Z6D5eUUYB9-:-MW8bWo1AH3F7Mi90kch:b-kmup12) ([raw file](https://github.com/sillsdev/web-languageforge/blob/578781be7d7a7ecef53703009eca0058df737590/webpack.config.js#L123)):*

Should this be `process.env.NODE_ENV`? Or does Webpack pass in an environment that's guaranteed to include this? From https://webpack.js.org/api/cli/#environment-options it seems like the `env` that we'd receive does *not* have environment variables in it, and this should be `process.env.NODE_ENV`.
</blockquote></details>

yes, it should be `process.env.NODE_ENV`  good catch!

megahirt (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/885)* status: all files reviewed, 3 unresolved discussions (waiting on @longrunningprocess and @rmunn)

___
*[docker/docker-compose.yml, line 10 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/885#-MW8e4ahDky1KcKnTumT:-MW8e4ahDky1KcKnTumU:bteiykj) ([raw file](https://github.com/sillsdev/web-languageforge/blob/578781be7d7a7ecef53703009eca0058df737590/docker/docker-compose.yml#L10)):*
> ```YAML
>     image: lf-ui-builder
>     container_name: lf-ui-builder
>     command: npx npm-run-all -p -r sass:watch webpack:watch
> ```

turns out this line is failing.  I will need more time to investigate.  Since this is a developer-only command it's not as big of a deal, but would like to figure it out just the same.  Errors:

webpack.config.js
```diff
@@ -114,17 +110,17 @@ module.exports = function (env) {
...
   };
 
   // Set up and return a customized Webpack config according to the environment we're in
 
-  webpackConfig.entry.main = './src/angular-app/' + env.applicationName + '/main' +
+  webpackConfig.entry.main = './src/angular-app/languageforge/main' +
     (env.isTest ? '.specs' : '') + '.ts';
 
-  if (env.isProduction) {
+  if (env.NODE_ENV == "production") {
```
megahirt, line 123:
```suggestion
  if (process.env.NODE_ENV == "production") {
```

megahirt (commented):
Suggesting changes per @rmunn review

webpack.config.js
rmunn: Looks good.

rmunn (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/885)* status: all files reviewed, 2 unresolved discussions (waiting on @longrunningprocess and @megahirt)

___
*[webpack.config.js, line 123 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/885#-MW8_RGG6Z6D5eUUYB9-:-MW8lumvDelyMOdz5Y7e:b-l12iq8) ([raw file](https://github.com/sillsdev/web-languageforge/blob/578781be7d7a7ecef53703009eca0058df737590/webpack.config.js#L123)):*

yes, it should be `process.env.NODE_ENV`  good catch!
</blockquote></details>

Once this change is made I'll be happy with it; approving the review now so that you won't have to wait for me once the change gets pushed through.
___
*[docker/docker-compose.yml, line 10 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/885#-MW8e4ahDky1KcKnTumT:-MW8m509670HYseBhOKy:b-p2ezhf) ([raw file](https://github.com/sillsdev/web-languageforge/blob/578781be7d7a7ecef53703009eca0058df737590/docker/docker-compose.yml#L10)):*

turns out this line is failing.  I will need more time to investigate.  Since this is a developer-only command it's not as big of a deal, but would like to figure it out just the same.  Errors:



</blockquote></details>

I managed to get something similar, and I wonder if we're using npx wrong? If I do `node node_modules/npm-run-all/bin/npm-run-all/index.js sass:watch webpack:watch` it seems to work.

rmunn (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/885)* status: all files reviewed, 2 unresolved discussions (waiting on @longrunningprocess and @megahirt)

webpack.config.js
rmunn: I'll let you commit this since if I commit it, you'll end up having to rebase your branch next time you try to push it to GitHub and that would not be worth the complexity of the rebase for such a simple change.

megahirt (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/885)* status: 4 of 8 files reviewed, 1 unresolved discussion (waiting on @longrunningprocess, @megahirt, and @rmunn)

___
*[docker/docker-compose.yml, line 10 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/885#-MW8e4ahDky1KcKnTumT:-MW91yxK6u6I_T6M6JXm:bf7y076) ([raw file](https://github.com/sillsdev/web-languageforge/blob/578781be7d7a7ecef53703009eca0058df737590/docker/docker-compose.yml#L10)):*

I managed to get something similar, and I wonder if we're using npx wrong? If I do `node node_modules/npm-run-all/bin/npm-run-all/index.js sass:watch webpack:watch` it seems to work.
</blockquote></details>

I got it working by putting the `npm-run-all` inside a different script called `watch-all`, so I guess calling it via `npx npm-run-all` isn't the correct way to do it.  At least I was getting unexpected behavior that way.

megahirt (commented):
> Reviewed 4 of 7 files at r1, 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/885)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @longrunningprocess)

longrunningprocess (changes requested):
> Reviewed 1 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/885)* status: all files reviewed, 4 unresolved discussions (waiting on @megahirt)

___
*[gulpfile.js, line 58 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/885#-MWAwB8a7G2ffIWKYLVn:-MWAwB8b8X8RaG3aWBGk:b-itkfe) ([raw file](https://github.com/sillsdev/web-languageforge/blob/cb55a641faa3b7a64a4be6e71405d7bf73e8c759/gulpfile.js#L58)):*
> ```JavaScript
>   'test/**/*.php'
> ];
> ```

wow, I've never loved red more!
___
*[package.json, line 12 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/885#-MWAws4XAke6Znu9241g:-MWAws4Y64X3iFN63XIF:b-um09lv) ([raw file](https://github.com/sillsdev/web-languageforge/blob/88f2c4c424f9dfbe70d8eb6d1d5993b1a01122de/package.json#L12)):*
> ```JSON
>     "webpack:watch": "webpack -w",
>     "webpack:production": "webpack -p",
>     "watch-all": "npm-run-all -r -p sass:watch webpack:watch"
> ```

when you run these in parallel, I've found it useful to have the little labels next to the output lines so it's easy to know which script is responsible. You do that with an additional flag `-l`.
___
*[package.json, line 12 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/885#-MWAxPC37CRjWULywRRz:-MWAxPC37CRjWULywRS-:bephmnj) ([raw file](https://github.com/sillsdev/web-languageforge/blob/88f2c4c424f9dfbe70d8eb6d1d5993b1a01122de/package.json#L12)):*
> ```JSON
>     "webpack:watch": "webpack -w",
>     "webpack:production": "webpack -p",
>     "watch-all": "npm-run-all -r -p sass:watch webpack:watch"
> ```

Is the intention behind the race flag if one of these processes fails due to a compile problem?  I've never the seen the us of the `-r` and the docs are a little confusing to me, sounds like if the exit with 0, all processes will be stopped...
___
*[webpack.config.js, line 113 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/885#-MWAxoCK9at1ECTwXlk4:-MWAxoCK9at1ECTwXlk5:bnq95dc) ([raw file](https://github.com/sillsdev/web-languageforge/blob/88f2c4c424f9dfbe70d8eb6d1d5993b1a01122de/webpack.config.js#L113)):*
> ```JavaScript
>     clientsClaim: true,
>     importScripts: [
>       '/service-worker/languageforge/service-worker.js'
> ```

excellent!
___
*[webpack.config.js, line 120 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/885#-MWAxtev1MgHnddd4MOt:-MWAxtev1MgHnddd4MOu:bnq95dc) ([raw file](https://github.com/sillsdev/web-languageforge/blob/88f2c4c424f9dfbe70d8eb6d1d5993b1a01122de/webpack.config.js#L120)):*
> ```JavaScript
>   // Set up and return a customized Webpack config according to the environment we're in
> 
>   webpackConfig.entry.main = './src/angular-app/languageforge/main' +
> ```

excellent!
___
*[webpack.config.js, line 141 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/885#-MWAxxPp3O43senXGPdA:-MWAxxPp3O43senXGPdB:b-3ebkyv) ([raw file](https://github.com/sillsdev/web-languageforge/blob/cb55a641faa3b7a64a4be6e71405d7bf73e8c759/webpack.config.js#L141)):*
> ```JavaScript
>       .concat(serviceWorkerConfig.importScripts);
>     serviceWorkerConfig.skipWaiting = true;
>   }
> ```

yes please!
___
*[docker/docker-compose.yml, line 10 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/885#-MW8e4ahDky1KcKnTumT:-MWAyG1vDeh9AL-dhvsc:b-swyyq7) ([raw file](https://github.com/sillsdev/web-languageforge/blob/578781be7d7a7ecef53703009eca0058df737590/docker/docker-compose.yml#L10)):*

I got it working by putting the `npm-run-all` inside a different script called `watch-all`, so I guess calling it via `npx npm-run-all` isn't the correct way to do it.  At least I was getting unexpected behavior that way.
</blockquote></details>

correct, you typically see `npx` used on the command line if you're doing something utility in nature, more ephemeral...whereas the npm-run-all is part of our dev-deps, we need all "stuff"
___
*[docker/ui-builder/Dockerfile, line 29 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/885#-MWB-GsT2VWR8o4pVWes:-MWB-GsUBHZxdKnQ3-XN:b-i7vz2s) ([raw file](https://github.com/sillsdev/web-languageforge/blob/88f2c4c424f9dfbe70d8eb6d1d5993b1a01122de/docker/ui-builder/Dockerfile#L29)):*
> ```Dockerfile
> 
> # artifacts built to /data/src/dist
> RUN npm run sass && npm run webpack
> ```

just curios why you did `&&` here?

if there's a sequential relationship here, it would be better to hide that away in the package.json where the coupling is known and controlled...that is sort of "build intelligence"

consumers of your package.json should just need an `npm build` option here ideally.

longrunningprocess (commented):
I love this effort, excellent work! :trophy: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/885#-:-MWB-r5r6hBij01B3Q2s:ba6quti)* status: all files reviewed, 4 unresolved discussions (waiting on @megahirt)

megahirt (commented):
Thanks Billy!

> Reviewed 2 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/885#-:-MWKJHjD9h1No5J5fRXk:b67kgv5)* status: 7 of 8 files reviewed, 2 unresolved discussions (waiting on @longrunningprocess and @megahirt)

___
*[package.json, line 12 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/885#-MWAxPC37CRjWULywRRz:-MWKJNpn6RlZv1hOENt_:b-hu3rpo) ([raw file](https://github.com/sillsdev/web-languageforge/blob/88f2c4c424f9dfbe70d8eb6d1d5993b1a01122de/package.json#L12)):*

Is the intention behind the race flag if one of these processes fails due to a compile problem?  I've never the seen the us of the `-r` and the docs are a little confusing to me, sounds like if the exit with 0, all processes will be stopped...
</blockquote></details>

My understanding of the -r is that it means stop both if one fails.  This seems reasonable and desirable to me.  The SO answer I found on npm-run-all had a `-r`
___
*[package.json, line 12 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/885#-MWAws4XAke6Znu9241g:-MWKMh7N7xG8NwBZuCQP:bvg6hmn) ([raw file](https://github.com/sillsdev/web-languageforge/blob/88f2c4c424f9dfbe70d8eb6d1d5993b1a01122de/package.json#L12)):*

when you run these in parallel, I've found it useful to have the little labels next to the output lines so it's easy to know which script is responsible. You do that with an additional flag `-l`.
</blockquote></details>

done.  Added `-l`.  Thanks for that.
___
*[docker/ui-builder/Dockerfile, line 29 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/885#-MWB-GsT2VWR8o4pVWes:-MWKJtQi1cYpW3D8Y0Yk:butgd1s) ([raw file](https://github.com/sillsdev/web-languageforge/blob/88f2c4c424f9dfbe70d8eb6d1d5993b1a01122de/docker/ui-builder/Dockerfile#L29)):*

just curios why you did `&&` here?

if there's a sequential relationship here, it would be better to hide that away in the package.json where the coupling is known and controlled...that is sort of "build intelligence"

consumers of your package.json should just need an `npm build` option here ideally.
</blockquote></details>

an `npm run build` script to do both of these sounds good to me.  I've added that now.

megahirt (commented):
> Reviewed 1 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/885)* status: all files reviewed, 2 unresolved discussions (waiting on @longrunningprocess)

longrunningprocess (approved):
you bet, glad to offer some useful observations

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/885#-:-MWOzHKSClTFuJN_ICl_:b-yaophn)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @megahirt)

___
*[package.json, line 12 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/885#-MWAxPC37CRjWULywRRz:-MWOz1ZE3udLjnstao1g:b-l6591z) ([raw file](https://github.com/sillsdev/web-languageforge/blob/88f2c4c424f9dfbe70d8eb6d1d5993b1a01122de/package.json#L12)):*

My understanding of the -r is that it means stop both if one fails.  This seems reasonable and desirable to me.  The SO answer I found on npm-run-all had a `-r`
</blockquote></details>

that is what I'd expect the intention is, the docs just confused me.

longrunningprocess (approved):
> Reviewed 3 of 7 files at r1, 1 of 4 files at r2, 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/885)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @megahirt)

longrunningprocess (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/885#-:-MWOzO-yE06LbxUjGx7Z:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @megahirt)

PR #891, opened 2021-03-25
move protractor e2e test command to package.json

megahirt:
This removes the protractor command (and related tsc compile command) from the gulpfile.  

This adds two new npm scripts:
- `compile-test-e2e`
- `test-e2e`

`compile-test-e2e` runs during image build and `test-e2e` runs as the container command.

I also took the opportunity to remove comments in the protractorConf.js file.

TODO: make custom specs  work by passing in arguments to npm run test-e2e (still figuring that out)

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/891)

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -1,7 +1,7 @@
 # https://docs.docker.com/compose/reference/overview
 
 .PHONY: start
-start: wait
+start:
   # starts the entire runtime infrastructure
 	docker-compose up -d ssl
 
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -121,9 +121,12 @@ services:
     volumes:
       # for developer convenience
       - ../src:/data/src
+      - ../test:/data/test
       - ../typings:/data/typings
       - ../webpack.config.js:/data/webpack.config.js
       - ../gulpfile.js:/data/gulpfile.js
+      - ../package.json:/data/package.json
+      - ../package-lock.json:/data/package-lock.json
 
   app-for-e2e:
     build:
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -121,12 +121,14 @@ services:
     volumes:
       # for developer convenience
       - ../src:/data/src
-      - ../test:/data/test
       - ../typings:/data/typings
       - ../webpack.config.js:/data/webpack.config.js
       - ../gulpfile.js:/data/gulpfile.js
       - ../package.json:/data/package.json
       - ../package-lock.json:/data/package-lock.json
+# uncomment on dev machine for convenience but don't commit - it messes up the e2e test run by
+# clobbering the JS files produced by tsc in the built image.
+#      - ../test:/data/test
 
   app-for-e2e:
     build:
--- a/docker/test-e2e/Dockerfile
+++ b/docker/test-e2e/Dockerfile
@@ -8,6 +8,6 @@ COPY docker/test-e2e/run.sh /run.sh
 # copy in test folder
 COPY test/ /data/test/
 
-RUN gulp test-e2e-clean-compile
+RUN npm run compile-test-e2e
 
 CMD ["/run.sh"]
--- a/docker/test-e2e/run.sh
+++ b/docker/test-e2e/run.sh
@@ -17,7 +17,7 @@ attach_debugger () {
 
 attach_debugger &
 cd /data
-gulp test-e2e-doTest --webserverHost app-for-e2e "$@"
+npm run test-e2e
 STATUS=$?
 
 exit $STATUS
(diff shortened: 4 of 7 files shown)


megahirt:
E2E build failing due to `pattern /data/test/app/allspecs/**/*.e2e-spec.js did not match any files.`  I suspect that the TSC compilation step isn't running properly, but still need to investigate.  Of course, it works on my dev machine but it's probably due to some artifacts on my machine.

megahirt:
> E2E build failing due to `pattern /data/test/app/allspecs/**/*.e2e-spec.js did not match any files.` I suspect that the TSC compilation step isn't running properly, but still need to investigate. Of course, it works on my dev machine but it's probably due to some artifacts on my machine.

OK after a `git clean -f -x -d` I can reproduce this locally so will be pushing up a fix soon.

rmunn:
> TODO: make custom specs work by passing in arguments to npm run test-e2e (still figuring that out)

To pass arguments to `npm run`, you use the `--` separator, and anything after the `--` will be passed as additional arguments to the npm script you're running. So if the `test-e2e` script runs `protractor test/app/protractorConf.js`, then doing `npm run test-e2e -- --specs some-spec-name` will run `protractor test/app/protractorConf.js --specs some-spec-name`.

longrunningprocess (changes requested):
well, I love all the removals, excellent!!!

`wait` must be built prior to many of the other targets, please let me know what the impetus was for removing it, maybe we need to look at changing something else.

> Reviewed 6 of 7 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/891#-:-MWf-96S7bZ3e0o8G0uJ:b-hmoquo)* status: all files reviewed, 1 unresolved discussion (waiting on @megahirt and @rmunn)

___
*[docker/Makefile, line 4 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/891#-MWezt6E16n4XuPFno4w:-MWezt6E16n4XuPFno4x:b-8iram1) ([raw file](https://github.com/sillsdev/web-languageforge/blob/af0357fa50ff338a87c81d3fd339105ead614c0e/docker/Makefile#L4)):*
> ```Makefile
> 
> .PHONY: start
> start: wait
> ```

I don't think you want to remove this...many images depend on `wait` to be there already.

megahirt (commented):
Yes, `wait` is a pre-requisite for many containers.  So here's my thinking on this.  I removed it from the :start target because building happens down in a different target "build", which also depends upon `wait`.  The assumption is that `build` and `start` are two distinct processes that happen sequentially, invoked by the developer per the readme.  Since we already build `wait` in the build step, we don't need to build it again in the `start` step.  I realize that docker-compose builds services as they need them, however our use of docker-compose goes outside of it's intended design by including services that perform specific functions outside of application run (e.g. testing and db-reset).  If our application architecture was simple and more cookie cutter, then we could let docker-compose do the building on demand, get rid of the build make target entirely, and have only `wait` be a dependent target on `make start`.  We could shift some of the complexity to multiple docker-compose files but I don't think that buys us a lot at this point, given that we have existing architecture constraints we have to deal with which are not as amenable to dockerized application architecture.

That's a long explanation for why I deleted four characters :)  Basically didn't see the need to duplicate the wait build, since it's done during the build step as well.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/891#-:-MWgF4MJE8GD93IDuWFS:b-epkif1)* status: all files reviewed, 1 unresolved discussion (waiting on @longrunningprocess and @rmunn)

___
*[docker/Makefile, line 4 at r2](https://reviewable.io/reviews/sillsdev/web-languageforge/891#-MWezt6E16n4XuPFno4w:-MWgVVfHCCOkTUHVQCoY:b-ri0sgc) ([raw file](https://github.com/sillsdev/web-languageforge/blob/af0357fa50ff338a87c81d3fd339105ead614c0e/docker/Makefile#L4)):*

I don't think you want to remove this...many images depend on `wait` to be there already.
</blockquote></details>

Thanks.  See my reply to your earlier comment.

megahirt (commented):
> Dismissed @longrunningprocess from a discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/891)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rmunn)

megahirt:
So I'm going to move this PR forward and merge it.  @longrunningprocess we can have more discussion about the wait being removed from the make target dependency if you'd like after reading my thinking on that.  I'm actually inclined to move the wait image out to docker hub so we don't have to build it locally at all anymore.  Totally static utility image - probably best to move that out to the cloud.  I will create a separate PR for that.

PR #894, opened 2021-03-26
remove karma and filters.spec.ts test

megahirt:
Reposting this message from Slack (here is a better long-term place for it)

List on why I think we should cut our losses with Karma unit testing in Docker:

- I can only find one spec.ts file that contains any tests: [filters.spec.ts](https://github.com/sillsdev/web-languageforge/blob/master/src/angular-app/bellows/core/filters.spec.ts)
- I seem to remember there being more tests, but I cannot find them.  Maybe @irahopkinson  can help?
- The setup was sufficiently complex at one point that only @ddaspit  could get it working again.
- Removing Karma unit testing with significantly simplify our webpack config, as that is used for both karma and app typescript compilation and bundling
- The way it's written, it won't work out of the box in Docker.  Karma attaches to an available browser like Chrome, so we'd have to bring in another "karma chrome container" in order to make this work I'm guessing, and that's more bloat to the infrastructure.  If we only have a few tests, then I don't think it's worth it.
- Karma is old tech and was designed and built by the angularjs team, which is also old tech.  Same deal with protractor, but we have a lot of e2e tests that are adding value.
- I don't understand karma and webpack well enough that I suspect it would take me a number of full-time days to get it working again.

I did a sanity check and searched for any file with the name *spec* in it:
image
Only the filters.spec.ts came up.   The other files are e2e test related or test spec entry/aggregator files.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/894)


irahopkinson (approved):
:lgtm:

> Reviewed 8 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/894#-:-MWh7pV18dd7GDTCJUfe:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @megahirt)

megahirt (commented):
> Reviewed 5 of 8 files at r1, 2 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/894)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @megahirt)

PR #900, opened 2021-04-02
Github Action for Tests

megahirt:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/900)


megahirt:
This was a good experiment in getting Unit tests and E2E tests running on GA.  In the end it doesn't seem advantageous enough to switch away from a working system (TC) when the GA isn't providing us with performance benefits like I had perhaps hoped.

PR #901, opened 2021-04-05
Update search endpoints in LD API

rmunn:
Commit 607e5d51 in the LD API has changed the search endpoints from`/api/v2/searchUsers` to `/api/v2/search/users` (and similar for projects). This commit updates the API endpoints that Language Forge calls to match these new locations.

Since the ld-api container isn't part of the default "make build" list, you'll need to run `docker-compose build ld-api` after this PR is merged into master, or you'll see some error messages in the Language Forge UI as LF tries to access the new endpoints while your local ld-api container hasn't yet been rebuilt and is thus still serving up the old endpoints.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/901)

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -34,7 +34,7 @@ endif
 
 .PHONY: build
 build:
-	docker-compose build mail ui-builder app
+	docker-compose build mail ui-builder app ld-api
 
 .PHONY: db-reset
 db-reset:
--- a/docker/ld-api/Dockerfile
+++ b/docker/ld-api/Dockerfile
@@ -8,20 +8,24 @@
 
 # Step 1: Build app
 FROM node:14.15.4-alpine AS builder
+RUN npm i -g pnpm
 WORKDIR /git
 ADD https://github.com/sillsdev/web-languagedepot-api/tarball/nodejs ./
 # Github tarball has top-level directory USER-REPO-HASH, and HASH is unpredictable ahead of time, so we rename it
 RUN tar xvzf nodejs && rm nodejs && mv sillsdev-web-languagedepot-api-* ldapi
 WORKDIR /git/ldapi
-RUN npm ci
-RUN npm run build && npm run adapt
+RUN pnpm i --frozen-lockfile
+RUN pnpm run build
 
 # Step 2: Create minimal container with running app
 FROM node:14.15.4-alpine
+COPY --from=builder /usr/local/lib/node_modules/pnpm /usr/local/lib/node_modules/pnpm
+RUN ln -s ../lib/node_modules/pnpm/bin/pnpm.js /usr/local/bin/pnpm && \
+    ln -s ../lib/node_modules/pnpm/bin/pnpx.js /usr/local/bin/pnpx
 WORKDIR /app
 COPY --from=builder /git/ldapi/static assets
-COPY --from=builder /git/ldapi/package*.json ./
-RUN npm ci
+COPY --from=builder /git/ldapi/package.json /git/ldapi/pnpm-lock.yaml ./
+RUN pnpm i --prod --frozen-lockfile
 COPY --from=builder /git/ldapi/build ./
 EXPOSE 3000
 USER node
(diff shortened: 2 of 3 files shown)


megahirt (approved):
Also worth mentioning the addition of pnpm (performant npm) which replaces npm in the ld-api Dockerfile.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/901#-:-MXW9Ho6AyeLU-7v2IY2:b-8x231t)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rmunn)

___
*[docker/ld-api/Dockerfile, line 11 at r1](https://reviewable.io/reviews/sillsdev/web-languageforge/901#-MXW942n6JperpJTKMIn:-MXW942n6JperpJTKMIo:bqj5jyd) ([raw file](https://github.com/sillsdev/web-languageforge/blob/f89325e71fbfa701f208ae6abc82b3e0d5a2d418/docker/ld-api/Dockerfile#L11)):*
> ```Dockerfile
> # Step 1: Build app
> FROM node:14.15.4-alpine AS builder
> RUN npm i -g pnpm
> ```

Interesting that you are using npm to install an npm competitor :)  Looks good - I am always looking at ways to speed up the build.

megahirt:
per our conversation, @rmunn just add the `ld-api` service name to the `docker-compose build` line in the Makefile so that devs don't have to remember that this service should be built.  A future PR should ideally move LD-API image building to it's own repo.

megahirt (approved):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/901)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rmunn)

megahirt (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/901#-:-MXWGfkkAYX3ovXQP1DI:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rmunn)

rmunn:
Also worth noting is that one E2E test appears flaky: when I first created this PR, the "Lexicon E2E Editor Comments: comments panel" test suite had one failed test, "click +1 button on first comment". The failure error message was "Failed: stale element reference: element is not attached to the page document." But when I re-ran the build again with no changes, that test passed.

I believe the flaky test is flaky because the *previous* test is failing to include a `waitForAngular()` step at the end of the test. If I'm right, the `test/app/languageforge/lexicon/editor/editor-comments.e2e-spec.ts` file will need a "wait for angular" step at the end of the "comments panel: check regarding value is hidden when the field value matches" test. But that's for a separate PR.

rmunn:
BTW, the big advantage of [pnpm](https://pnpm.js.org/) is that it de-duplicates your `node_module` directories. If you have multiple Typescript projects sitting around on your disk, the `node_modules/typescript` directory (35 megabytes in Typescript 2.7.2) will be duplicated among each project. But with pnpm, the contents of each `node_modules` directory are actually hardlinks to files in `$HOME/.pnpm-store`. So two projects with Typescript 2.7.2 will only use a total of 35 megabytes rather than 70 megabytes of your disk space. Add in all the different dependencies-of-dependencies in a typical Javascript developer's projects folder and that adds up fast.

longrunningprocess (approved):
Much nicer endpoints, thank you for improving this.

Slick use of pnpm, never heard of it but am with Chris, if it's better, let's do it.  Let us know how things go with this utility, maybe we can incorporate it into our other node-based images.

I agree with the idea of putting LD into a separate repo if it's not coupled to LF in any way.  It would've been nice to have a tags on an image as a way for a consumer to pull in these new endpoints rather than having to time it right, e.g., language-depot-api:2.0.0 or something like that.  It's moot in our current circumstances but managing deps through semantic versioning should be our future approach.

> Reviewed 2 of 2 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/901#-:-MXXDh_3CKi80jphSh_Y:bxyam55)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rmunn)

PR #905, opened 2021-04-06
Fix failing ld-api docker build

rmunn:
This is a quick-and-dirty fix rather than a proper fix, meant to get the build up and working while I investigate the actual cause and put together a proper fix for the problem.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/905)

PR #907, opened 2021-04-07
Better fix for ld api docker build

rmunn:
This is a better fix than the one in #905 (which this PR reverts). The root cause was that pnpm just released version 6, which has some breaking changes that, surprise surprise, broke something. :-) So we'll use pnpm 5 until Vite adapts to the breaking change in pnpm 6.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/907)


rmunn:
Obsoleted by https://github.com/sveltejs/kit/issues/902 - closing without merging since the bug in pnpm 6.0.0 is now fixed with the release of pnpm 6.0.1. We can now continue to use `npm install -g pnpm` without specifying a version number.

rmunn:
Replaced by #908.

PR #925, opened 2021-04-27
Fix LfMerge config in base Docker image

rmunn:
This will allow LfMerge to find the Mongo database container. Future improvements might include looking for environment variables in LfMerge, but this is enough to get LfMerge working in the Docker environment.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/925)

### DIFF
--- a/docker-static/base-app/Dockerfile
+++ b/docker-static/base-app/Dockerfile
@@ -13,6 +13,8 @@ RUN curl -L http://linux.lsdev.sil.org/downloads/sil-testing.gpg | apt-key add -
 && apt-get update \
 && apt-get install --yes --no-install-recommends lfmerge rsyslog logrotate cron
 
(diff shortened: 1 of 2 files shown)


rmunn (commented):
Self-review for the sake of testing. I'm allowed to comment, but not approve: the Approve radio button is disabled if you're self-reviewing. Good.

image

megahirt (approved):
looks good - thanks Robin.

rmunn:
Note: I used admin access to override the approvals, so that I wouldn't have to wait for the E2E tests to complete. This PR changes the base app image, and the new base app won't be pushed to Docker Hub until the PR is merged, which means that the E2E tests will actually not be testing any changes from this PR anyway. So I decided to save 20 minutes. It's been tested by hand on QA, so I feel okay about merging this.

PR #926, opened 2021-04-27
Add publishing of base-app to build-and-publish GA

megahirt:
This adds several steps to the existing build-and-publish Github Action such that it now builds both the base-app followed by the deployable app.

The base-app must be published to Docker Hub before building the deployable app (its build references the base-app).

The build-and-publish tag was moved to this branch to test this this PR works.  The build-and-publish GA succeeded:
image

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/926)


.github/workflows/build-and-deploy-images.yml
```diff
@@ -40,24 +40,35 @@ jobs:
...
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+
+      - name: Build and tag base-app image
+        working-directory: ./docker-static/base-app
+        run: |
+          docker build . -t lf-base-app
+          docker tag lf-base-app sillsdev/web-languageforge:base-app-latest
+
```
rmunn, line 54: I think I'd like to build the base app image first, *then* log into Docker. That way if there are any errors logging in we'll still get to see the base app build happen, which will inform us of any build problems. The way it is right now, a temporary issue logging into Docker Hub could mask a build failure.

```suggestion
      - name: Build and tag base-app image
        working-directory: ./docker-static/base-app
        run: |
          docker build . -t lf-base-app
          docker tag lf-base-app sillsdev/web-languageforge:base-app-latest

      - name: Log in to Docker Hub
        uses: docker/login-action@v1
        with:
          username: ${{ secrets.DOCKERHUB_USERNAME }}
          password: ${{ secrets.DOCKERHUB_PASSWORD }}
```

rmunn (approved):
I suggest swapping the order of logging in and building (see my suggested patch), but I'll approve this regardless of whether that suggestion is taken.

PR #928, opened 2021-04-27
Mercurial needs python

rmunn:
The Send/Receive feature uses Mercurial, which is a Python app. So our base image needs a working Python 2 installation.

This PR also adds an entrypoint script which will run `/etc/init.d/rsyslog start` so that lfmerge logs will go to /var/log/syslog rather than being silently swallowed.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/928)

### DIFF
--- a/docker-static/base-app/entrypoint.sh
+++ b/docker-static/base-app/entrypoint.sh
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+# rsyslog needs to run so that lfmerge can log to /var/log/syslog
+/etc/init.d/rsyslog start
+
+# Now chain to Docker entrypoint from base php:apache image
+exec docker-php-entrypoint "$@"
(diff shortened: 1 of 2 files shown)


rmunn:
Running a `docker build` locally right now to test this, but that won't finish for a few more minutes because the Debian packaging step takes a while to download all the packages. So I'm creating the PR early before testing results are in, so that it can be reviewed before the end of the way.

rmunn:
I believe this is working from my limited local testing; time to get it reviewed, merged, and then deployed to the QA Rancher environment.

docker-static/base-app/entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+# rsyslog needs to run so that lfmerge can log to /var/log/syslog
+/etc/init.d/rsyslog start
+
+# Now chain to Docker entrypoint from base php:apache image
+exec docker-php-entrypoint "$@"
```
megahirt, line 7: this is really slick.  I like how you have chained the two entry points together.  Well done.  I would have probably chained the CMD together but the entry point chaining is better.

megahirt (approved):
Looks good, thanks Robin.

docker-static/base-app/entrypoint.sh
rmunn: I checed and some of our Dockerfiles use CMD, so entrypoint chaining lets those work without change.

PR #930, opened 2021-04-28
Run rsyslog on startup of base app

rmunn:
This fixes https://github.com/sillsdev/web-languageforge/pull/928 by restoring the CMD line to the Dockerfile, which was the cause of Apache not starting after #928 was merged. I had assumed that giving an ENTRYPOINT line without a CMD line would keep the CMD line of the parent Dockerfile, but it turns out that's not the case. Setting ENTRYPOINT resets CMD, so if you want to keep the same CMD you have to re-issue the CMD line.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/930)

### DIFF
--- a/docker-static/base-app/Dockerfile
+++ b/docker-static/base-app/Dockerfile
@@ -23,5 +23,10 @@ RUN adduser www-data fieldworks \
 && mkdir -m 02775 -p /var/www/.local \
 && chown www-data:www-data /var/www/.local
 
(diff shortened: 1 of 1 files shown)


rmunn:
Merging without review since #928 was already approved, and this is just #928 with the parent Dockerfile's CMD restored so that it works correctly.

PR #931, opened 2021-04-28
Fix race condition in E2E test startup

rmunn:
Sometimes the E2E tests were starting before the app-for-e2e container was fully available. This PR fixes that race condition.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/931)

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -144,7 +144,7 @@ services:
       - mail
       - selenium
     environment:
-      - WAIT_HOSTS=db:27017, mail:25, selenium:4444
+      - WAIT_HOSTS=db:27017, mail:25, selenium:4444, app-for-e2e:80
       - ENVIRONMENT=development
       - LANGUAGE_DEPOT_API_TOKEN=bogus-development-token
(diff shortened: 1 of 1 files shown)


rmunn:
Running E2E tests locally, I would sometimes get "Angular could not be found on the page http://app-for-e2e/auth/logout" on the first test run (and only the first test). A bit of experimentation showed that adding `app-for-e2e:80` to the list of hosts being waited for was enough to run the tests consistently, and correctly, six or seven times in a row (and probably forever, but I stopped after a while).

megahirt (approved):
This is a great little addition!  I wonder if @longrunningprocess has also been experiencing failures locally, and if this will fix it?  Of course this race condition will vary machine to machine.

longrunningprocess:
> This is a great little addition! I wonder if @longrunningprocess has also been experiencing failures locally, and if this will fix it? Of course this race condition will vary machine to machine.

yes I definitely was, based on the fact I'm seeing #932 I'm guessing this doesn't do it alone...but I'm going to run some tests on my machine to confirm.

PR #932, opened 2021-04-28
Try to fix intermittent E2E failure

rmunn:
One E2E test is failing intermittently with some kind of race condition. This might fix it.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/932)

### DIFF
--- a/test/app/languageforge/lexicon/lexicon-new-project.e2e-spec.ts
+++ b/test/app/languageforge/lexicon/lexicon-new-project.e2e-spec.ts
@@ -221,6 +221,7 @@ describe('Lexicon E2E New Project wizard app', () => {
 
       await page.namePage.projectNameInput.sendKeys('a' + Key.TAB);
       await page.namePage.projectNameInput.clear();
+      await browser.waitForAngular();
       await page.namePage.projectNameInput.sendKeys(Key.TAB);
       await browser.sleep(CHECK_PAUSE);
       expect<any>(await page.namePage.projectCodeExists.isPresent()).toBe(false);


megahirt (approved):
LGTM

longrunningprocess:
I'm finding that #931 resolved the issue altogether for me...was there an explicit need for this or did you do it just to be safe?  It feels redundant to me, I'm wondering whether this was the right solution all along but #931 makes "more" sense to me :-D

longrunningprocess:
I did a little TC testing and this one is not as reliable a fix as the wait approach taken in #931 .  Maybe you intended to have both, not sure, but this does not keep the first test from failing when I run the build in the docker agent rather than the VM the way the wait does.

rmunn:
> I'm finding that #931 resolved the issue altogether for me...was there an explicit need for this or did you do it just to be safe? It feels redundant to me, I'm wondering whether this was the right solution all along but #931 makes "more" sense to me :-D

This fixes a different failure than #931. #931 fixes the issue where the *first* test fails because the `app-for-e2e` container isn't available yet. This PR fixes https://build.palaso.org/buildConfiguration/LanguageForgeDocker_E2eTests/267145?buildTab=tests&status=failed which is due to a different race condition.

longrunningprocess:
> > I'm finding that #931 resolved the issue altogether for me...was there an explicit need for this or did you do it just to be safe? It feels redundant to me, I'm wondering whether this was the right solution all along but #931 makes "more" sense to me :-D
> 
> This fixes a different failure than #931. #931 fixes the issue where the _first_ test fails because the `app-for-e2e` container isn't available yet. This PR fixes https://build.palaso.org/buildConfiguration/LanguageForgeDocker_E2eTests/267145?buildTab=tests&status=failed which is due to a different race condition.

very well, if you don't mind rebasing (or letting me know I can, I'll get this merged for you.

rmunn:
There's no need to rebase unless there's a merge conflict; the "Squash and merge" button will effectively do a rebase because it builds a commit on top of the current head of the `master` branch whenever you push the button. Once the E2E tests finish running, go ahead and do a squash-and-merge.

longrunningprocess (approved):

PR #935, opened 2021-05-03
Build a production version of the UI app

longrunningprocess:
The contents of `src/dist` in this image will be much smaller but they are not hashed (or cache-busted), e.g., `main.bundle.jr2jbf.js`, as I expected.  Maybe that was handled differently in the past @megahirt ?

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/935)


megahirt (approved):
LGTM Thank you.

PR #936, opened 2021-05-04
Get XDebug 3 working in Docker

megahirt:
These changes enable XDebug PHP debugging using VSCode + PHP Debug extension.

I used a number of instructions/tidbits from the internet to figure out how to get this working.  Most of the documentation on the internet refers to XDebug 2 configurations, however we are running XDebug v3 which has different configuration options.

In addition, for this to work, the XDebug program running inside Docker needs to connect to VSCode on the host, port 9003.  Normally services running inside Docker don't need the ability to connect to services running on the host, however XDebug is a different kind of animal and needs this ability.

I added the `extra_hosts` in docker-compose to make this possible.  This requires Docker v20.10 or later to use the special `host.docker.internal` hostname.

Some additional notes on debugging PHP using VSCode and XDebug:

- Because PHP composer dependencies only exist inside the docker container and not in your source tree, XDebug cannot step into vendor/composer dependencies unless you also download dependencies on your host.  This is an extra step that might not be practical.  If you get a VSCode error stating that it cannot find a file, check to make sure that it is not in the vendor folder.

VSCode allows you to break on 'everything'.  I suggest turning this off so that the debugger doesn't break inside composer dependencies by default.

XDebug can probably safely be included in the production build since it defaults to "off".  If we create a separate app build for production, then XDebug can be omitted entirely.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/936)


longrunningprocess:
> XDebug can probably safely be included in the production build since it defaults to "off". If we create a separate app build for production, then XDebug can be omitted entirely.

I'm guessing we'll have a separate image for this build since we don't `wait` in there either and we'll probably want a different `php.ini` as well, not sure what else at this point but I hope to chat through it a little at tonight's stand-up.

longrunningprocess (approved):

longrunningprocess:
FYI @megahirt , there was a merge conflict that I resolved locally and force pushed up.

PR #937, opened 2021-05-04
Build a production version of the PHP portion of the app

longrunningprocess:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/937)


docker/Makefile
```diff
@@ -46,7 +46,11 @@ deployable-app: build-ui
 
 .PHONY: prod
 prod: build-ui
-	docker build -f ui-builder/Dockerfile.prod -t lf-ui-builder:prod .
+	docker build -f ui-builder/Dockerfile.prod . -t lf-ui-builder:prod
+	docker build app-base -t lf-app-base
+	docker build -f app-base/Dockerfile.prod app-base -t lf-app-base:prod
+	docker-compose build app
```
megahirt, line 52: It looks like you are taking advantage of `docker-compose build` so you don't have to write `docker build` lines, correct?  It's fine the way it is presumably, but I note that it's opaque at this point what docker-compose is actually building and which images you will actually use.

docker/app/Dockerfile
```diff
@@ -1,25 +1,26 @@
-FROM sillsdev/web-languageforge:base-app-latest
+FROM sillsdev/web-languageforge:app-base-latest
```
megahirt, line 1: before this PR gets merged, it would be great to ensure that this image is pushed and exists so that the master build doesn't break.

docker/app/Dockerfile.prod
```diff
@@ -0,0 +1,25 @@
...
+
+# php customizations
+RUN mv $PHP_INI_DIR/php.ini-production $PHP_INI_DIR/php.ini
+COPY docker/app/customizations.php.ini $PHP_INI_DIR/conf.d/
+
+# apache2 customizations
+RUN a2enmod headers rewrite
+COPY docker/app/000-default.conf /etc/apache2/sites-enabled
+
+# install app dependencies
```
megahirt, line 11: The composer step seems ripe for a MULTI STAGE build since the only thing we actually need is a directory full of files that composer generates but not the tool itself.

docker/app/Dockerfile.prod
```diff
@@ -0,0 +1,25 @@
+FROM lf-app-base:prod
+
+# php customizations
+RUN mv $PHP_INI_DIR/php.ini-production $PHP_INI_DIR/php.ini
+COPY docker/app/customizations.php.ini $PHP_INI_DIR/conf.d/
+
+# apache2 customizations
```
megahirt, line 7: it's too bad there is a lot of duplication cut/paste between the two Dockerfiles and that there doesn't seem to be a supported way to INCLUDE another Dockerfile.  There's a long heated thread about people for/against (mostly for) an INCLUDE directive over at https://github.com/moby/moby/issues/735 and that led me to https://github.com/edrevo/dockerfile-plus which allows you to organize common commands into an include file so you can DRY.  I don't think anything is really actionable, just thought I would point out the duplication.  Not sure exactly how to minimize that with vanilla Dockerfile syntax.

docker/app/customizations.php.ini
```diff
@@ -0,0 +1,3 @@
+memory_limit = 256M
+post_max_size = 60M
+upload_max_filesize = 60M
```
megahirt, line 3: this is awesome - well done in pulling these customizations out!

megahirt (approved):
Looks good!  This is a great contribution.  I have no blocking comments, just observations and a request to publish the renamed base image as soon as the PR is merged to minimize a broken master build.

docker/Makefile
longrunningprocess: it has to do with whether or not the existing definition in the `docker-compose.yml` will work as-is or not.  For example, line 49 requires a different `Dockerfile` than the one already established in the `docker-compose.yml` so it seems silly to add another entry just to reference a different `Dockerfile`.

docker/app/Dockerfile
longrunningprocess: yep, already did that in early commits on this branch (https://github.com/sillsdev/web-languageforge/pull/937/commits/1c4b8e772cb8f759e8bb1d17f48a2b10810d24da).

My hope is that most devs are not powerwashing, the CI environment is the same so everyone should already have this old image anyway.  Once this merges into master all will be addressed in a normal CI fashion.

docker/app/Dockerfile.prod
longrunningprocess: agreed on the disappointment, this is symptomatic of some of the coupling.  Also, I've never com across the need for `include` so I think it's probably just a matter of engineering things appropriately.

docker/app/customizations.php.ini
longrunningprocess: thanks, I was happy to learn more about php's config options, this is much better.

longrunningprocess:
> Looks good! This is a great contribution. I have no blocking comments, just observations and a request to publish the renamed base image as soon as the PR is merged to minimize a broken master build.

good thinking, I had already published it early in the branch commits.

PR #939, opened 2021-05-05
Restore refreshDeps.sh script (used by LfMerge build)

rmunn:
The LfMerge build on Jenkins expects the refreshDeps.sh script to exist in this repo. (The LfMerge build clones the web-languageforge repo to get access to some of its PHP files during unit testing). The simplest way to fix the Jenkins LfMerge build is to just make sure that there's a refreshDeps.sh script in the LF Git repo, for now. Eventually I'll update the Jenkins LfMerge build, but that's a more complicated process.

Thankfully, now that we're no longer using Gulp, it's much easier to see what refreshDeps.sh is doing, and we also no longer need to specify the applicationName in the script either. So it's just "composer install" and "npm install" in the appropriate directories.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/939)

### DIFF
--- a/refreshDeps.sh
+++ b/refreshDeps.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+# Usage
+# ./refreshDeps.sh : refreshes PHP and NPM dependencies for LF
+
+(cd src && composer install)
+npm install
+echo -e "\n---------Finished refreshing npm and composer----------"


rmunn:
The parentheses in the `(cd src && composer install)` line run that in a sub-shell, which has its own working directory. This lets us run `npm install` in the root directory without needing an extra `cd ..` step.

megahirt:
@rmunn it looks like this PR is missing a commit?  I think you intended to simplify the buildDeps.sh script to just `composer install` but I'm still seeing the original one with gulp.

Also - does LFMerge even need anything in npm land?  I would expect it to only use composer dependencies, so you can probably get away without doing npm install.

rmunn:
Yep, somehow I pushed the wrong version of `refreshDeps.sh`. I'll fix that.

PR #940, opened 2021-05-05
Fix flaky E2E test

rmunn:
Another attempt to fix the flaky E2E test. I looked into it and noticed that the "no error" check only looks for the presence of the string "alert" in the element's class attribute... but if the form is valid then there's supposed to be a green checkmark with the class `alert-success`. So this PR is attempting to fix the flaky test by only considering the text `alert-danger` to represent an error, and not failing the test if `alert-success` is present in the element's class attribute.

I'll create this PR in draft mode, then set it as ready to review & merge if the E2E tests pass on TeamCity.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-languageforge/940)

### DIFF
--- a/test/app/languageforge/lexicon/shared/new-lex-project.page.ts
+++ b/test/app/languageforge/lexicon/shared/new-lex-project.page.ts
@@ -30,7 +30,7 @@ export class NewLexProjectPage {
 
   formStatus = {
     async expectHasNoError() {
-      expect(await element(by.id('form-status')).getAttribute('class')).not.toContain('alert');
+      expect(await element(by.id('form-status')).getAttribute('class')).not.toContain('alert-danger');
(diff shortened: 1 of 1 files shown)


rmunn:
The flaky test passed, but a *different* test failed. Re-running E2E tests again to double-check.

rmunn:
That didn't work: the element still contains `alert-danger` and the test fails. OTOH, I think the semantics of this PR are correct and we really should be testing for `alert-danger` instead of `alert`, so I'll un-draft this PR and let it be reviewed. But something else, perhaps involving `browser.wait()` waiting for some condition, will probably be needed to make this test actually pass consistently.

longrunningprocess (approved):

PR #943, opened 2021-05-06
Added some volume mappings to help developers

longrunningprocess:
The only thing I could think devs might've been doing is touching the index or config files expecting to see the changes but those weren't volume mapped... 🤷‍♂️


megahirt (approved):
These two lines are fine to add.  I haven't needed them yet personally but I'm not developing much yet either.

PR #944, opened 2021-05-11
prevent user from unchecking "word" in project dictionary configuration #938

### LINKED ISSUE #938: prevent user from unchecking "word" in dictionary configuration
In the dictionary configuration tool, it is possible to uncheck all the fields, including "word" / "lexeme".  This field is required by the UI and so should never be unchecked.

palmtreefrb:
Fixes #938
Changed several methods that control "select", "select all" check boxes on the project dictionary configuration page. The changes prevents user from un-checking word(lexeme) field. This field must always remained checked for the project list page to function properly.


src/angular-app/languageforge/lexicon/settings/configuration/configuration-fields.component.html
```diff
@@ -276,7 +276,7 @@
                     <input class="position-static" data-ng-class="'checkbox-group-' + $index"
                            data-ng-model="group.show"
                            data-ng-change="$ctrl.checkIfAllGroupSelected(entryField, $ctrl.unifiedViewModel.entryFields.settings, $ctrl.unifiedViewModel.entryFields.selectAllColumns, $index)"
-                           aria-label="..." type="checkbox">
+                           aria-label="..." type="checkbox" data-ng-disabled="$ctrl.requiredConfigurationFields(entryField)">
```
megahirt, line 279: yes!

src/angular-app/languageforge/lexicon/settings/configuration/configuration-fields.component.ts
```diff
@@ -72,6 +72,7 @@ export class FieldsConfigurationController implements angular.IController {
   overrideAll = ConfigurationFieldUnifiedViewModel.overrideAll;
   resetInputSystemsForRole = ConfigurationFieldUnifiedViewModel.resetInputSystemsForRole;
   resetInputSystemsForGroup = ConfigurationFieldUnifiedViewModel.resetInputSystemsForGroup;
+  requiredConfigurationFields = ConfigurationFieldUnifiedViewModel.disableRequiredFields;
```
megahirt, line 75: name change: remove "disabled"
```suggestion
  ensureRequiredConfigurationFields = ConfigurationFieldUnifiedViewModel.ensureRequiredFields;
```

src/angular-app/languageforge/lexicon/settings/configuration/field-unified-view.model.ts
```diff
@@ -96,16 +95,52 @@ export class ConfigurationFieldUnifiedViewModel {
...
 
-  static selectAllRow(setting: SettingsBase, settings: SettingsBase[], selectAll: SettingsBase): void {
+  static disableRequiredFields(fieldSettings: FieldSettings): boolean {
+    const disableFieldList = DisableFields.disableFieldList();
+    let disabled = false;
+    if (disableFieldList.includes(fieldSettings.fieldName)) {
+      disabled = true;
+    }
+    return disabled;
+  }
```
megahirt, line 105: name change to replace "disabled" with "required".
perhaps you could also simplify the function to a single `return` statement:

```suggestion
  static ensureRequiredFields(fieldSettings: FieldSettings): boolean {
    return RequiredFields.requiredFieldList().includes(fieldSettings.fieldName);
  }
```

src/angular-app/languageforge/lexicon/settings/configuration/field-unified-view.model.ts
```diff
@@ -96,16 +95,52 @@ export class ConfigurationFieldUnifiedViewModel {
...
+    const disableFieldList = DisableFields.disableFieldList();
     for (const role of roles) {
-      setting[role] = setting.isAllRowSelected;
+      fieldSettings[role] = fieldSettings.isAllRowSelected;
       ConfigurationFieldUnifiedViewModel.checkIfAllRoleColumnSelected(settings, selectAll, role);
     }
-    for (const group of setting.groups) {
-      group.show = setting.isAllRowSelected;
+    for (const group of fieldSettings.groups) {
+      if (disableFieldList.includes(fieldSettings.fieldName)) {
```
megahirt, line 115:
```suggestion
      if (requiredFieldList.includes(fieldSettings.fieldName)) {
```

src/angular-app/languageforge/lexicon/settings/configuration/field-unified-view.model.ts
```diff
@@ -96,16 +95,52 @@ export class ConfigurationFieldUnifiedViewModel {
...
+    let disabled = false;
+    if (disableFieldList.includes(fieldSettings.fieldName)) {
+      disabled = true;
+    }
+    return disabled;
+  }
+
+  static selectAllRow(fieldSettings: FieldSettings, settings: SettingsBase[], selectAll: SettingsBase): void {
     const roles = RoleType.roles();
+    const disableFieldList = DisableFields.disableFieldList();
```
megahirt, line 109:
```suggestion
    const requiredFieldList = RequiredFields.requiredFieldList();
```

src/angular-app/languageforge/lexicon/settings/configuration/field-unified-view.model.ts
```diff
@@ -96,16 +95,52 @@ export class ConfigurationFieldUnifiedViewModel {
...
+      }
       ConfigurationFieldUnifiedViewModel
-        .checkIfAllGroupColumnSelected(settings, selectAll, setting.groups.indexOf(group));
+        .checkIfAllGroupColumnSelected(settings, selectAll, fieldSettings.groups.indexOf(group));
+    }
+  }
+
+  static checkIfAllEntryFieldsRowSelected(fieldSettings: FieldSettings): void {
+    const roles = RoleType.roles();
+    const disableFieldList = DisableFields.disableFieldList();
```
megahirt, line 127:
```suggestion
    const requiredFieldList = RequiredFields.requiredFieldList();
```

src/angular-app/languageforge/lexicon/settings/configuration/field-unified-view.model.ts
```diff
@@ -96,16 +95,52 @@ export class ConfigurationFieldUnifiedViewModel {
...
+    for (const role of roles) {
+      if (!fieldSettings[role]) {
+        fieldSettings.isAllRowSelected = false;
+        break;
+      }
+    }
+    if (fieldSettings.isAllRowSelected) {
+      for (const group of fieldSettings.groups) {
+        if (!group.show) {
+          if (disableFieldList.includes(fieldSettings.fieldName)) {
```
megahirt, line 138:
```suggestion
          if (requiredFieldList.includes(fieldSettings.fieldName)) {
```

src/angular-app/languageforge/lexicon/settings/configuration/field-unified-view.model.ts
```diff
@@ -652,3 +687,10 @@ export interface GroupList {
   label: string;
   userId: string;
 }
+
+/* disable all fields that should not be available for changes by the user */
+export class DisableFields {
+  static disableFieldList(): string[] {
+    return ['lexeme'];
+  }
+}
```
megahirt, line 696:
```suggestion
export class RequiredFields {
  static requiredFieldList(): string[] {
    return ['lexeme'];
  }
}
```

megahirt (changes requested):
hi Rick!  This is really great work.  The only request I have is to rename the variables and methods from "disabled" to "required" in most cases below.  I feel that "required" more accurately describes what the field is, and then in the UI you simply say that required fields are "disabled".  As you asked about in standup, there may be other uses for a structure that tells us which fields are system-required, and naming it "required" does not lock us in to the disable feature, as far as naming goes.

megahirt:
I should also say that, while I used the "suggest changes" in my review, I didn't test out any changes so it's very possible that I missed or mistyped something and that accepting my changes breaks the application :)  Feel free to make the changes in your IDE if that is simpler for you.

palmtreefrb:
The suggested changes seem reasonable and I committed your suggestions.

src/angular-app/languageforge/lexicon/settings/configuration/field-unified-view.model.ts
```diff
@@ -600,7 +630,7 @@ export class ConfigurationFieldUnifiedViewModel {
 }
 
 export class Group {
-  show: boolean = false;
+  show: boolean = true;
```
rmunn, line 633: Why change the default to `true` here? I don't see the reasoning behind this change.

src/angular-app/languageforge/lexicon/settings/configuration/field-unified-view.model.ts
```diff
@@ -135,10 +165,10 @@ export class ConfigurationFieldUnifiedViewModel {
     }
   }
 
-  static selectAllGroupColumn(settings: SettingsBase[], selectAll: SettingsBase, groupIndex: number): void {
+  static selectAllGroupColumn(settings: FieldSettings[], selectAll: SettingsBase, groupIndex: number): void {
```
rmunn, line 168: You can't change SettingsBase here, as this function is also called in the input systems settings portion of this view ([here](https://github.com/sillsdev/web-languageforge/blob/59dcc8b2a1fd41f84d47dea1fbfcf729c5bb444f/src/angular-app/languageforge/lexicon/settings/configuration/configuration-fields.component.html#L63)). When it receives input systems settings, there's no `fieldName` property on the settings object; instead there's a `tag` property containing "en" for English, etc.

So what you'll need to do is check for the presence of a `fieldName` property, and if it's present, cast the SettingsBase object as FieldSettings before passing it to checkIfAllEntryFieldsRowSelected.

src/angular-app/languageforge/lexicon/settings/configuration/field-unified-view.model.ts
```diff
@@ -96,16 +95,47 @@ export class ConfigurationFieldUnifiedViewModel {
     }
   }
 
-  static selectAllRow(setting: SettingsBase, settings: SettingsBase[], selectAll: SettingsBase): void {
+  static ensureRequiredFields(fieldSettings: FieldSettings): boolean {
+    return RequiredFields.requiredFieldList().includes(fieldSettings.fieldName);
+  }
+
+  static selectAllRow(fieldSettings: FieldSettings, settings: SettingsBase[], selectAll: SettingsBase): void {
```
rmunn, line 102: You can't change the type from SettingsBase to FieldSettings here, because input systems settings also use this function. See my comment on the selectAllGroupColumn method for more details.

rmunn (changes requested):
There are a few places where you've changed the type of a function input from SettingsBase to FieldSettings, but those functions are also called with InputSystemSettings objects, which might be why the E2E tests are failing for this PR. See my comments on individual code sections for more details.

src/angular-app/languageforge/lexicon/settings/configuration/field-unified-view.model.ts
megahirt: @palmtreefrb didn't provide an explanation in his PR or commit, but I am pretty sure this is a desired change.  We want the default to be "true" so that when an admin adds a new configuration column for a user, all of the boxes are checked.  It would seem to me that the sensible default is "true" and the admin can remove field visibility one at a time if that is desired.  I think the current state of "false" is less desired because it is likely to require more clicking than the default "true" state.

src/angular-app/languageforge/lexicon/settings/configuration/field-unified-view.model.ts
megahirt: good catch @rmunn

megahirt (approved):
I approve of this PR with the following assumptions:

1) @rmunn 's approval
2) E2E tests are passing

Thanks Rick!

palmtreefrb:
> There are a few places where you've changed the type of a function input from SettingsBase to FieldSettings, but those functions are also called with InputSystemSettings objects, which might be why the E2E tests are failing for this PR. See my comments on individual code sections for more details.

The above was reverted back.I added the fieldName property to SettingsBase as both FieldSettings and InputSettings require it now.

src/angular-app/languageforge/lexicon/settings/configuration/field-unified-view.model.ts
palmtreefrb: The above was reverted back.I added the fieldName property to SettingsBase as both FieldSettings and InputSettings require it now.

palmtreefrb:
Changes e2e tests to expect all user configuration checkbox to be checked.

megahirt:
This PR currently has one failing E2E test

image

megahirt:
The failing test is in [config-fields.e2e-spec.ts Line 675](https://github.com/sillsdev/web-languageforge/blob/master/test/app/languageforge/lexicon/settings/config-fields.e2e-spec.ts#L675)
image

megahirt:
It's difficult to debug this one test because we don't yet have a documented way to run E2E test in a small section.  We only have a "run all tests" make target but nothing that lets us run a single file of tests.  So this is challenging.

palmtreefrb:
> It's difficult to debug this one test because we don't yet have a documented way to run E2E test in a small section. We only have a "run all tests" make target but nothing that lets us run a single file of tests. So this is challenging.
What is "can remove member-specific user settings" test case purpose and requirement?

megahirt:
> What is "can remove member-specific user settings" test case purpose and requirement?

I believe the purpose is to demonstrate that you can remove the user columns that were set up in previous tests in this file.  The way that the test goes about checking that the column is removed might be overkill, or even bogus based upon the error.
We should have a test that demonstrates that you can remove a member-specific column, but I'm not convinced this test expresses that in the best or most succinct way.

palmtreefrb:
> > What is "can remove member-specific user settings" test case purpose and requirement?
> 
> I believe the purpose is to demonstrate that you can remove the user columns that were set up in previous tests in this file. The way that the test goes about checking that the column is removed might be overkill, or even bogus based upon the error.
> We should have a test that demonstrates that you can remove a member-specific column, but I'm not convinced this test expresses that in the best or most succinct way.

Thanks for the quick reply. Guess I will need some instruction and guidance from Robin on how to set up a dev environment to debug and test the e2e tests.

megahirt:
That's kind of what I was thinking.  Either Robin or I will need to spend
some time to document the process and figure out how to run e2e tests
file-by-file.  Debugging E2E tests are quite different than debugging the
actual application due to running the tests through an automated Selenium
environment.  This makes debugging much more difficult, so different
debugging tactics like setting pause() are necessary.

rmunn (approved):
Now that the E2E tests are passing, this LGTM.

rmunn:
@megahirt One question about this PR. It currently disables unchecking "Word" only for individual users ("groups" in the config code's slightly-confusing terminology) but still allows unchecking it for certain roles. Is that the intended design? Or was the intent that neither roles nor individual users can have the "Word" field unchecked?

If it's the latter, this will still require a couple of changes before it's truly ready to merge. They should be relatively minor now that the functionality is in place (just need to apply the same attribute to the HTML checkboxes for the four roles as we did for the "groups" section of the HTML), so it shouldn't take too long to make that change. Perhaps we can discuss it at tomorrow's standup meeting.

megahirt:
The intention was to make the "word" checkbox always checked and disabled
to prevent changes.

Kind regards,

Chris

src/angular-app/languageforge/lexicon/settings/configuration/field-unified-view.model.ts
```diff
@@ -94,19 +93,50 @@ export class ConfigurationFieldUnifiedViewModel {
...
+    const roles = RoleType.roles();
+    const requiredFieldList = RequiredFields.requiredFieldList();
+    setting.isAllRowSelected = true;
+    for (const role of roles) {
+      if (!setting[role]) {
+        setting.isAllRowSelected = false;
+        break;
+      }
+    }
+    if (setting.isAllRowSelected) {
```
rmunn, line 128: Found a bug here. I can get the "Select All" column to disable the lexeme field by the following steps:

1. Uncheck the checkbox for manager (could do this for any role)
2. Uncheck the "Select All" column for a user (a "group" in our code's current terminology)

Result: this `if` statement is never executed, and so the "Select All" logic ends up turning off the checkbox that we're trying to force to be always on.

I'll fix this bug, and while I'm at it I'll add the "can't unselect the lexeme field" logic to roles as well since Chris has indicated that that was indeed the intent of the feature. Which will involve a bit of a code rewrite, so I'm going to set my review status to "changes requested" in order to block this PR from merging until I've finished fixing the bug.

rmunn (changes requested):
I found a bug (see comment above for details), and I'm working on fixing it. Don't merge until I've finished, at which point I'll set my review status back to Approve.

src/angular-app/languageforge/lexicon/settings/configuration/field-unified-view.model.ts
rmunn: Fixed now.

rmunn (approved):
Bug fixed, and logic updated to what it was intended to be (checkbox is disabled for required fields in user *and* role columns). Ready to merge now.

PR #950, opened 2021-05-19
enable xDebug for PHP unit-tests

megahirt:
This is a follow up PR to the initial enabling of XDebug PHP debugging.  With this change, you can now set a breakpoint in a PHP test and debug it in VSCode using XDebug.

In order to use XDebug for Unit Tests, 3 lines in the docker-compose file must be uncommented.  They are left commented because the XDebug settings do not play well with our CI environment.

Specifically the 
```
   # extra_hosts:
      # - "host.docker.internal:host-gateway"
```
fails on TeamCity for reasons I don't understand right now.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -7,7 +7,8 @@
       "request": "launch",
       "port": 9003,
       "pathMappings": {
-        "/var/www/html/": "${workspaceRoot}/src/"
+        "/var/www/html/": "${workspaceRoot}/src/",
+        "/var/www/test/": "${workspaceRoot}/test/"
       }
     },
     {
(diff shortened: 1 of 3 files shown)


longrunningprocess:
> In order to use XDebug for Unit Tests, 3 lines in the docker-compose file must be uncommented. They are left commented because the XDebug settings do not play well with our CI environment.

:-( unfortunate but thankfully this is not something I would anticipate being used very much, thanks for working all this out for those devs that will use this technique.

longrunningprocess (approved):

longrunningprocess:
as part of the README updates, maybe consider dropping links to your youtube walk-throughs as well

megahirt:
> Do you think there's anything we might want to add to the README about this?

@longrunningprocess good call.  I will add a commit with changes to the README.

longrunningprocess (approved):
excellent, thank you

longrunningprocess:
re-running E2E tests in case the failure was just a flaky one.

PR #952, opened 2021-05-20
parameterized the mail host as an ENV var

longrunningprocess:
there may be times when the host name of the mailer is not "mail" so it's better to have this parameterized so there is more flexibility in deployments.

### DIFF
--- a/src/Api/Library/Shared/Communicate/Email.php
+++ b/src/Api/Library/Shared/Communicate/Email.php
@@ -1,8 +1,8 @@
 <?php
 
-use Sil\PhpEnv\Env; // https://github.com/silinternational/php-env#class-env-summary-of-functions
-
 namespace Api\Library\Shared\Communicate;
+
+use Sil\PhpEnv\Env; // https://github.com/silinternational/php-env#class-env-summary-of-functions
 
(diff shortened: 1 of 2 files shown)


src/Api/Library/Shared/Communicate/Email.php
```diff
@@ -14,7 +16,7 @@ class Email
     public static function send($from, $to, $subject, $content, $htmlContent = '')
     {
         // Create the Transport
-        $transport = \Swift_SmtpTransport::newInstance('mail', 25);
```
megahirt, line 17: I note that you removed the port number from the constructor, however I checked the [source code](https://github.com/swiftmailer/swiftmailer/blob/master/lib/classes/Swift/SmtpTransport.php#L33) and this parameter has a default of 25 as expected, so good to go.

megahirt (approved):
This is a great, clean addition.  Thank you!

src/Api/Library/Shared/Communicate/Email.php
longrunningprocess: yes, that's why I did it, thanks for double-checking.

PR #955, opened 2021-05-26
make languageforge.org the default for Website::get()

megahirt:
This is a small "hack" necessary for Language Forge to function with production data under any site e.g. localhost, qa.languageforge.org or languageforge.org.  There is no longer a business case to keep a separate site distinction in this code-base.  A future change may remove the Website class entirely.

### DIFF
--- a/scripts/tools/csvInsights.php
+++ b/scripts/tools/csvInsights.php
@@ -9,4 +9,4 @@ use Api\Library\Shared\Website;
 
 (php_sapi_name() == 'cli') or die('this script must be run on the command-line');
 
-ProjectInsightsDto::csvInsightsToFile(Website::get('languageforge.org'), 'languageforge.csv');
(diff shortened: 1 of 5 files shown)


longrunningprocess (approved):

longrunningprocess:
I suppose the only headache this might introduce is following validation links when testing the sign-up process in QA but that should be easy enough to just change the address in the browser after clicking the link.

longrunningprocess:
looks like everything is good here, I'll go ahead and merge for you.

megahirt:
Yes, that's a good point.  I guess it's a small price to pay to get this
working.

rmunn:
One issue this is causing is that OAuthBase.php, which uses `Website->baseUrl()` to figure out the redirect URL to return to after login, is always getting `languageforge.org` instead of `qa.languageforge.org` even when trying to log in to the QA site. So we can't do Google OAuth logins on the QA site any more.

PR #967, opened 2021-05-27
Long-term deployment strategy

longrunningprocess:
The goal for `staging` will be the following:

Any push to `staging` or fast-forwarding the branch to a commit on `master` will trigger the build-n-publish workflow on GHA.  While that's happening, TC should be running unit and e2e tests.  Once those tests finish successfully, TC should deploy the infrastructure to QA.  It's important at this point to know that the build-n-publish workflow in GHA is in no way coupled/serialized to the unit/e2e tests that are running, these two platforms should be running at the same time and, if they do, the newly built images will be available by the time the QA deployment goes looking for the latest images.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,8 @@ To use **Language Forge** go to [languageforge.org](https://languageforge.org).
 To report an issue using the **Language Forge** service, email "languageforgeissues @ sil dot org".
 
 ## Special Thanks To ##
-
+- ![BrowserStack Logo](readme_images/browserstack-logo.png "BrowserStack") for mobile device testing.
+-
 [![Bugsnag logo](readme_images/bugsnag-logo.png "Bugsnag")](https://bugsnag.com/blog/bugsnag-loves-open-source) for error reporting.
 
 ## Developers ##
--- a/README.md
+++ b/README.md
@@ -12,30 +12,23 @@ To report an issue using the **Language Forge** service, email "languageforgeiss
 
 ## Special Thanks To ##
 
-![BrowserStack Logo](readme_images/browserstack-logo.png "BrowserStack") for end-to-end test automation.
-
 [![Bugsnag logo](readme_images/bugsnag-logo.png "Bugsnag")](https://bugsnag.com/blog/bugsnag-loves-open-source) for error reporting.
 
 ## Developers ##
 
-We use [Gitflow](http://nvie.com/posts/a-successful-git-branching-model/) with two modifications:
+We use [Gitflow](http://nvie.com/posts/a-successful-git-branching-model/) mostly, however we do not utilize a `develop` branch.  All work starts and returns to the `master` branch.  Deployment of software to a staging and/or a production environment occur through the use of the respective branches being fast-forwarded to the appropriate commit on `master` which will in turn kick off our automated deployment processes.
 
-- The Gitflow **master** branch is our **live** branch.
-- The Gitflow **develop** branch is our **master** branch. All pull requests go against **master**.
-
-We merge from **master** to testing (**qa** branch) then ship to **live**.
-
-| Master Branch | QA/Staging Branch | Live/Production Branch |
+| Master Branch | Staging Branch (for testing) | Production Branch |
 | ------------- | --------- | ----------- |
-| `master` | `lf-qa` | `lf-live` |
+| `master` | `staging` | `prod` |
 
 ### CI Builds ###
 
 Status of builds from our continuous integration (CI) [server](https://build.palaso.org):
 
-| PHP Unit Tests | E2E Tests | Staging | Production |
-| ----------- | ---------- | -- | ---- |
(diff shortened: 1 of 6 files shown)


longrunningprocess:
left to do atm:

- [ ] ensure the TC trigger for the deployment job is only triggered on pushes to `staging`
- [ ] need to parameterize the `app-deployment` file for the staging tag on the image.
- [ ] clean up `-latest` images on Docker Hub that are being replaced by `-[staging|prod]`
- [ ] build out `prod` deployment mechanisms and processes once `staging` is all done

README.md
```diff
@@ -12,30 +12,23 @@ To report an issue using the **Language Forge** service, email "languageforgeiss
 
 ## Special Thanks To ##
 
-![BrowserStack Logo](readme_images/browserstack-logo.png "BrowserStack") for end-to-end test automation.
-
```
megahirt, line 16: Let's leave this in the README.md as we still have a registered OSS license for Browser Stack and are still using it.  I will make a separate GH suggestion above to add it back in.   We aren't using it for e2e testing anymore but are actively using it for device testing.

README.md
```diff
@@ -12,30 +12,23 @@ To report an issue using the **Language Forge** service, email "languageforgeiss
 
 ## Special Thanks To ##
```
megahirt, line 14:
```suggestion
- ![BrowserStack Logo](readme_images/browserstack-logo.png "BrowserStack") for mobile device testing.
-
```

megahirt (changes requested):
> It's important at this point to know that the build-n-publish workflow in GHA is in no way coupled/serialized to the unit/e2e tests that are running, these two platforms should be running at the same time and, if they do, the newly built images will be available by the time the QA deployment goes looking for the latest images.

So if I understand this correctly, the build-and-publish GA (step A) take about 10 minutes to run to completion, meanwhile the Unit tests + E2E tests (step B) take 30 minutes to complete.    Step A and B run in parallel, and B always takes longer to complete than A, from our experience.  Only after the tests are complete and passing is the deployment step on TC (step C) activated which deploys the code to our staging server.  Step C cannot start until step B has finished.  C also has a dependency on the existence of A, and we feel comfortable with this arrangement even though A and C are not synchronized.

One thing not mentioned is that the tests, step B, also depend upon the existence of A.  I would guess that in this scenario the unit tests/e2e tests would actually be run using a previous version of `sillsdev/web-languageforge:app-base-staging` which would not likely be updated until step A is finished.  I'm documenting this mostly so we can agree on the state of things, but what you have going is certainly a great step forward and a great improvement!  In our previous deployment, we actually didn't run any tests for staging or production.  Our previous practice was to only run tests on the master branch.

I'll also note that with this change we no longer have a way to deploy to app-base-latest for TC/locally run code.  This *could* be an issue in the future, but I'm unsure how often we would actually encounter this scenario.  A possible workaround is to build the app-base-latest locally, tag it, and then point the local Dockerfile to the updated image.

Requested change:
- put the Browser Stack credit back in the readme (we still use it).

Thanks again Billy!  This is great work that we can merged in straight away!

README.md
longrunningprocess: thanks, I did not realize that.

megahirt (approved):

longrunningprocess:
> > It's important at this point to know that the build-n-publish workflow in GHA is in no way coupled/serialized to the unit/e2e tests that are running, these two platforms should be running at the same time and, if they do, the newly built images will be available by the time the QA deployment goes looking for the latest images.
> 
> So if I understand this correctly, the build-and-publish GA (step A) take about 10 minutes to run to completion, meanwhile the Unit tests + E2E tests (step B) take 30 minutes to complete. Step A and B run in parallel, and B always takes longer to complete than A, from our experience. Only after the tests are complete and passing is the deployment step on TC (step C) activated which deploys the code to our staging server. Step C cannot start until step B has finished. C also has a dependency on the existence of A, and we feel comfortable with this arrangement even though A and C are not synchronized.
> 
> One thing not mentioned is that the tests, step B, also depend upon the existence of A. I would guess that in this scenario the unit tests/e2e tests would actually be run using a previous version of `sillsdev/web-languageforge:app-base-staging` which would not likely be updated until step A is finished. I'm documenting this mostly so we can agree on the state of things, but what you have going is certainly a great step forward and a great improvement! In our previous deployment, we actually didn't run any tests for staging or production. Our previous practice was to only run tests on the master branch.
> 
> I'll also note that with this change we no longer have a way to deploy to app-base-latest for TC/locally run code. This _could_ be an issue in the future, but I'm unsure how often we would actually encounter this scenario. A possible workaround is to build the app-base-latest locally, tag it, and then point the local Dockerfile to the updated image.
> 
> Requested change:
> 
> * put the Browser Stack credit back in the readme (we still use it).
> 
> Thanks again Billy! This is great work that we can merged in straight away!

It's probably best to take a look at the stuff I put in the README because my original comment here was a desired state but I've decided to go with the cruder manual deployment process due to challenges in TC.

On a very important point you make, yes the tests are asking for `app` which will be built off the `base` which has not yet been built.  If there had been a change in the base, you're absolutely right, the tests would be going off the wrong batch of code, fantastic catch!  However, unless we get GHA/TC in sync I'm not sure of a quick and easy way to do this, coupled with the fact that base should not be changing much, I'd like to assume we won't hit this little flaw and I'd like to assume that whatever code was put in place during one of these situations will have a test with it that would fail if it didn't get the changes that would've been part of the "new" base.  I think dancing in two CI/CD worlds is leaving us open to risks...maybe we should rethink sticking with one.

PR #968, opened 2021-05-28
Disable Protractor Control Flow for Stability

megahirt:
According to 
- https://www.protractortest.org/#/control-flow
- https://www.protractortest.org/#/async-await

We should not be using the control flow since it can cause stability with tests writting usingasync/await (that's us).  I thought we already had this disabled...hmm.

Maybe this will improve stability in the tests?

From the docs:
> Don’t forget to turn off control_flow, you cannot use a mix of async/await and the control flow: async/await causes the control flow to become unreliable (see [github issue](https://github.com/SeleniumHQ/selenium/issues/3037)). So if you async/await anywhere in a spec, you should use the SELENIUM_PROMISE_MANAGER: false

### DIFF
--- a/test/app/protractorConf.js
+++ b/test/app/protractorConf.js
@@ -5,6 +5,7 @@ var specs = [
   "/data/test/app/languageforge/**/" + specString + ".e2e-spec.js",
 ];
 exports.config = {
+  SELENIUM_PROMISE_MANAGER: false,
   seleniumAddress: 'http://selenium:4444/wd/hub',
   baseUrl: 'http://app-for-e2e',
   allScriptsTimeout: 12000,


rmunn (approved):
:rocket: LGTM :rocket:

rmunn:
Kicking off E2E tests again to verify whether the failures are consistent or not.

rmunn:
The build now has a *lot* of failed E2E tests... but they all *start* with the old flaky "with a cleared name (...)" test. I'll rebase this on top of `master`, which now has the fix for that test merged in, and see what happens.

megahirt:
So after some discussions, @rmunn and I decided that our E2E tests are not in good enough shape to turn off the protractor control flow and so I'll close this PR without merging.  It was educational to see how many E2E tests fail, but we've decided not to pursue this at this point.

PR #969, opened 2021-05-31
Use ld-api image from Docker Hub

rmunn:
Now we no longer need to build the ld-api image locally, saving time.


megahirt (approved):
yeah!  This will really speed up the build.  Thank you.

PR #970, opened 2021-05-31
Select E2E tests with TEST_SPECS env var

rmunn:
Passing a `TEST_SPECS` variable to Make, either via an environment variable (`export TEST_SPECS=foo`) or as `make TEST_SPECS=foo e2e-tests` on the command line, will now select only that spec file to run. E.g.:

```bash
make TEST_SPECS=test/app/languageforge/lexicon/lexicon-new-project.e2e-spec.js e2e-tests
```

Note that the `TEST_SPECS` value should end in .js, not .ts, and should be relative to the root of the repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,25 @@ Other useful resources:
 1. `make e2e-tests`
 1. Individual test results will appear in your terminal but if you'd like to watch them in real-time, simply VNC into the running tests via `localhost:5900`, e.g., Mac OSX users simply `open vnc://localhost:5900` and use `secret` as the password.  Other operating systems may require installing a separate VNC Viewer tool.
 
+To run a single E2E spec file, put its path (relative to the repo root) into the `TEST_SPECS` environment variable (don't forget to `export` it), or pass it as an option to `make e2e-tests` as follows:
+
+```bash
+make TEST_SPECS=test/app/languageforge/lexicon/lexicon-new-project.e2e-spec.js e2e-tests
+# Or:
+export TEST_SPECS=test/app/languageforge/lexicon/lexicon-new-project.e2e-spec.js
+make e2e-tests
+```
+
+**Important:** the `TEST_SPECS` file must end in `.js`, not `.ts`, because the test runner we're using doesn't understand Typescript.
+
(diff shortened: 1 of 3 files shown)


rmunn:
The value this provides is the ability to iterate over one specific file when trying to fix a flaky E2E test, so that the tests take far less time to run and you can iterate faster. To get the filename easily, in VS Code you can right-click the tab of the spec file you want to run, then choose "Copy Relative Path" from the dropdown menu. Then in a command line, type:

1. `export TEST_SPECS=`
2. Ctrl+V
3. Backspace over `.ts` and change it to `.js`
4. Enter
5. `make e2e-tests`

And now you're running only the tests in that one file. To go back to running all the tests, just run `unset TEST_SPECS` in that command prompt and then `make e2e-tests` will run all the E2E tests again.

megahirt (changes requested):
This is really great Robin!  I had actually come up with a similar unfinished solution using "suites" instead of specs, but yours is simpler and very usable.

One requested change: can you add the instructions from this PR into the Readme.md under the appropriate section?  These instructions are really valuable.

rmunn:
@megahirt - Done, ready to be approved & merged now.

README.md
```diff
@@ -93,6 +93,25 @@ Other useful resources:
...
+**Important:** the `TEST_SPECS` file must end in `.js`, not `.ts`, because the test runner we're using doesn't understand Typescript.
+
+The easiest way to get the `TEST_SPECS` variable set up correctly is to go into VS Code and right-click the tab containing the spec file you want to run, then choose "Copy Relative Path" from the dropdown menu. Then do the following at the command line:
+
+1. `export TEST_SPECS=`
+1. Ctrl+V (or possibly Ctrl+Shift+V on a Linux command line)
+1. Backspace over `.ts` and change it to `.js`
+1. Enter
+1. `make e2e-tests`
+
```
megahirt, line 114: I'll add the additional steps we talked about to quickly run the tests again without rebuilding.

```suggestion

To quickly re-run the tests without going through the `make build` process, you can restart the `app-for-e2e` container and run the tests as follows:
`docker-compose restart app-for-e2e && docker-compose run -e TEST_SPECS= test-e2e` where the relative path to the test spec file is optionally given after the `=` sign.
```

megahirt (approved):
Approved!  Thank you for including the README instructions.  I've added a small addition as well.

PR #971, opened 2021-05-31
Fix flaky E2E test in Lexicon New Project suite

rmunn:
I discovered that Selenium's `inputElement.clear()` function [does not actually trigger Angular change detection](https://stackoverflow.com/questions/50677760/selenium-clear-command-doesnt-clear-the-element)! Which explains why this particular test has been so flaky for so long. With this new approach, the test should be more consistent.

### DIFF
--- a/test/app/languageforge/lexicon/lexicon-new-project.e2e-spec.ts
+++ b/test/app/languageforge/lexicon/lexicon-new-project.e2e-spec.ts
@@ -212,18 +212,10 @@ describe('Lexicon E2E New Project wizard app', () => {
     });
 
     it('with a cleared name does not show an error but is still invalid', async () => {
-
-      /**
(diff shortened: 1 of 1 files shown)


rmunn:
E2E test run failed first time with four completely unrelated tests failing, so I don't know what's up with that. Second E2E test run was all green. The flaky test I was trying to fix worked both times, so I'm reasonably confident at this point. I'll still re-run the E2E tests about five more times just to make sure.

rmunn:
While some *other* tests failed on two of the ten runs I did, with failures that look like race conditions of some kind, the formerly-flaky E2E test just passed ten times in a row. I'm confident that I've solved the underlying issue and that this test is going to stop being flaky.

rmunn:
Self-merging since everyone who could review this is on vacation or otherwise unavailable, and this should provide stability to the other PRs awaiting reviews.

PR #974, opened 2021-06-09
Establish a persistent volume for project assets

longrunningprocess:


docker/deployment/app-deployment.yaml
```diff
@@ -57,10 +70,19 @@ spec:
       labels:
         app: app
     spec:
+      securityContext: # see https://github.com/sillsdev/web-languageforge/pull/974 for further explanation of the need for a securityContext here
+        fsGroup: 33 # this is www-data from /etc/passwd, k8s requires explicit assignment otherwise the volume will get mounted with root and www-data won't be allowed to create folders on-the-fly in there.
```
rmunn, line 74: I think we should also use `fsGroupChangePolicy: "OnRootMismatch"` (details [here](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#configure-volume-permission-and-ownership-change-policy-for-pods)) so that the `chown` operation will only happen once. The `assets` folder in Language Forge typically gets LOTS of files uploaded into it, because most people who use the audio feature upload one file per entry, so a single project could have over five thousand audio files. Multiply that by the number of projects, and that's a LOT of `chown` operations to run recursively every time the volume is mounted. But by using `fsGroupChangePolicy: "OnRootMismatch"` then we won't have to run the operation every time.

Note that the Kubernetes docs say this is available starting in Kubernetes version 1.20. Our Rancher server in Dallas is running v1.20.4 (verified by running `kubectl version`), and any commercial Kubernetes service we may choose to use will almost certainly have at least v1.20, so I believe this will work.

docker/deployment/app-deployment.yaml
```diff
@@ -57,10 +70,19 @@ spec:
...
+      volumes:
+      - name: assets
+        persistentVolumeClaim:
+          claimName: lf-project-assets
       containers:
       - name: app
         image: sillsdev/web-languageforge:app-staging
         imagePullPolicy: Always
```
rmunn, line 82: Worth considering adding a `securityContext` here with `runAsGroup` matching the `fsGroup` from the pod's security context. That's the default group for the `www-user` account, but at the Kubernetes level where named user & group IDs are not available, only numeric IDs, it's worth making sure that the configuration matches between the volume mount and the container that will be using the volume.

Important to note that the standard UID and GID for the `www-data` user and group are 33 in Debian-based systems, but 82 in Alpine-based systems. So we can't just blithely assume that `www-data`'s GID will always be 33, we need to document it or configure it explicitly. Otherwise if we start using Alpine-based containers mixed with Debian-based containers we could be in for a surprise later on when things don't work.

rmunn (approved):
One change requested, but I'm marking the review as "Approve" so that you're not stuck waiting for me to re-approve after you make the change.

docker/deployment/app-deployment.yaml
longrunningprocess: this really helpful, I agree with you, it's probably a good approach to take anytime we have to do this...I wish it were the default actually.  I'll commit a change for this before merging

docker/deployment/app-deployment.yaml
longrunningprocess: > Important to note that the standard UID and GID for the www-data user and group are 33 in Debian-based systems, but 82 in Alpine-based systems. So we can't just blithely assume that www-data's GID will always be 33, we need to document it or configure it explicitly. Otherwise if we start using Alpine-based containers mixed with Debian-based containers we could be in for a surprise later on when things don't work.

Excellent point, too bad we have to introduce this kind of maintenance, it'll def make these more brittle unfortunately.  I think I'll make a reference in the Dockerfile's where it's most likely to be noticed.

megahirt: > I actually played with those during my testing and it caused problems, the container would not start.

Was there any error output besides "the container would not start"?  It would be helpful to add documentation to this thread if either of you are able to get farther in this problem.  Getting `runAsGroup` working seems like an important component to make this less brittle.  So a question: can we get by with just `runAsGroup` and remove the `fsGRoup` all together?  If so, this would get around the GID-specific issues @rmunn identified between Alpine/Debian

longrunningprocess: yeah, there was a little output but it was ambiguous, something about not starting some IP address or something...it didn't seem helpful to me at the time and I was more focused on simply trying things that would work rather than going down another rabbit trail at the moment.

All of those options still require a gid rather than a name so that brittleness will still be there but I was having some thoughts about another approach where we could chown the name, i.e., www-data, in an init container which would protect us from the gid maintenance.

PR #979, opened 2021-06-16
Establish a persistent volume for lfmerge S/R process

longrunningprocess:


docker/app/Dockerfile
```diff
@@ -22,8 +22,6 @@ RUN ln -s /var/www/html /var/www/src
 COPY --from=lf-ui-builder /data/src/dist /var/www/html/dist
 
 # ensure correct write permissions for assets folders, 
-# WARNING:  k8s requires an explicit configuration of the ownership via a gid and on deb it's 33 for www-data but it's 82 on alpine so be very careful 
-# see app-deployment.yaml for "fsGroup" usage
 RUN    chown -R www-data:www-data /var/www/html/assets /var/www/html/cache \
     && chmod -R g+ws /var/www/html/assets /var/www/html/cache \
     && umask 0002
```
rmunn, line 27: This `umask 0002` line won't do anything. The `umask` value is set per-process (and inherited by child processes), so if you set it in a Dockerfile then it will be set in the process (and container) that runs during Dockerfile build, but that container is then discarded. The umask setting will not be saved in any way in the *filesystem image* that `docker build` produces. If we need to set a umask for our processes, it has to be set in the entrypoint script. Then it will be inherited by child processes.

However, if we make sure to run everything as `www-data` instead of as `root`, then we won't need the umask, so the fact that this line is useless won't hurt us.

docker/deployment/app-deployment.yaml
```diff
@@ -100,22 +100,28 @@ spec:
...
       - name: assets
         persistentVolumeClaim:
           claimName: lf-project-assets
       - name: sendreceive-data
         persistentVolumeClaim:
           claimName: lfmerge-sendreceive-data
+      initContainers:
+      - name: initialize-volume-ownership
+        image: busybox:1.33
+        command: ['sh', '-c', 'mkdir -p /var/www/html/assets && chown www-data:www-data /var/www/html/assets && mkdir -p /var/lib/languageforge/lexicon/sendreceive && chown www-data:www-data /var/lib/languageforge/lexicon/sendreceive']
```
rmunn: I double-checked the `busybox:1.33` image and it does have UID and GID 33 assigned for the user and group named `www-data`. So doing this will assign UID 33 and GID 33, which is what we want.

docker/app/Dockerfile
rmunn: I verified this by going into the currently-running `app` container on Rancher QA, and the umask there was still 0022, despite this Dockerfile line:

image

Also, `/var/www/html/assets` was not group-writable, so this `chmod` is having no effect on the Kubernetes deployment. The changes made to /var/www/html/assets inside the Dockerfile are only affecting the files in the Docker image. When you mount a Kubernetes volume over the top of the Dockerfile, Kubrnetes uses the permissions of the volume root, not the permissions of the mount point from the Docker image, and the volume root being mounted into `/var/www/html/assets` has not had `chmod g+ws` applied to it.

docker/deployment/app-deployment.yaml
```diff
@@ -110,8 +110,7 @@ spec:
       initContainers:
       - name: initialize-volume-ownership
         image: busybox:1.33
-        command: ['sh', '-c']
-        args: ['mkdir -p /var/www/html/assets /var/lib/languageforge/lexicon/sendreceive && chown www-data:www-data /var/www/html/assets /var/lib/languageforge/lexicon/sendreceive']
+        command: ['sh', '-c', 'chown www-data:www-data /var/www/html/assets /var/lib/languageforge/lexicon/sendreceive']
```
rmunn, line 113: I double-checked the `busybox:1.33` image and it does have UID and GID 33 assigned for the user and group named `www-data`. So doing this will assign UID 33 and GID 33, which is what we want.

docker/app/Dockerfile
longrunningprocess: I noticed the umask thing as well but wanted to discuss why Chris put in place...but I think with both of us confirming this seems to be useless, I'm inclined to remove it altogether.

docker/deployment/app-deployment.yaml
longrunningprocess: Thanks for the second set of eyes...this unfortunately is also vulnerable to the underlying image then I take it, e.g., if we went to alpine, would you expect this to break because of the id's being 82 there?

megahirt:
sounds like any permissions related settings (umask, chmod, etc) need to be
done at container run-time (umask) or baked into the volume mount settings
(chmod).  I'm pretty sure I adapted these settings from the ansible script
without thinking too much about image build vs. container runtime
implications.  Thanks for clarifying and catching the differences.

docker/app/Dockerfile
longrunningprocess: well, I'm sorry to say that removing the umask line breaks the lfmerge (receive at least) functionality locally.

longrunningprocess: Additionally, I tried running app as `www-data` and ran into issues with unit and e2e tests as well as broken lfmerge functionality.  It's probably best to either pair on this effort or have you and/or @megahirt tackle this one.

At first glance, the unit test failures seem to be benign but and the e2e failures might just need a little research to get `www-data` in `app-for-e2e` as well but I'd just be banging my head against the wall a bit trying to get through all these things, it'll be much faster to work closer with you all on this one.

PR #980, opened 2021-06-21
Use WEBSITE env var to get base URL

rmunn:
To fix the OAuth login issue introduced by #955, we need to actually use the `WEBSITE` environment variable that is being passed in. This allows qa.languageforge.org to create redirection URLs after OAuth login that return to the correct site instead of having the QA site redirect to the live site.

### DIFF
--- a/src/Site/OAuth/OAuthBase.php
+++ b/src/Site/OAuth/OAuthBase.php
@@ -3,6 +3,7 @@
 namespace Site\OAuth;
 
 use Api\Library\Shared\Website;
+use Api\Library\Shared\UrlHelper;
 use Api\Model\Shared\UserModel;
 use League\OAuth2\Client\Provider\AbstractProvider;
 use League\OAuth2\Client\Token\AccessToken as OAuthAccessToken;
@@ -36,8 +37,7 @@ abstract class OAuthBase extends Base
 
     public function oauthCallback(Request $request, Application $app)
     {
-        $website = Website::get();
-        $redirectUri = $website->baseUrl() . '/oauthcallback/'. $this->getProviderName();
+        $redirectUri = UrlHelper::baseUrl() . '/oauthcallback/'. $this->getProviderName();
         $provider = $this->getOAuthProvider($redirectUri);
 
         $error = $request->query->get('error', null);
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -28,7 +28,7 @@ ifeq ("$(TEAMCITY_VERSION)",$())
 	docker-compose run test-php
 else
 	# teamcity CI
-	docker-compose run --name unittests test-php
+	docker-compose run -e WEBSITE=languageforge.org --name unittests test-php
 	docker cp unittests:/var/www/PhpUnitTests.xml ..
 endif
 
--- a/src/Api/Library/Shared/Communicate/Communicate.php
+++ b/src/Api/Library/Shared/Communicate/Communicate.php
@@ -5,6 +5,7 @@ namespace Api\Library\Shared\Communicate;
 use Api\Library\Shared\Communicate\Sms\SmsModel;
 use Api\Library\Shared\Communicate\Sms\SmsQueue;
 use Api\Library\Shared\Website;
+use Api\Library\Shared\UrlHelper;
 use Api\Model\Shared\ProjectModel;
 use Api\Model\Shared\ProjectSettingsModel;
 use Api\Model\Shared\MessageModel;
@@ -110,7 +111,7 @@ class Communicate
         $subject = $website->name . ' account signup validation';
         $vars = [
                 'user' => $userModel,
-                'link' => $website->baseUrl() . '/validate/' . $userModel->validationKey,
+                'link' => UrlHelper::baseUrl() . '/validate/' . $userModel->validationKey,
                 'website' => $website,
         ];
 
@@ -128,7 +129,7 @@ class Communicate
         $subject = 'Welcome to ' . $website->name;
         $vars = [
             'user' => $userModel,
-            'link' => $website->baseUrl(),
+            'link' => UrlHelper::baseUrl(),
             'website' => $website,
         ];
 
@@ -218,7 +219,7 @@ class Communicate
         $subject = $website->name . ' Forgotten Password Verification';
         $vars = [
             'user' => $user,
-            'link' => $website->baseUrl() . '/auth/reset_password/' . $user->resetPasswordKey,
+            'link' => UrlHelper::baseUrl() . '/auth/reset_password/' . $user->resetPasswordKey,
             'website' => $website,
         ];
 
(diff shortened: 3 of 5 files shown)


longrunningprocess:
This looks like the right approach, thank you @rmunn .  I'm hoping a simple rebase off `master` will address the unit test failure for you.

megahirt:
@rmunn This is an interesting solution that might just work, and the E2E tests passing give confidence that it's probably fine.  Probably the reason you're asking for second eyes is that your PR changes the *intent* of the get() method.  It kind of turns it inside-out, where as the original intent was to "get the website instance of the supplied $hostname parameter, and if not supplied get it from the environment" to "get the Website instance from the environment, and if not defined get it from the parameter".  The LF code as written uses Website::get($hostname) as a contract to always return the website instance referred to by $hostname.  Your change breaks this contract, so I'd need to look through all of the instances of Website::get($hostname) to see how your change would affect those cases.  Cases of Website::get() should be unaffected.

I actually [have a branch ](https://github.com/sillsdev/web-languageforge/tree/chore/removeWebsiteClass) where I have attempted to remove all traces of the Website class, but it probably requires a few more hours of work to get it PR ready.

The things that the Website class provides for QA vs PROD is the domain name string switching in places like the email template.  We could fairly easily remove use of the class in these few cases in favor of an if/else statement based upon the WEBSITE environment variable.

One of the goals of #955 was to make the underlying website "think" that it is always languageforge.org for the purposes of operation.  This makes operation of the site simpler as a whole so that there are no longer data distinctions between localhost, e2e, QA and PROD instances.  It also solves the problem of "migrating" projects in the db from PROD to QA or to a developer environment which has been a chore in the past (we had scripts to do this, but still).

What you figured out and what I neglected to solve in #955 are the few places where URLs (email text, OAUTH callbacks, etc) have to be customized between QA and PROD.  For these few cases I would suggest that we simply modify the logic directly in those methods, which is what I have done in my unfinished branch removing Website class altogether.

Since I'm still mostly on vacation, I'm happy for this PR to be merged as is if it solves your problem and tests are passing.  I have mostly kicked the can down the road a bit on getting rid of the Website class (not a small task) and this PR continues to kick the can a bit further, which I'm fine with.  I mostly wanted to explain where my other effort (removing Website class) was going, and if you wanted to align with that, then I would suggest making the change in the "send email" method and the "oauth callback method" that have a conditional based directly on the WEBSITE env.

rmunn:
The "send email" methods had a lot of passing `$website` into the functions and using it, and I decided that the simplest way to make that change would be to change the `$website` instance at the point where it was created, which turned out to be at the top of Sf.php. By using the `WEBSITE` environment variable, we'll ensure that the QA instance of the Website class is used, rather than the live instance. The only place where that matters is in the BugsnagExceptionHandler, where now the release stage will be set to "qa" instead of "live" when a bug comes from qa.languageforge.org — which I think is what we want.

The OAuth class uses a different approach, and isn't called directly from Sf.php and handed a `$website` instance, so it needed its own fix.

src/Site/OAuth/OAuthBase.php
```diff
@@ -37,7 +38,10 @@ abstract protected function getOAuthProvider($redirectUri): AbstractProvider;
     public function oauthCallback(Request $request, Application $app)
     {
         $website = Website::get();
-        $redirectUri = $website->baseUrl() . '/oauthcallback/'. $this->getProviderName();
+        // TODO: Can we assume https here and get rid of the Website::get() call entirely? Or are there times when we need http here?
```
megahirt, line 41: yes, we can assume HTTPS.  Safe to delete this comment.

src/Site/OAuth/OAuthBase.php
```diff
@@ -37,7 +38,10 @@ abstract protected function getOAuthProvider($redirectUri): AbstractProvider;
     public function oauthCallback(Request $request, Application $app)
     {
         $website = Website::get();
```
megahirt, line 40: Delete this line since it's not needed anymore.
```suggestion
```

src/Site/OAuth/OAuthBase.php
```diff
@@ -37,7 +38,10 @@ abstract protected function getOAuthProvider($redirectUri): AbstractProvider;
...
-        $redirectUri = $website->baseUrl() . '/oauthcallback/'. $this->getProviderName();
+        // TODO: Can we assume https here and get rid of the Website::get() call entirely? Or are there times when we need http here?
+        $protocol = ($website->ssl) ? "https" : "http";
+        $baseUrl = $protocol . "://" . Env::requireEnv('WEBSITE');
```
megahirt, line 43:
```suggestion
        $baseUrl = "https://" . Env::requireEnv('WEBSITE');
```

src/Site/OAuth/OAuthBase.php
```diff
@@ -37,7 +38,10 @@ abstract protected function getOAuthProvider($redirectUri): AbstractProvider;
     public function oauthCallback(Request $request, Application $app)
     {
         $website = Website::get();
-        $redirectUri = $website->baseUrl() . '/oauthcallback/'. $this->getProviderName();
+        // TODO: Can we assume https here and get rid of the Website::get() call entirely? Or are there times when we need http here?
+        $protocol = ($website->ssl) ? "https" : "http";
```
megahirt, line 42: Delete this line since we hardcode to HTTPS
```suggestion
```

megahirt (changes requested):
This is good - simple changes that should work.  Just a bit of cleanup in the oauth url construction.

src/Site/OAuth/OAuthBase.php
rmunn: Done in latest commit.

src/Site/OAuth/OAuthBase.php
rmunn: Done in latest commit.

src/Site/OAuth/OAuthBase.php
rmunn: Done in latest commit.

src/Site/OAuth/OAuthBase.php
rmunn: Done in latest commit.

rmunn:
Rebasing on top of latest `master` branch commit.

megahirt (approved):

rmunn:
Latest force push is to drop the five commits that ended up being abandoned, so now there are only two commits. The first is the minimal-change commit that only changes how `$website->baseUrl()` looks for the hostname. The second commit is the more extensive change that replaces `$website->baseUrl()` with a static method in a `UrlHelper` class. If we go with the first commit, there will be no merge conflicts once [the chore/removeWebsiteClass branch](https://github.com/sillsdev/web-languageforge/tree/chore/removeWebsiteClass) is merged, but I think we actually *want* there to be merge conflicts with that branch, otherwise all those URLs will end up being hardcoded to `languageforge.org` when we want some of them to end up as `qa.languageforge.org`. The second commit will cause merge conflicts in those hardcoded locations, which should be resolved in favor of the `UrlHelper::baseUrl()` calls.

megahirt (approved):
This looks complete - thank you!

PR #984, opened 2021-06-22
Allow debugging E2E tests

rmunn:
This setup uses VS Code's Remote Containers feature to run VS Code inside the test-e2e container, where it can directly attach to Node processes and let you set breakpoints on the E2E tests.


longrunningprocess (approved):
excellent work, very thorough!

Issue #981, opened 2021-06-21
Send/Receive functionality broken in staging environment


longrunningprocess:
I noticed during work on https://github.com/sillsdev/web-languageforge/pull/979 that "receive" seemed to work just fine locally but not in staging.  @rmunn is already in the midst of some LFMerge work that might or might not have something to do with this but it seems this use case may have been broken all along, if not in the "receive" portion, definitely in the "send" portion due to missing `ping` in the app container.  I've added `ping` to our base image though so hopefully that will help this functionality along.

longrunningprocess:
team discussion on the matter so far:

> The reason why Send/Receive is failing on QA right now is because the /var/lib/languageforge/lexicon/sendreceive/state directory doesn't exist. LfMerge expects that directory to exist, and doesn't have any code to create the directory if it doesn't exist. Perhaps it should have such code, but that's the issue right now: that directory needs to be created (and owned by www-data like the others) during the app setup process.

error from `/var/log/syslog`: `Exception:#012System.IO.DirectoryNotFoundException: Could not find a part of the path '/var/lib/languageforge/lexicon/sendreceive/state'`

seems like a difference between volumes in docker/k8s...will confirm this today.

longrunningprocess:
# local testing

## without the volume configs
I temporarily commented out the sendreceive volume: `lfmerge-sendreceive-data:/var/lib/languageforge/lexicon/sendreceive`

and see the following (as expected):
```
dc exec app ls -l /var/lib/languageforge/lexicon/sendreceive
total 36
drwxrwsr-x 1 www-data www-data 4096 Jun 21 14:47 ReleaseData
drwxrwsr-x 1 www-data www-data 4096 Jun 10 13:58 Templates
drwxrwsr-x 1 www-data www-data 4096 Jun 10 13:58 commitqueue
drwxrwsr-x 1 www-data www-data 4096 Jun 10 13:58 mergequeue
drwxrwsr-x 1 www-data www-data 4096 Jun 10 13:58 receivequeue
drwxrwsr-x 1 www-data www-data 4096 Jun 10 13:58 sendqueue
drwxrwsr-x 1 www-data www-data 4096 Jun 10 13:58 state
drwxrwsr-x 1 www-data www-data 4096 Jun 21 14:47 syncqueue
drwxrwsr-x 1 www-data www-data 4096 Jun 10 13:58 webwork
```

## with the volume configs
after uncommenting the volume, here's what I see (as expected):
```
[...]

PR #985, opened 2021-06-22
Fixed S/R bug in staging (#981)

longrunningprocess:
due to diff's in k8s/docker volumes, S/R was not working in staging ultimately because LFMerge's directory setup (https://github.com/sillsdev/LfMerge/blob/master/debian/postinst#L24) was being blown away during a volume mount in staging, see discussion in #981

### DIFF
--- a/docker/deployment/app-deployment.yaml
+++ b/docker/deployment/app-deployment.yaml
@@ -110,7 +110,12 @@ spec:
       initContainers:
       - name: initialize-volume-ownership
         image: busybox:1.33
-        command: ['sh', '-c', 'chown www-data:www-data /var/www/html/assets /var/lib/languageforge/lexicon/sendreceive']
+        command:
+          - 'sh'
+          - '-c'
+          - |-
+            mkdir /var/lib/languageforge/lexicon/sendreceive/state
+            chown www-data:www-data /var/www/html/assets /var/lib/languageforge/lexicon/sendreceive /var/lib/languageforge/lexicon/sendreceive/state
         volumeMounts:
           - mountPath: /var/www/html/assets
             name: assets


longrunningprocess:
this is a narrow, tactical fix that allows users in staging to perform S/R properly again but there are more dirs lfmerge expects to be in place but it's unclear atm which ones get built on the fly and which ones will cause issues like `state` did, so I'd prefer to correct those as they occur unless someone already knows the use cases that have an effect on `mergequeue`, `commitqueue`, `receivequeue`, `sendqueue`, `ReleaseData` and `Templates`.

docker/deployment/app-deployment.yaml
```diff
@@ -110,7 +110,12 @@ spec:
       initContainers:
       - name: initialize-volume-ownership
         image: busybox:1.33
-        command: ['sh', '-c', 'chown www-data:www-data /var/www/html/assets /var/lib/languageforge/lexicon/sendreceive']
+        command:
+          - 'sh'
+          - '-c'
+          - |-
+            mkdir /var/lib/languageforge/lexicon/sendreceive/state
+            chown www-data:www-data /var/www/html/assets /var/lib/languageforge/lexicon/sendreceive /var/lib/languageforge/lexicon/sendreceive/state
```
rmunn, line 118: This isn't going to run two commands. The newline character is *not* a separator when running `sh -c`. This is going to try to create the following directories:

- `/var/lib/languageforge/lexicon/sendreceive/state`
- `chown`
- `www-data:www-data`
- and so on

You need a semicolon in between these arguments.

longrunningprocess: yeah, that makes perfect sense, I need to double-check my tests and see if I missed something or this special syntax works differently for some reason, thanks for pointing it out.

longrunningprocess: I double-checked, this syntax works fine and treats them as separate commands already, nothing to change here.

PR #988, opened 2021-06-29
Clean up of old, Ansible stuff

longrunningprocess:
I don't believe any of this is needed anymore.


megahirt (commented):
Thanks for this.  I'm fine deleting these to tidy up our repo, especially since I can easily look at them with git history (if I ever need to - probably not).

Removing submodules is hairy.  Is git happy after removing the `deploy` directory?  When I removed a submodule a few months ago there were separate developer steps needed in addition to this change, and we should put those instructions in the PR description for developers, and also post to Slack when this PR is merged.

See https://stackoverflow.com/questions/1260748/how-do-i-remove-a-submodule/21211232#21211232

I believe the correct incantation for local devs is:
`sudo rm -r .git/modules/deploy`
but would want you to confirm that.

Of course this change cannot be checked-in / tracked because it's part of the git repo itself :)

longrunningprocess:
I'll have to check that @megahirt , I was seeing some strange behavior in GitKraken when I tried checking out a different branch so I thought it was just a bug or something.  When I encountered that, I simply worked in the command line and wasn't seeing issues there...maybe I need to do some of this stuff you're talking about until GitKraken works again.

longrunningprocess:
image

looks like they could probably be removed from GitKraken pretty easily and I see another command referenced in the SO link above about `git rm`...

longrunningprocess:
I did see some remnants int he config file so if that's the case for you too, you can do the following to get rid of it:
```sh
grep submodule .git/config

[submodule "deploy/ops-ansible-common-roles"]
[submodule "deploy/roles/ansible-role-apache"]
```

to see which ones are still listed

and then 
```sh
git config -f .git/config --remove-section submodule.deploy/ops-ansible-common-roles
git config -f .git/config --remove-section submodule.deploy/roles/ansible-role-apache
```

to clean it up finally.

megahirt (approved):
LGTM, thank you Billy!

PR #991, opened 2021-07-01
Extract common Webpack config from dev and prd

rmunn:
The dev and prd configs shared almost all of their values, so we can extract those parts of the config to the common config file.

Verified locally by re-running build and running E2E tests; Webpack config still works as it should.


longrunningprocess (approved):
This is a nice clean up, thanks!

PR #994, opened 2021-07-07
Establish regular backups for our PVC's

longrunningprocess:


docker/deployment/db-deployment.yaml
```diff
@@ -23,8 +23,37 @@ spec:
...
+      template:
+        spec:
+          containers:
+          - name: snapshot-runner
+            image: docker.dallas.languagetechnology.org/ba-kb-snapshot
+            command:
+            - /bin/sh
+            - -c
+            - |- 
+              cat <<EOF | kubectl apply -f -
```
megahirt, line 47: Wow, a snapshot container that does the backup, run from a `kubectl` run from a shell script, run from a cronjob defined in a kubernetes YAML.  This is complex!

longrunningprocess: yes, I agree...too complex imo but this is a POC and things like this are what it's all about.  The syntax is pretty gross but if the only thing teams needed to change were the names, we could parameterize this in a way that teams could just send in those two custom names and the underlying container would hide all this ugliness.

PR #996, opened 2021-07-08
add shell scripts to document kubectl SSH tunnel

megahirt:
These shell scripts are intended to be documentation for how I setup a SSH tunnel to both the APP and Mongo containers for the purposes of using rsync to sync files between US containers (as opposed to bouncing files off Thailand between containers).

The process was educational and it was fun to see working.

These can probably be discarded after we complete our migration tasks and have a new production environment.  They honestly don't belong as part of any working solution, since they were only written as one-off scripts that would only work from my machine.

The rough process is:

1) run kubeSetup.sh to install ssh and rsync on the target container
2) run kubePortForward.sh to open a port-forward for the SSH tunnel (do this in a separate terminal)
3) run kubeRsync*.sh to setup an SSH tunnel with agent forwarding and rsync the path on production to the equivalent path on staging.
4) depending upon whether it is night or day, you either go to lunch or go to sleep.  Check back later to see if the transfer completed.


longrunningprocess:
This is pretty amazing dude!

docker/deployment/scripts/kubeRsyncAssets.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Documentation:
+# -A  agent forwarding
+# -p 55555 this port is being forwarded to 22 on the container
+# -o "UserKnownHostsFile=/dev/null" -o "StrictHostKeyChecking=no" these are a hack to bypass security checks for "localhost".  localhost isn't a real host so I don't care about "host key verification failed" errors when doing localhost port forwarding
+
+ssh -A -p 55555 -o "UserKnownHostsFile=/dev/null" -o "StrictHostKeyChecking=no" root@localhost "rsync -avzhP chirt@sysops.languageforge.org:/var/www/languageforge.org/htdocs/assets/lexicon /var/www/html/assets/lexicon"
```
longrunningprocess, line 8:
```suggestion
ssh -A -p 55555 -o "UserKnownHostsFile=/dev/null" -o "StrictHostKeyChecking=no" root@localhost "rsync -avzhP chirt@sysops.languageforge.org:/var/www/languageforge.org/htdocs/assets/lexicon /var/www/html/assets/"
```

longrunningprocess: Looks like it's off by one dir, here's from the container:

```sh
root@app-7c76db448d-cmxlw:/var/www/html/assets/lexicon# ls -l 
total 40
drwxrwxr-x 734 www-data www-data 32768 Jul  8 07:30 lexicon
drwxr-xr-x   2 www-data www-data  4096 Jul  5 08:24 sf_test-rmunn-food-02
```

docker/deployment/scripts/kubeRsyncSendReceive.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Documentation:
+# -A  agent forwarding
+# -p 55555 this port is being forwarded to 22 on the container
+# -o "UserKnownHostsFile=/dev/null" -o "StrictHostKeyChecking=no" these are a hack to bypass security checks for "localhost".  localhost isn't a real host so I don't care about "host key verification failed" errors when doing localhost port forwarding
+
+ssh -A -p 55555 -o "UserKnownHostsFile=/dev/null" -o "StrictHostKeyChecking=no" root@localhost "rsync -avzhP chirt@sysops.languageforge.org:/var/lib/languageforge/lexicon /var/lib/languageforge/lexicon"
```
longrunningprocess, line 8:
```suggestion
ssh -A -p 55555 -o "UserKnownHostsFile=/dev/null" -o "StrictHostKeyChecking=no" root@localhost "rsync -avzhP chirt@sysops.languageforge.org:/var/lib/languageforge/lexicon /var/lib/languageforge/"
```

longrunningprocess: I think we'll need this here too.

docker/deployment/scripts/kubeSetup.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+MONGO=db-d8fd84c89-vjs7j
+APP=app-7c76db448d-sq5st
```
longrunningprocess, line 3: `kubectl exec -it deploy/app --`

would always work, no need to manually enter the instance name

longrunningprocess:
```suggestion
APP=deploy/app
```

docker/deployment/scripts/kubeRsyncAssets.sh
longrunningprocess: actually, just learned you can simply add the trailing `/` as well.

megahirt: thanks for finding this!  I ended up going with the trailing slash option.

docker/deployment/scripts/kubeRsyncSendReceive.sh
megahirt: yes, thanks!

docker/deployment/scripts/kubeSetup.sh
megahirt: I ended up getting the pod name programmatically since "deploy/all" works for the `exec` command but not the `cp` command.  Go figure.

megahirt: I am now happily exec'ing into a container using your suggest syntax of `kubectl exec -it deploy/app -- bash` so thank you for teaching me that.

longrunningprocess:
Really like these latest changes!

megahirt:
We can go ahead and merge this so I don't need to be a branch to run my scripts :)  We can remove them once the migration is complete.

longrunningprocess (approved):
please see my private Slack message related to this one

PR #997, opened 2021-07-09
Better handling of field input systems overrides

rmunn:
In project config, field input systems are supposed to override user input systems, not the other way around. Previously the logic was incorrectly having user input systems override field input systems.

This fix changes the logic for input systems so that the list of input systems for the field and the list of input systems for the user are ANDed together to get the override list. I.e., if the user has input systems A, B and D selected but C turned off, and the field is supposed to show input systems B, C and D, then the user will see just input systems B and D for that field.

We also get rid of the Reset buttons on the input systems settings. It was insufficiently obvious to users what they did, and with the new logic for calculating field overrides they're not needed any longer.

Demo:

input systems 1 
field input systems 1

Produces:

result 1

If manager turns off French:

input systems 2
field input systems 2

Result (when logged in as manager) is:

result 2

### DIFF
--- a/src/angular-app/languageforge/lexicon/settings/configuration/configuration-fields.component.html
+++ b/src/angular-app/languageforge/lexicon/settings/configuration/configuration-fields.component.html
@@ -112,12 +112,7 @@
                         <i class="fa fa-plus"></i>
                     </button>
                 </th>
-                <td></td>
-                <td></td>
-                <td></td>
-                <td></td>
-                <td></td>
-                <td></td>
+                <td colspan="6"></td>
                 <!--suppress JSUnusedLocalSymbols -->
                 <td data-ng-repeat="group in $ctrl.unifiedViewModel.inputSystems.selectAllColumns.groups"></td>
                 <td></td>
@@ -125,12 +120,7 @@
             <tr style="background-color: initial; height: 2rem">
                 <td></td>
                 <th></th>
-                <td></td>
-                <td></td>
-                <td></td>
-                <td></td>
-                <td></td>
-                <td></td>
+                <td colspan="6"></td>
                 <!--suppress JSUnusedLocalSymbols -->
                 <td data-ng-repeat="group in $ctrl.unifiedViewModel.inputSystems.selectAllColumns.groups"></td>
                 <td></td>
@@ -288,12 +278,7 @@
                         <i class="fa fa-plus"></i>
                     </button>
                 </th>
-                <td></td>
-                <td></td>
-                <td></td>
-                <td></td>
-                <td></td>
-                <td></td>
+                <td colspan="6"></td>
                 <!--suppress JSUnusedLocalSymbols -->
                 <td data-ng-repeat="group in $ctrl.unifiedViewModel.entryFields.selectAllColumns.groups"></td>
                 <td></td>
@@ -303,12 +288,7 @@
                 <th class="text-left align-top">
                     <small>Custom Fields shown in <i>italics</i></small>
                 </th>
-                <td></td>
-                <td></td>
-                <td></td>
-                <td></td>
-                <td></td>
-                <td></td>
+                <td colspan="6"></td>
                 <!--suppress JSUnusedLocalSymbols -->
                 <td data-ng-repeat="group in $ctrl.unifiedViewModel.entryFields.selectAllColumns.groups"></td>
                 <td></td>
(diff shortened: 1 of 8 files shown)


src/angular-app/languageforge/lexicon/settings/configuration/configuration-fields.component.html
```diff
@@ -120,42 +114,12 @@
...
-                    <button type="button" class="btn btn-std" data-ng-click="$ctrl.resetInputSystemsForRole($ctrl.unifiedViewModel.inputSystems.settings, $ctrl.unifiedViewModel.inputSystems.hasCustomInputSystemsOverride, $ctrl.unifiedViewModel.inputSystems.selectAllColumns, $ctrl.fccConfigDirty, 'manager')"
-                            data-ng-show="$ctrl.unifiedViewModel.inputSystems.hasCustomInputSystemsOverride.manager"
-                            id="reset-input-system-manager-btn" title="Reset input systems for manager">
-                        Reset
-                    </button>
-                </td>
+                <td></td>
+                <td></td>
+                <td></td>
+                <td></td>
```
megahirt, line 120: I'm thinking you could get away with a `<td colspan="6"></td>` here as an equivalent expression of table structure.

megahirt (changes requested):
Robin, I'm excited that you have so quickly investigated the user-issue, understood the problem in code and in our data, and proposed a solution while also simplifying the UI.  Great work!

This is a lot of essentially unknown code for me to review, but your PR demonstrates visually that the bug is fixed in your sample data.

Could you perhaps add an E2E regression test that demonstrates this feature works?  Something similar to what you have demonstrated in your screenshot.

Thanks a bunch!

rmunn:
Added E2E regression tests.

The E2E tests turned up one feature of the config view where the code was working as intended, but was causing the E2E tests to fail because the intent of the code was not something I expected. Specifically, when the input system config is populated, we called the `isInputSystemUsed` function and if it returned false, we unchecked that input system in the config. The intent was for projects that had extraneous input systems in their config (left over from a bug in early Language Forge where the Thai input system would be added to projects no matter what language they were intended for), those input systems would be automatically ignored. But in my E2E tests, I added the Thai input system, used only by the Word (lexeme) field. Then when I unchecked the Thai input system for the lexeme field, the code detected that it was no longer being used and automatically unchecked it from the input systems config for all four roles.

Now that the feature is in place where input systems are ANDed together between role config and field config, that feature (of unchecking unused input systems) is no longer desired, so I've removed it. This makes the user experience easier to understand, *and* makes the E2E tests pass.

src/angular-app/languageforge/lexicon/settings/configuration/configuration-fields.component.html
rmunn: Done, and also in all the other places where this exists. I thought about using `colspan="0"` but it seems that feature was *removed* from HTML 5, sadly, due to browsers other than Firefox not implementing it. Apparently `rowspan="0"` is still valid (this cell should extend downwards across all rows), but `colspan="0"` is not allowed as a way of saying "this cell should extend rightwards across all columns). Sigh; that means your HTML has to know how many columns it contains, and it's easy to get it wrong when modifying a table.

There's another possible improvement: everywhere where I just made `colspan="6"`, we also have `<td data-ng-repeat="group in $ctrl.unifiedViewModel.exampleFields.selectAllColumns.groups"></td>`. We could do `<td colspan="$ctrl.unifiedViewModel.exampleFields.selectAllColumns.groups.length">" here. There'd be a problem if there are no groups so the length was 0, but everywhere where this appears, there's a final `<td></td>` so we could use `length + 1` and colspan would always be at least 1. However, I'm thinking this would be a bit too clever for our own good, as later someone might remove the final column of the table and delete the `+1` and then we'd have `colspan="0"` which would cause issues in Chrome. So I'm going to leave well enough alone regarding the `ng-repeat`.

megahirt: great thanks.  Thanks also for the detailed explanation.

megahirt:
> Now that the feature is in place where input systems are ANDed together between role config and field config, that feature (of unchecking unused input systems) is no longer desired, so I've removed it. This makes the user experience easier to understand, _and_ makes the E2E tests pass.

Wow - again, great detective work!

megahirt (approved):
LGTM

PR #999, opened 2021-07-13
Better setup for E2E test container

rmunn:
This ensures that the E2E test container will be ready for debugging as soon as it starts up. (Previously, VS Code could attach to the container before it was actually ready, and it wasn't always clear why you had to wait a minute before pressing the "Launch Debug Session" button).


rmunn:
Self-merging without waiting for E2E tests, as this only affects local dev setups and doesn't affect the E2E tests that TeamCity will run.

PR #1001, opened 2021-07-14
Clean up more project files that may not be needed anymore

longrunningprocess:


longrunningprocess:
The first phase of possible improvements is related to `index.html` files that are scattered around.  I assumed these files were there so someone can't get a directory listing but I can't reproduce something like with or without these files so I'm going to remove them for now.  @megahirt or @rmunn please let me know if having those `index.html` files there were for another vector I'm just not aware of.

longrunningprocess:
I was going to attempt to move the contents of `src/.htaccess` into `docker/000-default.conf` but as I did, there were more `404`'s in the app for additional resources so I had to start adding additional rules to allow the app access to it's resources.  Here's an example of what I started doing in `docker/000-default.conf`:

```
...
  RewriteEngine On

  RewriteCond  %{REQUEST_URI}  "^/dist/"
  RewriteRule  "^/(.*)$"  "/var/www/html/$1"

  RewriteCond  %{REQUEST_URI}  "^/angular-app/"
  RewriteRule  "^/(.*)$"  "/var/www/html/$1"

  RewriteCond %{REQUEST_FILENAME} !-f
  RewriteRule ^ /var/www/html/index.php [QSA,L]
...
```

Either there's a better way to do this or we'll have to add a bunch of rewrite rules to get things right again.  I my opinion, this is not really worth it, I think just having the simple rule in the `src/.htaccess` is better.  What do you think @megahirt or @rmunn ?

longrunningprocess:
as I'm looking at the security around the `cache` folder, are there other directories we'd like this we should ensure we don't allow web access?

longrunningprocess:
There are more listings in `.gitignore` that I don't believe are needed anymore, would you all (@megahirt @rmunn ) mind taking a look as well?  You may either let me know or push up a commit with the removals, thanks.

.gitignore
```diff
@@ -4,18 +4,15 @@
 .externalToolBuilders
 .idea/
 .settings
-docs/semdom/*.xml
-docs/semdom/semdom lists/*.xml
 node_modules/
 output/chrome/
```
megahirt, line 8:
```
output/chrome/
```
I don't think this is necessary.  Items in .gitignore are fairly harmless to delete, because the worst thing that can happen is you check-in something you didn't mean to check-in, which is a pain, but not as bad as breaking the application 😁

.gitignore
```diff
@@ -4,18 +4,15 @@
...
 PhpUnitTests.xml
 phpUnitConsole.txt
 src/angular-app/**/helps/**/page/*.html
```
megahirt, line 11: pretty sure this is left over from an old feature that was removed a while back
```
src/angular-app/**/helps/**/page/*.html
```

docker/app/000-default.conf
```diff
@@ -11,6 +11,10 @@
 	ServerAdmin webmaster@localhost
 	DocumentRoot /var/www/html
 
+	<Directory "/var/www/html/cache">
+			Require all denied
```
megahirt, line 15: cool - thanks for moving this

src/.htaccess
```diff
@@ -1,8 +1,6 @@
 <IfModule mod_rewrite.c>
     Options -MultiViews
 
-    RewriteEngine On
```
megahirt, line 4: I assume `RewriteEngine On` is needed?

megahirt (approved):
Approved, assuming you have tested all this :)

I would have guessed that `RewriteEngine On` line is necessary. but I haven't looked at mod_rewrite in a long time

src/.htaccess
longrunningprocess: moving it from the conf to this file.

PR #1006, opened 2021-07-16
Allow uploading images in .gif format

### LINKED ISSUE #1005: Should we allow more image file types?
While testing the Language Forge picture upload, I tried uploading a .gif file, and got the error "(filename).gif is not an allowed image file. Ensure the file is one of the following types: .jpg, .jpeg, .png."

Should we allow .gif files? What about .svg files? There may be other common image types that people might want to upload, but .gif and .svg are both going to be small files (unless it's an animated .gif), so those two file types, at least, seem like good candidates to allow people to upload.

rmunn:
Fixes #1005.

I decided to not allow .svg images yet, since to be totally safe we need to write code that would strip out any embedded Javascript. Embedded Javascript in an .svg image can be useful for totally legitimate reasons such as tecnical images that can hide or reveal sections on click, but they can also be a vector (heh) for remote code execution. So before we allow SVGs, I want to add code that will examine the uploaded .svg file and strip out any embedded Javascript, because I cannot think of any reason why .svg images in Language Forge (and FieldWorks) would need embedded Javascript.

### DIFF
--- a/src/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php
+++ b/src/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php
@@ -153,11 +153,16 @@ class LexUploadCommands
         $allowedTypes = array(
             "image/jpeg",
             "image/jpg",
+            // SVG disabled until we can ensure no embedded Javascript; see https://github.com/w3c/svgwg/issues/266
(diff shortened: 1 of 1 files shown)


megahirt (approved):
LGTM

PR #1009, opened 2021-07-22
Feature/UI scrolling enhancement

### LINKED ISSUE #1004: Toolbar surface is scrollable and is not visible as as the user scrolls down the entry
I observed a LF user have trouble with navigation between entries because the navigation control surface scrolled off the screen as they browsed the entry information.

For example:

image

The above image shows the toolbar outlined in red, which scrolls off the screen as the user scrolls down the entry.

Additionally, the side navigation list (blue marker) and the "rendered entry" (yellow highlight) also scroll off the screen.

Interim solution:

1) Fix the toolbar (red marker) to the top of the application space in a separate non scrollable div

2) Move the left navigation down so that it is in line with the top of the entry editor space.

3) add the rendered entry (yellow highlight) to the fixed toolbar space in 1) above

Important considerations: LF is a responsive UI with 4 (I think) break points.  The changes above must not break existing UI in other viewport sizes.

Our project uses Bootstrap v4 for UI components and viewport size (responsive breakpoints) definitions.

A proper solution will demonstrate that the fixed toolbar works across all viewport sizes.

**Mockup 1**
image

**Mockup 2**
image

palmtreefrb:
Project Editor content was scrolling up and hiding header controls. Made css/html changes to allow scrolling within page boundaries.
Fixes #1004

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ src/cache/*
 !src/cache/.gitkeep
 !/src/releasenotes
 test/php/\.phpunit\.result\.cache
+node_modules
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,5 +4,6 @@
     "test/php/phpunit.xml"
   ],
   "phpcs.composerJsonPath": "src",
-  "php.suggest.basic": false
+  "php.suggest.basic": false,
+  "git.ignoreLimitWarning": true
 }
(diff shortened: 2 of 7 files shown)


megahirt (approved):
LGTM, thanks Rick!  I'm so excited to see this small change make it in.

I note that E2E tests failed due to the CSS selector in the test no longer being able to select dc-rendered because it moved within the page structure.  Fixing the CSS selector in the tests should fix the E2E tests.  I have pushed a commit to do this, and am waiting for the tests to run on CI.

PR #1010, opened 2021-07-22
Run lfmerge queue manager on container startup

rmunn:
This will allow us to fix any send/receive sessions that were partway through being run when the previous container was killed (e.g. if the pod was moved to a different Kubernetes node during a S/R process). We use `which lfmergeqm && ...` to ensure that this doesn't cause the container to fail if the entrypoint script ends up run on a container without lfmerge (which might happen once we separate lfmerge into its own container in the future).

### DIFF
--- a/docker/app-base/entrypoint.sh
+++ b/docker/app-base/entrypoint.sh
@@ -3,5 +3,8 @@
 # rsyslog needs to run so that lfmerge can log to /var/log/syslog
 /etc/init.d/rsyslog start
 
+# run lfmergeqm on startup to clear out any failed send/receive sessions from previous container
+which lfmergeqm && su www-data -s /bin/bash -c lfmergeqm
+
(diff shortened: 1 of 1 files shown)


rmunn:
I discovered while testing the QA deployment that if the container is restarted during a Send/Receive process, the Send/Receive can get "stuck" in a SYNCING state: its state file is set to SYNCING and PHP therefore disables the settings menu icon, so the project manager can't kick off another Send/Receive that would notice the failed clone and restart the process. However, the LfMerge queue manager does have code to detect this state, so all we need to do is run lfmergeqm on container startup and it will fix this issue. I expect the issue to occur very rarely, probably on the order of once every year or two, but I think it *will* occur from time to time, so I'm trying to anticipate it and fix it before it happens.

megahirt (approved):
LGTM

PR #1021, opened 2021-07-23
Don't show LD projects in LF projects list view

### LINKED ISSUE #1019: Projects list should not include Language Depot projects
This issue is basically me admitting that I made a terrible design decision in the past. :grin: 

When I implemented the UI for Language Forge users to be able to manage their Language Depot projects, I included Language Depot projects in the projects list page. But thinking about that now from a user experience perspective, I'm realizing that that was a really bad idea. I should not expect users to just know the difference between a "Dictionary" project and a "Language Depot" project with no other explanatory text to guide them. And the user experience of clicking on a "Language Depot" project (where you're taken to a barebones-looking admin interface that lets you configure your project membership via the LD API) is just totally different from the user experience of clicking on a "Dictionary" project, and that's going to cause needless confusion.

I'm going to create a PR that removes the Language Depot project-management feature from the LF projects page. Site admins will still be able to administer Language Depot projects via the Language Depot API by going to the siteadmin app, but ordinary users aren't going to see the Language Depot project-management UI at all, because I'm now realizing that it would just be confusing to them.

rmunn:
Language Depot projects don't belong in the Language Forge projects list view. Too many ways to confuse the user, because clicking on those two types of projects presents very different UI but the only visible difference between them is a single line of text, shaded in a gray color that practically shouts (or rather, whispers) "this text isn't at all important, pay no attention to it".

This will fix #1019.


rmunn:
Have now gone through and removed all the now-unused code. There was quite a lot of it in the end; the only thing remaining is the ldproject-members component, which is used in the siteadmin app. I could move that component into siteadmin, but I decided to leave it in place, because at some point in the future we'll want to bring the ldproject app back (with a better design) and at that point we'll want that component to be where it is right now. It's perfectly okay for the siteadmin app to depend on the ldproject-members app, but I don't want the ldproject app to be depending on anything from the siteadmin app. Only a few people should have access to the siteadmin app, but lots of people will (eventually) have access to the ldproject app.

megahirt (approved):
LGTM - deleting code is always nice :)

PR #1030, opened 2021-07-29
merge app-base into app Dockerfile; establish BUILD_VERSION in GA

### LINKED ISSUE #1027: Implement GitHub Action version stamping
This will solve the following problems:
- cache busting (JS files stamped with version number)
- docker images stamped with version number
- feeds into the version.php which informs the application what it's version is

### LINKED ISSUE #1031: remove xdebug and composer from production image
consider removing `xdebug` and `composer` from app-base Dockerfile (production) - @longrunningprocess

### LINKED ISSUE #1032: remove php-extension-installer from production app image
consider removing `php-extension-installer` from app-base Dockerfile after it's used - @longrunningprocess

### LINKED ISSUE #1033: remove pnupg2 from production app image
consider removing `gnupg2` from app-base/Dockerfile after it's used. - @longrunningprocess

megahirt:
The app-base Dockerfile was created for developer convenience to simplify and speed up building of the docker image during development of the Dockerfile.

In preparation for a production image, we can merge these two back together and provide a means for switching between production and development images via a build argument "environment"

Resolves #1033, resolves #1032, resolves #1031 

This should really be two PRs :(

Establish a BUILD_VERSION variable in github actions.
Resolves #1027


docker/app/Dockerfile
```diff
@@ -1,26 +1,84 @@
-FROM sillsdev/web-languageforge:app-base-staging
+ARG ENVIRONMENT=production
+
+# APP-BASE
+FROM php:7.3.28-apache AS base-app
+
+# install apt packages
+# p7zip-full - used by LF application for unzipping lexicon uploads
+# unzip - used by LF application for unzipping lexicon uploads
+# git - TODO: why do we need git?
```
longrunningprocess, line 9: this was added to avoid cert warnings...

docker/app/Dockerfile
```diff
@@ -1,26 +1,84 @@
...
+# gnupg2 - necessary for LFMerge package installation via SIL sources (will be uninstalled in production)
+# curl - used by LF application
```
longrunningprocess, line 11: thank you for adding the "why" on these, rally appreciated!

docker/app/Dockerfile
```diff
@@ -1,26 +1,84 @@
...
+RUN apt-get update && apt-get -y install p7zip-full unzip git gnupg2 curl
+
+# see https://github.com/mlocati/docker-php-extension-installer
+# PHP extensions required by the LF application
+COPY --from=mlocati/php-extension-installer /usr/bin/install-php-extensions /usr/local/bin/
+RUN install-php-extensions gd mongodb intl
```
longrunningprocess, line 17: this might be the point where you would add a compound statement to remove anything that's not needed anymore.  We can chat about it if needed.

docker/app/Dockerfile
```diff
@@ -1,26 +1,84 @@
...
+
+# install and configure LFMerge
+RUN curl -L http://linux.lsdev.sil.org/downloads/sil-testing.gpg | apt-key add - \
+&& echo "deb http://linux.lsdev.sil.org/ubuntu bionic main" > /etc/apt/sources.list.d/linux-lsdev-sil-org.list \
+&& apt-get update \
+&& apt-get install --yes --no-install-recommends python lfmerge rsyslog logrotate cron iputils-ping
```
longrunningprocess, line 23: there be an opportunity here to remove anything not needed anymore, e.g., apt-get cache or anything like that.  If something exists like that, it should of course be done as part of a compound statement here.

docker/app/Dockerfile
```diff
@@ -1,26 +1,84 @@
...
+COPY docker/app/lfmerge.conf /etc/languageforge/conf/sendreceive.conf
+COPY docker/app/lfmergeqm-background.sh /
+RUN adduser www-data fieldworks \
+&& chown -R www-data:www-data /var/lib/languageforge \
+&& chmod 0755 /var/lib/languageforge \
+&& mkdir -m 02775 -p /var/www/.local \
+&& chown www-data:www-data /var/www/.local
 
 # php customizations
```
longrunningprocess, line 32: I think we should double-check the final version of the ini that gets used...I would assume there are things in the production version that are helpful from a security perspective.

docker/app/Dockerfile
```diff
@@ -1,26 +1,84 @@
...
-ENV COMPOSER_ALLOW_SUPERUSER=1
-RUN composer install
-
 # copy app into image
 COPY src /var/www/html/
 RUN ln -s /var/www/html /var/www/src
 
 # grab the built assets from the ui image
+# TODO distinguish between production and dev builds here
 COPY --from=lf-ui-builder /data/src/dist /var/www/html/dist
```
longrunningprocess, line 45: if we get the labels consistent, we should be able to do this with a build arg.

version_number.txt
```diff
@@ -0,0 +1 @@
+1.8.0
```
longrunningprocess, line 1: I wonder if there's a possibility of getting this from a git ref for the release version and then using something either random or static during development?  Maybe we can talk it through a bit in person.

megahirt (commented):
leaving some notes for me to address.  These were generated during the standup discussion.

docker/app/Dockerfile
megahirt: @rmunn #1034 creates a bash script which runs indefinitely in a loop - a simple man's cronjob.  This was to avoid having the cron dependency, however I see in this line that we do install cron.  I'm wondering if A) we can remove the cron dependency or B) we should just switch over to using cron since we've already got it.  Would prefer to remove cron if that was the only thing it's needed for - looking for your insight into this one.  Thanks.

rmunn (approved):
While there's a bit more we want to do (like address how the `ui-builder` container is handled), this LGTM and is passing tests on TeamCity, so it's probably best to "save" (merge) what we have, which is good forward progress. Then we'll start a new PR for any remaining work.

PR #1034, opened 2021-07-29
Run lfmergeqm every 24 hours

rmunn:
This will allow lfmergeqm's Janitor class to find and clean up any failed syncs that aren't in the queue (state is SYNCING but there's no active lfmerge process handling that project). The main reason this can happen is if the app container was restarted in the middle of a Send/Receive, killing a previous lfmerge process.

This will have a minor merge conflict with #1030, which moves and rearranges some of the files touched here.

### DIFF
--- a/docker/app-base/entrypoint.sh
+++ b/docker/app-base/entrypoint.sh
@@ -4,7 +4,10 @@
 /etc/init.d/rsyslog start
 
 # run lfmergeqm on startup to clear out any failed send/receive sessions from previous container
-which lfmergeqm && su www-data -s /bin/bash -c lfmergeqm
+if [ which /lfmergeqm-background.sh ]; then
+    # MUST be run as a background process as it kicks off an infinite loop to run every 24 hours
+    /lfmergeqm-background.sh &
+fi
 
 # Now chain to Docker entrypoint from base php:apache image
 exec docker-php-entrypoint "$@"
--- a/docker/app-base/Dockerfile
+++ b/docker/app-base/Dockerfile
@@ -24,6 +24,7 @@ RUN adduser www-data fieldworks \
 && chown www-data:www-data /var/www/.local
 
 COPY entrypoint.sh /
+COPY lfmergeqm-background.sh /
 
 ENTRYPOINT [ "/entrypoint.sh" ]
 CMD [ "apache2-foreground" ]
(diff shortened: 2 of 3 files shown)


megahirt (approved):
LGTM.

There's more than one way to do this.  And this is one way to do it.  Thanks Robin.

docker/app-base/entrypoint.sh
```diff
@@ -4,7 +4,10 @@
 /etc/init.d/rsyslog start
 
 # run lfmergeqm on startup to clear out any failed send/receive sessions from previous container
-which lfmergeqm && su www-data -s /bin/bash -c lfmergeqm
+if [ which /lfmergeqm-background.sh ]; then
```
longrunningprocess, line 7: `which` is not present in this image

megahirt: @rmunn sounds like the lfmergeqm isn't working right off the bat.  @longrunningprocess I have seen the errors in the log you mention and thought "we should fix those".  I have created an issue to track the errors in the log #1055

rmunn: *Boggle*. What kind of Docker image doesn't have `which` in it? I understand not having `ping`, but Busybox implements `which`. I just assumed it would be there, as a tool I could use in my scripts to verify that a *different* command exists before I run it.

I'll see what I can do about tweaking that script to run in the absence of a `/bin/which` command.

rmunn: The Docker image *does* have `which`, I just messed up the syntax. Instead of `if [ which some-command ]; then`, I needed to omit the `[ ]` and write `if which some-command; then`.

rmunn: https://github.com/sillsdev/web-languageforge/pull/1059 fixes the `[: which: unexpected operator` issue. I'll make a separate PR for the `imklog` issue.

rmunn: Changed my mind and put both fixes into #1059 since they're both small and easy to review. So that one PR will fix both of these issues.

longrunningprocess: My important point was simply to remove the conditional altogether, you're installing `lfmergeqm-background.sh` in the Dockerfile already, seems silly to me to check to make sure it's there.

PR #1038, opened 2021-07-30
Move ui-builder into app Dockerfile

### LINKED ISSUE #1028: refactor ui-builder use in app Dockerfile
Copy ui-builder Dockerfile into app Dockerfile and run as a multi-stage build step.

rmunn:
This will allow us to simplify the Makefile and the build process; now building the `app` container will just require building one single Dockerfile.

Resolves #1028.

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -33,22 +33,12 @@ else
 endif
 
 .PHONY: build
-build: build-ui
+build:
 	docker-compose build mail app ld-api
 
-.PHONY: build-ui
-build-ui:
-	docker-compose build --build-arg BUGSNAG_API_KEY=${BUGSNAG_API_KEY} ui-builder
-
 .PHONY: deployable-app
-deployable-app: build-ui
+deployable-app:
 	docker-compose build --build-arg BUILD_VERSION=${BUILD_VERSION} app
-
-.PHONY: prod
-prod: build-ui
-	docker build -f ui-builder/Dockerfile.prod . -t lf-ui-builder:prod
-	docker-compose build app
-	docker build -f app/Dockerfile.prod .. -t lf-app:prod
 
 .PHONY: logs
 logs:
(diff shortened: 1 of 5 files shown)


rmunn:
Not yet in this PR: changes to the Makefile and docker-compose.yml to account for `app` being only a single Dockerfile (no more separate ui-builder build step).

megahirt:
> Not yet in this PR: changes to the Makefile and docker-compose.yml to account for `app` being only a single Dockerfile (no more separate ui-builder build step).

done in latest commit.

longrunningprocess (approved):
seems good to me, I would not expect this to have any implications on the unit tests but we need to make sure things like the e2e tests, "developer use case" (i.e., can a dev still make work on the app locally while making changes and testing those?) and deployments are still ok.

megahirt:
> seems good to me, I would not expect this to have any implications on the unit tests but we need to make sure things like the e2e tests, "developer use case" (i.e., can a dev still make work on the app locally while making changes and testing those?) and deployments are still ok.

This is a very good point, @longrunningprocess .  My understanding is that the ui-builder which runs in a container during the dev use case will produce dist files that are volume mapped in via docker-compose orchestration.  It needs to be tested to confirm before this is merged in.

@rmunn can you test the developer use case for e2e and unit tests on this branch to confirm things are still working as intended?

rmunn:
> @rmunn can you test the developer use case for e2e and unit tests on this branch to confirm things are still working as intended?

Confirmed. Running `make dev` kicked off the `ui-builder` container with a Webpack watch in it, and then when I made changes in the code they were automatically reflected in Firefox after I hit Ctrl+R. (We don't have any browser auto-refreshing set up, so you still do have to refresh the browser to see the edits you just saved, and it does take Webpack about 3-5 seconds to compile so it's possible to be overeager and refresh too soon.) I was also able to run the E2E and unit tests and have my changes reflected in the result. I also tested the E2E debugging, and it still worked as it should.

In the process of testing E2E, I figured out a way to make the E2E tests more consistent by ensuring that the app-for-e2e container gets restarted each time you run `make e2e-tests`, but only on developer machines. That way if you interrupted the tests halfway through, or a test failed and therefore didn't manage to "clean up after itself" by resetting the state it had changed, the next run of the E2E tests will start with the intended initial state. It's a one-line change to the Makefile; I'll push that commit and then merge this.

PR #1041, opened 2021-07-30
use BUILD_VERSION to tag docker images

megahirt:
DO NOT MERGE until rebased properly and target set back to master.

This utilizes the BUILD_VERSION variable to tag and push docker images


longrunningprocess:
I get the sense we may not take this approach...not sure if I'm interpreting the Slack conversation right or not but please ping me if you do want me to look at this PR.  I'm not a fan of the version in the filename if we can find a better way to do this but am also not opposed to this being a stopgap with a plan to find a better solution later.

megahirt:
This PR will likely not be merged.  Keeping out there for a few more days until a more elegant solution is described and implemented.

megahirt:
This PR is being replaced by #1066 and will be closed.

PR #1042, opened 2021-08-02
address review comments

megahirt:
This change should have made it in with the original PR but didn't.


rmunn:
Rebased on top of master so that E2E tests can run.

rmunn (approved):
LGTM.

PR #1040, opened 2021-07-30
Remove Bugsnag

rmunn:
We aren't using it at the moment, so it's just making the Javascript bundle larger to no purpose.

Resolves #1035.

PR #1043, opened 2021-08-03
removed bugsnag refs

longrunningprocess:
these were accidentally missed in #1040


longrunningprocess:
that was a dumb attempt, I thought those were `docker` commands for some reason, not `docker-compose`...

rmunn (approved):
LGTM

PR #1046, opened 2021-08-04
Fix E2E tests not starting after clean-powerwash

rmunn:
The Selenium docker image tagged 3.141.59-20210729 contains a browser version that our current Protractor version doesn't know how to handle.

Temporarily pin our Selenium docker image to 3.141.59-20210713 so that our E2E tests will run; later when we upgrade Protractor we can unpin this and run the latest Selenium image again.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -109,7 +109,7 @@ services:
       - MONGO_INITDB_DATABASE=scriptureforge
 
   selenium:
-    image: selenium/standalone-chrome-debug
+    image: selenium/standalone-chrome-debug:3.141.59-20210713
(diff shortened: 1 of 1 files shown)


rmunn:
If the E2E tests pass on TeamCity then I'm going to merge this without waiting for review, as there's literally nothing to review. Either this version of the Docker image works, or it doesn't, and the E2E tests will tell me that without needing a code review.

rmunn:
E2E tests passed. Unit tests aren't starting up, and I don't know why. The agent is starting up but not picking up the build. But since the E2E tests passed, and this PR makes no changes to unit tests, I'm going to merge it.

PR #1047, opened 2021-08-05
Feature/refactor replace optimize libraries imports

palmtreefrb:
#1036
1. Added webpack plugin to remove unneeded localization.
2. Added webpack optimization option "usedExports: true" for tree shaking.
3. Changed package references to use min files "angular-moment-picker" and "ng-file-upload-all"
4. Removed unused class references throughout angular ui.
5. Added webpack-bundle-analyzer to dev  #build


rmunn:
This doesn't change the `moment.js` dependency yet; `moment` is still in the `package.json` file, and is still used in a couple places in the code ([src/angular-app/bellows/core/filters.ts](https://github.com/sillsdev/web-languageforge/blob/304cbd682e27b4fe6ca72f8a48ed99090a2142c8/src/angular-app/bellows/core/filters.ts) and [src/angular-app/languageforge/lexicon/editor/field/dc-audio.component.ts](https://github.com/sillsdev/web-languageforge/blob/304cbd682e27b4fe6ca72f8a48ed99090a2142c8/src/angular-app/languageforge/lexicon/editor/field/dc-audio.component.ts)).

I also suggest that you rebase this on top of the latest commit in the `master` branch so that the E2E tests will run. (I pushed a commit yesterday that fixes the issue with E2E tests not starting).

src/angular-app/bellows/apps/activity/activity-app.module.ts
```diff
@@ -1,6 +1,6 @@
 import * as angular from 'angular';
-import 'angular-moment-picker/dist/angular-moment-picker.css';
-import 'angular-moment-picker/dist/angular-moment-picker.js';
+import 'angular-moment-picker/dist/angular-moment-picker.min.css';
+import 'angular-moment-picker/dist/angular-moment-picker.min.js';
```
rmunn, line 3: I don't think this is a good idea. In development, we should be importing non-minified files so that we can actually read the code we're importing. Webpack should be able to minify the fles for us, and if it can't, that's a bug (probably our config needs to be changed and then it will work).

Also, off the top of my head I don't think importing `angular-moment-picker/dist/specific-file.js` is right either. Surely we should just be doing `import * as angularMomentPicker from 'angular-moment-picker'` and let its `package.json` file declare the correct entrypoint. (I haven't looked in-depth into this one so I could be off-base here).

src/angular-app/bellows/main.common.ts
```diff
@@ -5,7 +5,7 @@ import 'angular-ui-router';
 import 'angular-ui-validate';
 import 'angular-zxcvbn';
 
-import 'ng-file-upload/dist/ng-file-upload-all.js';
+import 'ng-file-upload/dist/ng-file-upload-all.min.js';
```
rmunn, line 8: Same comment as for the `angular-moment-picker` import. Why are we importing a specific file from a package? That doesn't seem right.

rmunn (commented):
I'm not sure why we're importing specific files from npm libraries instead of importing the library, but changing those imports to import minified versions doesn't feel like the right answer to me. See my specific comments for more details.

megahirt (changes requested):
Let's not change to using the min version of those JS files in the code.  This is a place where we need to work out some of these details - however it seems like too large for this PR.

longrunningprocess:
I just worked through some merge conflicts so we could get this passing e2e tests...

palmtreefrb:
1. Completely removed momentjs.
2. Added date-fns and refactored momentjs implementation code to use date-dns functions
3. Replaced javascript import references with proper npm imports
4. Moment date picker was commented out.

palmtreefrb:
...trying again.

src/angular-app/languageforge/lexicon/editor/field/dc-audio.component.ts
```diff
@@ -144,7 +144,8 @@ export class FieldAudioController implements angular.IController {
 
   audioRecorderCallback = (blob: Blob) => {
     if (blob) {
-      const fileName = 'recording_' + moment.utc().format('YYYY_MM_DD_HH_mm_ss') + '.mp3';
+      const date = new Date();
+      const fileName = 'recording_' + format(addMinutes(date, date.getTimezoneOffset()), 'YYYY_MM_DD_HH_mm_ss') + '.mp3';
```
rmunn, line 148: The logic is correct here, but `date-fns` uses the date formatting symbols from the [Unicode standard](https://www.unicode.org/reports/tr35/tr35-dates.html#Date_Field_Symbol_Table), where `YYYY` and `DD` don't mean what you probably expect them to mean. See https://github.com/date-fns/date-fns/blob/master/docs/unicodeTokens.md for a broader explanation. The correct format string to use here is `'yyyy_MM_dd_HH_mm_ss'`. With that format string, this will produce the same filenames that it used to with the moment.js function.

src/angular-app/bellows/core/filters.ts
```diff
@@ -20,10 +19,10 @@ export type RelativeTimeFilterFunction = (timestamp?: string, timeFormat?: strin
...
 export function RelativeTimeFilter(): RelativeTimeFilterFunction {
   return (timestamp?: string, timeFormat?: string): string => {
-    // see http://momentjs.com/docs/
-    const timeAgo = moment(timestamp, timeFormat);
-    if (timeAgo.isValid()) {
-      return timeAgo.fromNow();
+    const date = new Date(timestamp);
+    const dateNow = new Date();
+    if (isValid(date) && isValid(dateNow)) {
+      return formatDistance(date, dateNow, { addSuffix: true });
```
rmunn, line 25: This can be simplified by using [the `formatDistanceToNow` function](https://date-fns.org/v2.23.0/docs/formatDistanceToNow), and then you don't need `dateNow`.

src/angular-app/bellows/apps/activity/activity-container.component.html
```diff
@@ -52,7 +52,7 @@
                     </div>
                     <button class="btn btn-sm brn-default" data-ng-click="$ctrl.filterDateOpen = !$ctrl.filterDateOpen">Close <i class="fa fa-times"></i></button>
                 </div>
-            </div>
+            </div> -->
```
rmunn, line 55: Good that you commented this out instead of removing it, as that will help us remember to come back and add a replacement date picker at some point. Could you add a comment to that effect (like "TODO: Add a date picker that doesn't use moment.js") so we'll know why this is commented out in the future, when we've forgotten about this PR?

rmunn (changes requested):
Everything looks good except for the incorrect formatting code for the date-fns `format` function (see details below). Once that is fixed, I'm ready to approve this.

src/angular-app/bellows/core/filters.ts
rmunn: Also, [beware of parsing dates with the `new Date(someString)` constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#parameters)! The Mozilla documentation warns against it:

> **Note:** Parsing of date strings with the `Date` constructor (and `Date.parse()`, which works the same way) is strongly discouraged due to browser differences and inconsistencies.
> * Support for [RFC 2822](https://datatracker.ietf.org/doc/html/rfc2822) format strings is by convention only.
> * Support for ISO 8601 formats differs in that date-only strings (e.g. `"1970-01-01"`) are treated as UTC, not local.

Proof that this can have very surprising results (Firefox on Linux):

image

I ran the following snippet in Node to verify this behavior, and sure enough:

```js
import { isValid, formatDistanceToNow } from 'date-fns';

function test(timestamp) {
    const date = new Date(timestamp);
    if (isValid(date)) {
      console.log(date);
      return formatDistanceToNow(date, { addSuffix: true });
    }
    return ''
}

console.log(test('2021-07-15'))
console.log(test('2021/07/15'))
```

Result:

```plaintext
2021-07-15T00:00:00.000Z
27 days ago
2021-07-14T17:00:00.000Z
28 days ago
```

Since the date strings coming into the filter are from MongoDB and should be in ISO format (if they aren't, it's a bug) then it's probably best to use date-fns' [parseISO](https://date-fns.org/v2.23.0/docs/parseISO) function instead.

megahirt (approved):

rmunn:
Oh good, I was hoping my review wasn't going to hold things up. I checked in this morning to see if a new commit had been pushed and it hadn't, but I see my checking in just now wasn't necessary. Glad to see GitHub allows repo admins to do things like that. I'll go back to my holiday now. :-)

longrunningprocess:
> Oh good, I was hoping my review wasn't going to hold things up. I checked in this morning to see if a new commit had been pushed and it hadn't, but I see my checking in just now wasn't necessary. Glad to see GitHub allows repo admins to do things like that. I'll go back to my holiday now. :-)

have a great one!

Issue #1055, opened 2021-08-06
rsyslog lfmerge errors in app container


megahirt:
According to [comments in 1034](https://github.com/sillsdev/web-languageforge/pull/1034#discussion_r683710223) we are seeing:

- errors in the init container related to not having which installed
- errors in the app container related to rsyslog.  I don't know exactly what it means:
```
rsyslogd: activation of module imklog failed [v8.1901.0 try https://www.rsyslog.com/e/2145 ]
```
Looks like https://stackoverflow.com/a/60265997 has an acceptable way to comment out the imklog module from rsyslog.

rmunn:
The `which` issue is fixed by #1059. Turns out that `which` is installed, I just didn't use it properly in the `if` statement. (The brackets weren't needed.)

I'll submit another PR for the imklog issue.

rmunn:
Changed my mind and put both fixes into #1059, so both of these issues can be fixed with a single PR.

PR #1059, opened 2021-08-06
Fix mistake in entrypoint.sh script

rmunn:
It's not `if [ which some-cmd ]`, it's `if which some-cmd`. It can also be a good idea to redirect the output of `which` to `/dev/null` so that you don't end up printing the full path of the command in your server logs (which is almost never necessary).

### DIFF
--- a/docker/app/entrypoint.sh
+++ b/docker/app/entrypoint.sh
@@ -12,7 +12,8 @@ fi
 /etc/init.d/rsyslog start
 
 # run lfmergeqm on startup to clear out any failed send/receive sessions from previous container
-if [ which /lfmergeqm-background.sh ]; then
+if (which /lfmergeqm-background.sh > /dev/null)
+then
     # MUST be run as a background process as it kicks off an infinite loop to run every 24 hours
     /lfmergeqm-background.sh &
 fi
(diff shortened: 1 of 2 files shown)


rmunn:
Also decided to fix the `rsyslog: imklog: cannot open kernel log` issue in the same PR. The `imklog` module (which reads kernel messages) isn't allowed in Docker containers since it needs kernel access. But since we don't *want* kernel messages from the host kernel showing up in the container logs, we can just comment out the `imklog` line and be happy.

So this PR fixes both the issues reported in #1055.

docker/app/entrypoint.sh
```diff
@@ -12,7 +12,8 @@ fi
 /etc/init.d/rsyslog start
 
 # run lfmergeqm on startup to clear out any failed send/receive sessions from previous container
-if [ which /lfmergeqm-background.sh ]; then
+if (which /lfmergeqm-background.sh > /dev/null)
```
megahirt, line 15: Per our conversation, the parens for invoking a sub-shell are not necessary here.  But also doesn't seem to affect it.  Remove if you want, but going to approve this PR anyway.

megahirt (approved):
LGTM

docker/app/entrypoint.sh
longrunningprocess: I would've preferred to see this conditional removed altogether, seems silly to me (https://github.com/sillsdev/web-languageforge/pull/1034#discussion_r685168994)

PR #1064, opened 2021-08-09
Fix "data.slice is not a function" exception

rmunn:
The exception was caused by the Language Depot API returning HTTP 403 Forbidden instead of HTTP 200 OK, because the JWT token keys weren't actually set up correctly in development mode. Also, our Ldapi.php code doesn't handle HTTP 403 Forbidden correctly, which is something else I'll need to work on. (Somehow it ends up throwing an unhandled exception in Javascript land, in a way I haven't quite traced yet). But this PR will at least stop that error being thrown during E2E testing in development.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -145,6 +145,7 @@ services:
       - MYSQL_USER=redmine
       - MYSQL_PASSWORD=not-a-secret
       - MYSQL_ROOT_PASSWORD=also-not-a-secret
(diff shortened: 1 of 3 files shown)


megahirt (approved):
LGTM

longrunningprocess:
is `JWT_SIGNING_KEY` something we'll need to ensure is part of the `staging` and `production` environments as well?

Issue #1060, opened 2021-08-06
Branching, Tagging and Release Process


megahirt:
This issue lays out the branching, tagging and release process for LF

We will follow a modified version of [GitFlow as described in this seminal article](https://datasift.github.io/gitflow/IntroducingGitFlow.html).  A diagram from the article:
![image](https://user-images.githubusercontent.com/3444521/128802895-2e68ab9d-8abe-4176-a6ae-1bc550d7c92a.png)

### Develop
In conforming to GitFlow and the way that others would expect to interact with our repo, the **develop** branch will now be our default branch for feature work and PRs.  **develop** will be our default git branch.  All PRs for new features will be against **develop**.

### Staging
The staging branch is like the "release" branch in the illustration above.  We make stabilizing squash commits (still make PR branches) on staging much like we do on **develop**.  In the process of preparing to release, developers create PR branches against the **staging** branch.  These are squash commits just like on develop.

### Master
The **master** branch as seen in the illustration above represents our production release.  Git tags on the **master** branch will indicate a published image from that commit.  **Master** will contain only merge commits from **staging** and an unlikely hotfix.

## In What Ways Do We Differ From GitFlow?

- We do squash commits for single-author PRs. (most are single-author)
- Merge commits are reserved for merging branches (master, staging, develop) into each other
-  We have a single **staging** branch instead of multiple feature branches.

## Prerequisites
- The build number must be established/made available in GitHub Actions as that is where the application is built for production, the image is tagged and the image pushed to DockerHub.
- The build version number will be SemVer

## Release Process
- The LF team will ideally release a new minor version once a month e.g. v1.8 (August), v1.9 (September), etc.  These releases are characterized by a public announcement and "GitHub Release" created with accompanying release notes.
- In between minor releases, small stabilizing bugfix releases may be shipped.  These smaller releases are not publicized widely or call for a new Github Release.  Directly communication with users who may be affected by these stabilizing releases is necessary, but it's not necessary to communicate to the general public about bugfix releases unless it affects everyone.
- Each release should have a GitHub Release associated with the monthly major/minor release
- We also maintain a release notes file that is linked to from within the application.  This would be identical to the GitHub release in content.

## Build Number Usage
Each shipped release will be referenced by a unique build version number.  This number will be used for:
- Docker image tagging
- displaying the version number in the running application
- cache busting
- tagging a commit in our repo

## Build Number Determination
[...]

PR #1066, opened 2021-08-10
Update image build trigger and BUILD_VERSION

megahirt:
Update how we are defining BUILD_VERSION at image build time in GitHub Actions, including differentiating between staging image builds and production image builds.

Production image builds are triggered off of tags e.g. v1.8.0 whereas staging image builds are triggered by any commit pushed on the staging branch.

This commit implements the Docker image tagging and building as outlined in issue #1060.

version_number.txt has been deleted as we no longer reference it in the GA build.

This PR supersedes #1041


.github/workflows/build-and-deploy-images.yml
```diff
@@ -3,6 +3,8 @@ name: Deploy to Docker Hub
 # https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#on
 on:
   push:
+    tags:
+      - "v*"
```
longrunningprocess, line 7: if GHA has a regex way to do this without the `v` then it would simply the process a little by not requiring one tag with the `v` and another tag without...

.github/workflows/build-and-deploy-images.yml
```diff
@@ -40,25 +42,32 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Establish build version environment variable
-        run: echo "BUILD_VERSION=$(cat version_number.txt).$GITHUB_RUN_NUMBER" >> "$GITHUB_ENV"
+      - name: Set BUILD_VERSION for Staging
+        if: github.ref == 'refs/heads/staging'
```
longrunningprocess, line 46: nice construct, I wasn't aware it had this!

longrunningprocess (approved):
I'm very happy to see that `.txt` file go!  Slick work here, take a look the possibility of eliminating the need for extra tags and see if there's an opportunity to simplify things or not.  Either way, this is nice!

I suppose we'll also need to work in the parameterization of this build number into the deployment yamls and the cache busting mechanism in the twig containers but I guess those will come in another PR?

megahirt:
> I'm very happy to see that `.txt` file go! Slick work here, take a look the possibility of eliminating the need for extra tags and see if there's an opportunity to simplify things or not. Either way, this is nice!
> 
> I suppose we'll also need to work in the parameterization of this build number into the deployment yamls and the cache busting mechanism in the twig containers
Cache busting should be good to go with the existing code, see https://github.com/sillsdev/web-languageforge/blob/develop/docker/app/Dockerfile#L129 which injects the BUILD_VERSION into PHP as a variable which the application will use in the Twig template.  A future PR could get rid of version.php entirely in favor of a container environment variable.

.github/workflows/build-and-deploy-images.yml
rmunn: [GitHub Actions filter syntax](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet) has several regex-like features like being able to write `tags: v[1-9]+.[0-9]+.[0-9]+`, but it doesn't have capture groups. Also notice that the `.` was not escaped: these are not regular expressions either. We can't write `v(.*)` and capture the non-v-prefixed version of the release number, so we have to do it later. (I'll have a comment at that point, too).

.github/workflows/build-and-deploy-images.yml
```diff
@@ -40,25 +42,32 @@ jobs:
...
+        run: echo "BUILD_VERSION=staging-$GITHUB_SHA" >> "$GITHUB_ENV"
+        run: echo "BUILD_VERSION_LATEST=staging" >> "$GITHUB_ENV"
+
+      - name: Set BUILD_VERSION for Production from git tag
+        if: contains(github.ref, 'refs/tags/')
```
rmunn, line 51: I suggest adding a `v` at the end of the search string here, so that we're searching for `refs/tags/v`. Because that's exactly the number of characters we're stripping. And it will help provide a small amount of future-proofing if we ever change the `on: tags:` requirement to omit the `v` and change the tag we use to `releases/1.8` or something. I'd rather have the workflow fail to run than have it push something tagged as `eleases/1.8`, missing the initial `r`.

Also, [there's a `startsWith` function](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#startswith), so that's better than `contains` here.

.github/workflows/build-and-deploy-images.yml
```diff
@@ -40,25 +42,32 @@ jobs:
...
+        # define GIT_TAG as stripping off initial 11 characters, i.e. "refs/tags/v"
+        run: echo "GIT_TAG=$(echo ${GITHUB_REF:11})" >> $GITHUB_ENV
+        run: echo "BUILD_VERSION=$GITHUB_TAG" >> "$GITHUB_ENV"
```
rmunn, line 54: I'm pretty sure you meant `GIT_TAG` here. As far as I can tell, `GITHUB_TAG` isn't an env var that GitHub Actions provides, so this would make `BUILD_VERSION` an empty string.

rmunn (changes requested):
The `GITHUB_TAG` variable name should have been `GIT_TAG`. Once that's changed, I'm happy to approve this.

.github/workflows/build-and-deploy-images.yml
megahirt: done.  Thanks for the GIT_TAG catch!

.github/workflows/build-and-deploy-images.yml
megahirt: done.

.github/workflows/build-and-deploy-images.yml
megahirt: I wasn't as confident about using the tag filter syntax.  I realize that we could be triggering builds for any tag that starts with "v", but I'm willing to take that risk for the time being.

rmunn (approved):
LGTM now.

PR #1067, opened 2021-08-10
Parameterized docker image version for deployment

longrunningprocess:
This will piggy-back off the work going on in #1066 to utilize unique versioning of the app.


longrunningprocess:
no reason to question whether it works or not 😎

PR #1079, opened 2021-08-16
Tech Support user can manage project members

### LINKED ISSUE #1078: Tech Support role can't manage project membership
The Tech Support role is intended to have all the same rights as managers, and we've fixed bugs in the past where that wasn't the case. I just found another one, where the Tech Support role wasn't given manager rights to the Share With Others component (which is where project membership is managed).

rmunn:
The intent of the Tech Support user role is that it has the same rights as a project manager, i.e. it can manage all aspects of the project. The "Share With Others" component was not giving Tech Support users the same rights as the manager: Tech Support was unable to list project members, manage their roles, and invite new users. This fixes that omission.

Fixes #1078.

### DIFF
--- a/src/angular-app/languageforge/lexicon/shared/share-with-others/invite-member-form.component.ts
+++ b/src/angular-app/languageforge/lexicon/shared/share-with-others/invite-member-form.component.ts
@@ -42,7 +42,9 @@ export class InviteMemberFormController implements angular.IController {
     this.sessionService.getSession().then(session => {
       this.session = session;
       this.project = session.data.project;
-      this.currentUserIsManager = this.session.data.userProjectRole === LexRoles.MANAGER.key;
+      this.currentUserIsManager =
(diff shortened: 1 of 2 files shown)


rmunn:
Changing PR to merge into staging instead of develop, because staging is where I first noticed the bug.

rmunn:
TeamCity isn't set up to run PRs against `staging`, so I ran E2E tests locally. They all passed.

megahirt (approved):
LGTM  Thanks Robin!

rmunn:
Closing and re-opening to test whether TeamCity will pick it up now that TC settings allow PRs into `staging` to run E2E tests now.

longrunningprocess:
I would think we would always merge into `develop` and then into `staging`...

so is your plan to merge these into `develop` after the fact?

Maybe this process would be a good topic for discussion at our stand-up so I'm clear, thanks.

PR #1093, opened 2021-08-23
Slightly longer wait time for first lfmergeqm run

rmunn:
There's a chance that 1 minute is not quite long enough, so we'll wait 2 minutes to be safe.


megahirt (approved):

rmunn:
Closing without merging as experience has shown this isn't actually needed.

PR #1095, opened 2021-08-23
feat: multi-line auto-sizing multitext inputs

palmtreefrb:
First commit changing html element from text to textarea. Still need to add styling.


megahirt:
@palmtreefrb ok great, go ahead and commit whenever.  It's still a draft so feel free to commit as a "draft"

palmtreefrb:
This is ready for review. All changes are checked in.

Screenshot from 2021-08-30 18-00-01

.vscode/settings.json
```diff
@@ -5,5 +5,6 @@
   ],
   "phpcs.composerJsonPath": "src",
   "php.suggest.basic": false,
-  "git.ignoreLimitWarning": true
+  "git.ignoreLimitWarning": true,
+  "terminal.integrated.cwd": "${workspaceFolder}/docker"
```
megahirt, line 9: This seems like a personal preference, and not something you are intending to push to all developers, so I would leave this out.
```suggestion
```

docker/test-e2e/.vscode/launch.json
```diff
@@ -13,8 +13,8 @@
         "run-script",
         "test-e2e",
         // Uncomment the two lines below and edit the --specs value to run just one file (for more than one file, make it a comma-separated list)
-        // "--",
-        // "--specs=test/app/bellows/change-password.e2e-spec.js"
+        "--",
+        "--specs=test/app/languageforge/lexicon/settings/semantic-domains.e2e-spec.js"
```
megahirt, line 17: It would seem that you made these changes for debugging purposes but didn't intend to commit them.

I would prefer you to revert changes to launch.json in this case.  Thanks!

docker/test-e2e/.vscode/launch.json
```diff
@@ -37,8 +37,8 @@
...
```
megahirt, line 41: See above.

src/angular-app/languageforge/lexicon/editor/_editor.scss
```diff
@@ -3,15 +3,15 @@
...
-@import 'field/dc-semanticdomain';
+// @import 'comment/dc-comment';
+// @import 'field/dc-audio';
+// @import 'field/dc-text';
+// @import 'field/dc-multioptionlist';
+// @import 'field/dc-multiparagraph';
+// @import 'field/dc-multitext';
+// @import 'field/dc-picture';
+// @import 'field/dc-rendered';
+// @import 'field/dc-semanticdomain';
```
megahirt, line 14: If these lines are not needed, then we should just delete them.

```suggestion
```

src/angular-app/languageforge/lexicon/lexicon.scss
```diff
@@ -1,4 +1,4 @@
-@import 'editor/editor';
+// @import 'editor/editor';
```
megahirt, line 1: delete imports that are not necessary

```suggestion
```

src/js/lib/angularjs-autogrow.js
```diff
@@ -0,0 +1,98 @@
+(function(){
```
megahirt, line 1: I see that you checked-in the angularjs-autogrow library, presumably from https://github.com/evyros/angularjs-autogrow

This library is available as an npm library, but you've chosen to check it in.  This works because we (unfortunately) still have a way to include checked-in-dependencies outside of webpack.  Could you explain your rationale for choosing this method over the npm dependency, which would be preferred?

If we do end up checking it in, we should have a comment referencing where the code came from, as an internet citation or sorts.  e.g. https://github.com/evyros/angularjs-autogrow

megahirt (changes requested):
I have a few requested changes:
- remove debugging/personal vscode changes
- don't check-in commented out code - just delete those lines
- I need more information on why you are checking in the angularjs-autogrow dependency instead of using npm.

Thanks for your work on this @palmtreefrb !  I'm looking forward to trying this feature out!

src/js/lib/angularjs-autogrow.js
palmtreefrb: > I see that you checked-in the angularjs-autogrow library, presumably from https://github.com/evyros/angularjs-autogrow
> 
> This library is available as an npm library, but you've chosen to check it in. This works because we (unfortunately) still have a way to include checked-in-dependencies outside of webpack. Could you explain your rationale for choosing this method over the npm dependency, which would be preferred?
> 
> If we do end up checking it in, we should have a comment referencing where the code came from, as an internet citation or sorts. e.g. https://github.com/evyros/angularjs-autogrow

I made some changes to the library to allow passing in offset to the grow height.

docker/test-e2e/.vscode/launch.json
palmtreefrb: this have been reverted...

src/js/lib/angularjs-autogrow.js
palmtreefrb: I added a comment in the forked library.

rmunn (commented):
Normally the reason not to check dependencies into your Git repo is because then you won't be able to pull in changes easily. If you checked version 1.2 of the dependency into your repo, and then 1.3 and 1.4 come out with bugfixes and features that you need, you have a possibly-messy merge to do instead of simply changing "1.2" to "1.4" in your package.json.However, in the case of the angularjs-autogrow dependency, the last commit to [its GitHub repo](https://github.com/evyros/angularjs-autogrow) was over 4 years ago, so it's likely unmaintained. That means that 1) it's unlikely the repo owner will accept a PR to add our feature, and 2) it's unlikely the dependency will ever be updated. So the normal cons of checking a dependency into your own repo don't really apply here, and I'd be in favor of allowing him to do it. The angularjs-autogrow project is [MIT licensed](https://github.com/evyros/angularjs-autogrow/blob/dd1985defb32678074a0f6452f6413dfb83224ba/package.json#L21) so it's okay for us to do that. However, I would want to see two things:

1. Two different commits in the PR, one of which adds the angularjs-autogrow dependency as-is, and the second one which contains the code changes you made to that dependency. And,
1. The comment at the top of the angularjs-autogrow file in our repo should not only explain where it comes from, but also mention that its license is MIT so it's clear to anyone reading the code that we're allowed to do that, without their needing to follow the GitHub link and read the original project's package.json file.

With those two changes, I'd be okay with checking dependencies into our repo in this specific instance, precisely because the project appears unmaintained and is probably never going to release another version. As a general rule, though, it's better to fork the project and submit a PR, and if the project is slow about accepting the PR then you can use an `https://github.com/sillsdev/our-fork` URL in package.json instead of an NPM package version number.

megahirt (commented):
Thank you @palmtreefrb for your changes.  I agree with @rmunn 's review and would like to see those two changes before merging.  Thanks!

palmtreefrb:
Removed angularjs-autogrow and wrote my own "FitTextDirective" as a typescript directive. This will improve performance from not having to reference external npm dependencies every time docker container is built and npm install executed.

palmtreefrb:
Peek 2021-09-02 08-50

megahirt:
Looks great!  Thanks for taking the time to add the animation.

src/angular-app/languageforge/lexicon/editor/field/field.module.ts
```diff
@@ -28,7 +29,7 @@ export const EditorFieldModule = angular
     SoundModule,
     LexiconCoreModule,
     EditorCommentsModule,
-    AudioRecorder
+    AudioRecorder
```
rmunn, line 32: Tiny bit of extra end-of-line whitespace ended up here somehow.

rmunn (approved):
LGTM. One line in field.module.ts added some unneeded whitespace at the end of the line (line 32, containing "AudioRecorder"), but that doesn't have to be fixed for me to approve this. Nice work.

src/angular-app/languageforge/lexicon/shared/fit-text.directive.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export class FitTextDirective implements angular.IDirective {
+
+  link(scope: angular.IScope, element: angular.IAugmentedJQuery, attr: angular.IAttributes) {
+    let kInput: boolean = false;
+    function updateHeight(): any {
+      let height = element.prop('scrollHeight');
+      element.height = height;
+      element.css({ 'max-height': height + 'px' });
+      element.css({ height: height + 'px' });
```
megahirt, line 11: I'm unsure what the performance cost of executing element.css on every keystroke.  I wonder if we should compare the existing height to the scroll-height first, and only if they are different set them via element.css().

```suggestion
      if (element.height != height) {
        element.height = height;
        element.css({ 'max-height': height + 'px' });
        element.css({ height: height + 'px' });
      }
```

Also, is element.css dependent on JQuery ?  It kind of looks that way.  If so, is there a way to not use any JQuery in this directive?

src/angular-app/languageforge/lexicon/shared/fit-text.directive.ts
```diff
@@ -0,0 +1,33 @@
+import * as angular from 'angular';
+
+export class FitTextDirective implements angular.IDirective {
```
megahirt, line 3: You wouldn't have known this, but a while back we converted all of our directives to components in hopes that the code would be more compatible with a future upgrade to a new version of Angular.  I don't necessarily think that's a project goal anymore (upgrade to Angular) but I would prefer that we only write Components going forward instead of Directives.  Not a big deal, and since this code works - I'm fine with it as is.

megahirt (approved):
I made some comments and a suggestion.  You've done great work in digging into the code and making this change.  I approve of this PR and would like to get it in even if the suggestions aren't accepted.  They are optional at this point.  Thanks @palmtreefrb

src/angular-app/languageforge/lexicon/shared/fit-text.directive.ts
palmtreefrb: Attribute directives are not just an Angularjs concept and are still appropriate in [Angular ](https://angular.io/guide/attribute-directives) for manipulating dom elements. In this case changing it to a component would now also require changes to e2e. But It can be changed if you require.

megahirt:
Not required.  I would go ahead and merge.  Thanks for the link about
attribute directives in Angular.  I didn't know about that.

>

palmtreefrb:
Seems I still don't have elevated credentials to merge yet...

Screenshot from 2021-09-03 08-51-19

megahirt:
@palmtreefrb I was baffled by this until I looked closely at the error message.  This has to do with the branch protection rule that I had on develop that said "restrict branch pushes to the following people".  I have removed that from all of our protected branches.  Now that I've removed that, are you able to see the green merge button?

palmtreefrb:
That got it. I was able to merge.

PR #1100, opened 2021-08-24
Improve upgrade policy for DB

longrunningprocess:
LTOps recommended this since the underlying volume is RWO

### DIFF
--- a/docker/deployment/db-deployment.yaml
+++ b/docker/deployment/db-deployment.yaml
@@ -39,6 +39,12 @@ spec:
   selector:
     matchLabels:
       app: db
+  strategy:
+      # we don't ever want more than 1 "db" running at any time
+      type: RollingUpdate
+      rollingUpdate:
+        maxSurge: 0
+        maxUnavailable: 25%
   template:
     # https://kubernetes.io/docs/concepts/workloads/pods/#pod-templates
     metadata:


docker/deployment/db-deployment.yaml
```diff
@@ -39,6 +39,12 @@ spec:
   selector:
     matchLabels:
       app: db
+  strategy:
+      # we don't ever want more than 1 "db" running at any time since the pvc above is RWO
+      type: RollingUpdate
```
megahirt, line 44: My read of the site below suggests that `type: RollingUpdate` is not what we want.  Don't we want `type: Recreate` which forces old pods to terminate before starting the new one?

From https://www.bluematador.com/blog/kubernetes-deployments-rolling-update-configuration

> The most important option to configure rolling updates is the update strategy. In your Deployment manifest, spec.strategy.type has two possible values:
> - RollingUpdate: New pods are added gradually, and old pods are terminated gradually
> - Recreate: All old pods are terminated before any new pods are added
> In most cases, RollingUpdate is the preferable update strategy for Deployments. Recreate can be useful if you are running a pod as a singleton, and having a duplicate pod for even a few seconds is not acceptable.

docker/deployment/db-deployment.yaml
```diff
@@ -39,6 +39,12 @@ spec:
...
+      rollingUpdate:
+        maxSurge: 0
+        maxUnavailable: 25%
```
megahirt, line 47: These can be removed assuming that `type: Recreate` is what we want

megahirt (changes requested):
I'm pretty sure we want `type: Recreate`.  Thoughts?

docker/deployment/db-deployment.yaml
longrunningprocess: I appreciate all the research, we tested all three yesterday and were happiest with this one but I take your read on the `Recreate`.  It kills rather than stops...Greg and I were both a little hesitant about that when we saw similar behavior between it and the `RollingUpdate` the way we configured it.  We can dig further if you'd like but we went this way after testing all three.

docker/deployment/db-deployment.yaml
longrunningprocess: agreed, I would prefer this simpler config as well.

PR #1103, opened 2021-08-30
Release 1.8.8

longrunningprocess:
I think this is an administrative formality with our release coming tomorrow, please correct me if I'm misunderstanding anything.  I do expect we'll want a merge-commit from `staging` into `master` but I'm not totally confident whether we should rebase `staging` first and I'm a little surprised we won't actually have a release tag, `1.8.x` on this final commit (or sync commit)...maybe that's not a big deal, I'd like to chat about it if @megahirt has clear thoughts on this final piece of administration.

FYI: none of these changes will result in a different image, they've all been run against our prod environment already as well.


megahirt (commented):
I won't approve this PR just yet, because we actually need #1108 to be merged in first, which will become part of 1.8.8.  After that is merged, I would expect to see the Makefile update in this PR, at which point we can approve it.

longrunningprocess:
@megahirt please feel free to merge (merge-commit) this as part of the plans laid out in #1108.

PR #1138, opened 2021-09-07
Add default issue templates for bug, feature and s/r issues

megahirt:
I used the GitHub Issue template wizard to generate and customize these issue templates.  Most all of the lingo in these templates are stock GitHub, and they seem pretty good to me.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -26,13 +26,11 @@ If applicable, add screenshots to help explain your problem.
 **Desktop (please complete the following information):**
  - OS: [e.g. iOS]
  - Browser [e.g. chrome, safari]
- - Version [e.g. 22]
 
 **Mobile (please complete the following information):**
  - Device: [e.g. iPhone6]
  - OS: [e.g. iOS8.1]
  - Browser [e.g. stock browser, safari]
- - Version [e.g. 22]
 
 **Android App in Google Play**
 For Mobile, are you using the Language Forge Android app in the Google Play store?  [Yes/No]
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,41 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: 'bug: '
+labels: triage
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,41 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: 'bug: '
+labels: triage
+assignees: ''
```
longrunningprocess, line 6: would it be useful to go ahead and assign you or Alex here since you all are the initial "triagers"?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,41 @@
...
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**Desktop (please complete the following information):**
+ - OS: [e.g. iOS]
+ - Browser [e.g. chrome, safari]
+ - Version [e.g. 22]
```
longrunningprocess, line 29: do we care about the LF version here?

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: 'feat: '
+labels: triage
+assignees: ''
```
longrunningprocess, line 6: maybe this should be Alex?

longrunningprocess (approved):
some optional things from me for consideration, nothing blocking though

.github/ISSUE_TEMPLATE/bug_report.md
megahirt: I had considered that, however I opted simply for the *triage* label which anyone can filter for and then remove once the issue has been "accepted" by our team.

megahirt: thank you for bringing this up.  I decided not to auto-assign issues at this time.

.github/ISSUE_TEMPLATE/bug_report.md
megahirt: yeah, that's not necessary.  It was there from the standard GitHub template but I forgot to remove it.  Thanks for catching that.

.github/ISSUE_TEMPLATE/feature_request.md
megahirt: see my reply above

.github/ISSUE_TEMPLATE/bug_report.md
megahirt: In the future if we had two versions (e.g. beta.languageforge.org) and the live site, then we could ask which version of the site they were using.

.github/ISSUE_TEMPLATE/bug_report.md
longrunningprocess: 👍🏻

.github/ISSUE_TEMPLATE/bug_report.md
longrunningprocess: ok, sounds good.

Issue #1016, opened 2021-07-23
Go directly to project on sign-in


megahirt:
Upon sign in, if there is only one project in the user's profile, go straight to that project.

Don't show the user the project selection screen for non-admin users.

- [x] If the user is part of multiple projects, go to the project that was last visited/used.
- [x] If the user has zero projects, take them to the projects screen with the "create project" button.
- [ ] Remove the breadcrumb link to "My Projects".  Don't need it anymore.
- [ ] If the user navigates (how?) to the /app/projects page, and they have at least project, redirect to that project.
- [ ] The project page should still function as an admin-only tool for listing projects

alex-larkin:
Great ideas! How are users with multiple projects going to switch between them? I'm having a hard time visualizing it. Thanks.

megahirt:
Great question, @alex-larkin .  Language Forge already has a compact project picker menu in the upper right hand corner.  Here's a screenshot.  Perhaps this picker would need to be promoted so users know about it.

![image](https://user-images.githubusercontent.com/3444521/127518444-8451b589-b8d3-4077-9679-924d07329c5b.png)

alex-larkin:
Ah okay, looks good! I didn't remember that there were two "my projects" links- the breadcrumb one and this one.

rmunn:
As I've read the code and looked into how to implement this, I've discovered that there are lots of places in the code that have a hardcoded redirect to `/app/projects` in various situations: after logging in, after signing up on the website, after clicking on a signup link in an invitation email, and so on. I think the simplest implementation of this feature is going to be a rewrite of the /app/projects Angular component so that instead of displaying a project list, it simply figures out which project to redirect to (based on the logic Chris outlined in the issue description), and then uses Angular's `$location` service to redirect to the appropriate project page.

Also, there's a rare case where it might be impossible to choose a project to redirect to, when the user belongs to multiple projects, but the most-recently-used project no longer exists on the website so we can't redirect to that one. In that rare case we'll still display the projects list as before, and once the user chooses a project that one will become their most-recently-used project, and they'll no longer see the projects list. (E.g., the user used to belong to three projects, and their most-recently-used project was project A. Project A's manager just deleted project A while the user was signed out. Now the user signs in, and since we can't choose which of projects B or C to redirect to, we present the projects list and let the user click on either B or C.)

rmunn:
How did this get moved to the "Done" column? It's not done at all. Moving it back.

rmunn:
[...]

PR #1140, opened 2021-09-08
Go directly to project on sign-in

rmunn:
## Summary
Upon sign in, if there is only one project in the user's profile, go straight to that project.

Don't show the user the project selection screen.

- [X] If the user is part of multiple projects, go to the project that was last visited/used.
- [X] If the user has zero projects, take them to the projects screen with the "create project" button.

Original design had two more features, which I ended up choosing not to implement:

- [ ] Remove the breadcrumb link to "My Projects".  Don't need it anymore.
- [ ] If the user navigates to the /app/projects page, and they have at least project, redirect to that project.

The reason for this is that the projects list can actually be useful to people with multiple projects, and some of them have probably gotten used to clicking on the breadcrumb to navigate back to the projects list. So I decided that if they click on the breadcrumb or the "List All Projects" menu, they probably want to see the projects list in order to switch to another project, and decided to leave that behavior as-is so that users who like it won't have it taken away.

## Testing

- [ ] User with one project always goes directly to that project
- [ ] User with two projects is always taken to most recently viewed project
- [ ] User is removed from most recently viewed project - taken to project page


rmunn:
Converting to draft since we have a new design: https://github.com/sillsdev/web-languageforge/issues/1016#issuecomment-915640409

I'll update the top-level comment of this PR once I've implemented the new design.

rmunn:
That one failing E2E test will be fixed by https://github.com/sillsdev/web-languageforge/pull/1158.

src/Site/Controller/Page.php
```diff
@@ -29,7 +29,7 @@ public function view(
         // special case for "brochure" HTML5 homepage
         if ($pageName == 'home') {
             if ($this->_isLoggedIn) {
-                return $app->redirect('/app/projects');
+                return $app->redirect('/redirect/project');
```
longrunningprocess, line 32: it's a secondary point to this but I think I'd remove this altogether...if someone wants to get to the homepage, they should be able to get there...logging in isn't necessarily a big problem, most people will probably already be in the app and get redirected to the login page anyway.

src/Site/Controller/Redirect.php
```diff
@@ -0,0 +1,85 @@
+<?php
```
longrunningprocess, line 1: yikes man...the complexity is off the chart here...maybe there are parts here that actually belong in the UI or need to be resolved outside of this controller...not sure, I can just see there is way too much complexity in here.

longrunningprocess (commented):
I'm going to leave the approval of this to @megahirt since you all already hashed this out in person, there's just no way I can approve of what I'm seeing here without understanding the tradeoff's more.

src/Site/Controller/Page.php
megahirt: I see your point, @longrunningprocess .  I'm sure there are some websites that allow the user to access the brochure homepage while logged in, but there are a good amount of web applications that I can think of that don't either.

Some examples:

**Gmail** mail.google.com redirects to the brochure if you are logged out, and redirects to the app if you are logged in.
image

**PayPal** paypal.com redirects to a homepage if you are logged out, and takes you to the app if you are logged in.
image

In both of these cases, the user can't click to get to these pages directly once they are logged in, however they might be able to access them directly from the URL.

A compromise might be to make a valid route called /home which shows the brochure homepage no matter what.  Then have logic something like the above that if you are logged in, you go to the app, and if not you go to the `/home` page.

src/Site/Controller/Redirect.php
```diff
@@ -0,0 +1,85 @@
...
+                    }
+                } catch (\Exception $e) {
+                    return $app->redirect('/app/projects');
+                }
+                if ($project && $project->appName) {
+                    $projectId = $project->id->asString();
+                    return $app->redirect("/app/$project->appName/$projectId");
+                }
+            } else if (count($this->_user->projects->refs) == 0) {
+                return $app->redirect('/app/lexicon/new-project');
```
megahirt, line 80: YES!

src/Site/Controller/Redirect.php
```diff
@@ -0,0 +1,85 @@
...
+                $project = new ProjectModel($model->projectId);
+                if ($project && $project->appName) {
+                    return $app->redirect("/app/$project->appName/$model->projectId");
+                }
+            } catch (\Exception $e) {
+                return $app->redirect('/auth/logout');
+            }
+        }
+        if ($this->_user && $this->_user->projects) {
+            if (count($this->_user->projects->refs) == 1) {
```
megahirt, line 61: Where is the code for `if (count($this->_user->projects->refs) > 1)` ?

I expected to see a block of code for when the user has multiple projects.  In this case we should redirect to `$this->_user->lastUsedProjectId` .  I think we can keep this controller simple and let the `/app/lexicon` controller handle the case where somehow the user doesn't have permission on the project anymore (or it doesn't exist).

src/index.php
```diff
@@ -214,6 +214,8 @@
 $app->get('/app/{appName}/{projectId}',     'Site\Controller\App::view');
 $app->get('/app/{appName}/',    'Site\Controller\App::view')->value('appName', 'projects');
 $app->get('/app/{appName}',     'Site\Controller\App::view')->value('appName', 'projects');
+$app->get('/redirect/{appName}/',    'Site\Controller\Redirect::view')->value('appName', 'project');
+$app->get('/redirect/{appName}',     'Site\Controller\Redirect::view')->value('appName', 'project');
```
megahirt, line 218: Can we get this route/controller working like this?

```suggestion
$app->get('/redirect/project/',    'Site\Controller\Redirect::view');
$app->get('/redirect/project',     'Site\Controller\Redirect::view');
```

src/Site/Controller/Redirect.php
```diff
@@ -0,0 +1,85 @@
...
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+
+class Redirect extends App
+{
+    /**
+     * @param Request $request
+     * @param Application $app
+     * @param $appName
+     * @param string $projectId
```
megahirt, line 19:
```suggestion
```

megahirt (changes requested):
@rmunn in general this is good work.  I feel like maybe there's a chance I'm not understanding what your code is trying to do, but from my reading we don't need the first half of what you wrote in the controller.

Secondly, in #1016 you checked off
- [x] If the user is part of multiple projects, go to the project that was last visited/used.

however I didn't find any code that implements that, so I'd like some clarification on that as well.

Thanks again for tackling this.  Not easy code to work in.

src/Site/Controller/Redirect.php
rmunn: It's the final fallthrough case. I've pushed a new commit with comments explaining the redirects, to make the logic a bit easier to understand. The general principle that the logic should be understandable by just reading code, and that comments are superfluous, doesn't apply here as the reasoning behind the redirects isn't visible in the code.

src/index.php
rmunn: Done.

src/Site/Controller/Page.php
rmunn: I've left this as-is; we do want someone to be able to go to languageforge.org and be taken directly to their project if their login cookie is still valid.

src/Site/Controller/Redirect.php
rmunn: I've added some comments to explain the various redirect decisions, and reduced the complexity by getting rid of the AppModel instance. There are still a lot of if statements because there are multiple cases that need to be handled here:

* User logs in, but their most recent project was deleted by manager
* User logs in, but they have been removed from their most recent project
* User has an old session cookie with a most recent project ID in it, but their login has expired

And so on. But hopefully with the comments in my most recent commit the logic is comprehensible now.

rmunn:
> Secondly, in #1016 you checked off
> 
>     * [x]  If the user is part of multiple projects, go to the project that was last visited/used.
> 
> 
> however I didn't find any code that implements that, so I'd like some clarification on that as well.

That's part of the `requireValidProjectIdForThisWebsiteAndValidateUserMembership` call, which looks up the most recent project ID from the session cookie and, if it's not found in the session cookie, looks it up from the user's record in MongoDB.

src/Site/Controller/Redirect.php
```diff
@@ -0,0 +1,85 @@
...
+        if ($model->projectId) {
+            try {
+                $project = new ProjectModel($model->projectId);
+                if ($project && $project->appName) {
+                    return $app->redirect("/app/$project->appName/$model->projectId");
+                }
+            } catch (\Exception $e) {
+                return $app->redirect('/auth/logout');
+            }
+        }
```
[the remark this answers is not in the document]
rmunn: You're right that AppModel can be gotten rid of, and the projectID parameter is misleading (I never actually handle `/redirect/project/projectID`) so I got rid of it too. (It was only there because I copied App::view to begin with, and then didn't quite realize that I could trim that parameter off).

The `if ($model->requireProject)` branch can also be simplified, because we're forcing the project to be `lexicon` and requireProject is always true for a lexicon project. And the `if ($model->isPublicApp)` part can go, because lexicon projects are not public apps. But the requireValidProjectIdForThisWebsiteAndValidateUserMembership call (and boy, is that a "moutfhul" of a function name) has to stay since that's where the lookup of the most recent project ID happens.

The new commit that I just pushed implements most of these suggestions, so that now the complexity should be manageable.

src/Site/Controller/Redirect.php
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * @param Request $request
+     * @param Application $app
+     * @param $appName
+     * @param string $projectId
+     * @return Response
+     */
+    public function view(
+        /** @noinspection PhpUnusedParameterInspection */
+        Request $request, Application $app, $appName, $projectId = ''
```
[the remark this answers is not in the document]
rmunn: Done.

src/Site/Controller/Redirect.php
rmunn: Done.

megahirt (approved):
Nice work documenting all of those lines to clarify each logic step.  Thanks also for clarifying that requireValidProjectId... uses the lastUsedProjectId in Mongo

PR #1157, opened 2021-09-16
Add page-up page-down to editor component

palmtreefrb:
## Description

Added keyboard hooks, Page Up and Page Down to allow paging through the editor component.

### Type of Change
Feature request.

- UI change

## Screenshots

Peek 2021-09-16 16-22

## How Has This Been Tested?

Please describe the manual tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] Press Page Down to move to the next record
- [x] Press Page Up to move to the previous record
- [x] Press Page Down to move to the next record after reaching end
- [x] Press Page Up to move to the previous record after reaching end

## Checklist:

### DIFF
--- a/src/angular-app/languageforge/lexicon/editor/editor.component.ts
+++ b/src/angular-app/languageforge/lexicon/editor/editor.component.ts
@@ -150,7 +150,6 @@ export class LexiconEditorController implements angular.IController {
       }
       // destroy listeners when leaving editor page
       angular.element(window).unbind('keyup', (e: Event) => {});
-      this.$scope.$destroy();
     };
 
     this.show.entryListModifiers = !(this.$window.localStorage.getItem('viewFilter') == null ||
--- a/src/angular-app/languageforge/lexicon/editor/editor.component.ts
+++ b/src/angular-app/languageforge/lexicon/editor/editor.component.ts
@@ -1,26 +1,26 @@
 import * as angular from 'angular';
 
-import {ActivityService} from '../../../bellows/core/api/activity.service';
-import {ApplicationHeaderService} from '../../../bellows/core/application-header.service';
-import {ModalService} from '../../../bellows/core/modal/modal.service';
-import {NoticeService} from '../../../bellows/core/notice/notice.service';
+import { ActivityService } from '../../../bellows/core/api/activity.service';
+import { ApplicationHeaderService } from '../../../bellows/core/application-header.service';
+import { ModalService } from '../../../bellows/core/modal/modal.service';
+import { NoticeService } from '../../../bellows/core/notice/notice.service';
 import {
   EditorDataService,
   FilterOption,
   LabeledOption,
   SortOption
 } from '../../../bellows/core/offline/editor-data.service';
-import {LexiconCommentService} from '../../../bellows/core/offline/lexicon-comments.service';
-import {SessionService} from '../../../bellows/core/session.service';
-import {InterfaceConfig} from '../../../bellows/shared/model/interface-config.model';
-import {SemanticDomainsService} from '../../core/semantic-domains/semantic-domains.service';
-import {LexiconEntryApiService} from '../core/lexicon-entry-api.service';
-import {LexiconProjectService} from '../core/lexicon-project.service';
-import {LexiconRightsService, Rights} from '../core/lexicon-rights.service';
-import {LexiconSendReceiveService} from '../core/lexicon-send-receive.service';
-import {LexiconUtilityService} from '../core/lexicon-utility.service';
-import {LexEntry} from '../shared/model/lex-entry.model';
-import {LexPicture} from '../shared/model/lex-picture.model';
+import { LexiconCommentService } from '../../../bellows/core/offline/lexicon-comments.service';
+import { SessionService } from '../../../bellows/core/session.service';
+import { InterfaceConfig } from '../../../bellows/shared/model/interface-config.model';
+import { SemanticDomainsService } from '../../core/semantic-domains/semantic-domains.service';
+import { LexiconEntryApiService } from '../core/lexicon-entry-api.service';
+import { LexiconProjectService } from '../core/lexicon-project.service';
+import { LexiconRightsService, Rights } from '../core/lexicon-rights.service';
+import { LexiconSendReceiveService } from '../core/lexicon-send-receive.service';
+import { LexiconUtilityService } from '../core/lexicon-utility.service';
+import { LexEntry } from '../shared/model/lex-entry.model';
+import { LexPicture } from '../shared/model/lex-picture.model';
 import {
   LexConfig,
   LexConfigField,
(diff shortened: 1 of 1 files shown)


palmtreefrb:
Screenshot from 2021-09-15 17-47-46

src/angular-app/languageforge/lexicon/editor/editor.component.ts
```diff
@@ -87,26 +87,40 @@ export class LexiconEditorController implements angular.IController {
...
+    private readonly commentService: LexiconCommentService,
+    private readonly editorService: EditorDataService,
+    private readonly lexService: LexiconEntryApiService,
+    private readonly lexProjectService: LexiconProjectService,
+    private readonly rightsService: LexiconRightsService,
+    private readonly sendReceive: LexiconSendReceiveService,
+  ) { }
 
   $onInit(): void {
+    angular.element(window).bind('keyup', (e: Event) => {
```
longrunningprocess, line 110: I'm wondering whether `onInit` is the right lifecycle hook to set this up or not?  If that's convention, then ok, however I would probably be more inclined to set the listener up after the data is back and we actually realize we have a need for the ability to move around.  Additionally, will this clean-up the listeners after the user leaves this page?

longrunningprocess (commented):
Some of the template fields need to be updated before merging.

src/angular-app/languageforge/lexicon/editor/editor.component.ts
palmtreefrb: In Angularjs onInit will occur after all elements have been constructed and bound.

megahirt: Per our discussion, binding to `window` is acceptable.  However, we need to unregister the event listener on the $onDestroy() lifecycle hook.

megahirt: There are console errors that need to be addressed before taking a look at this again.

The console errors have to do with calling the method in a context that it doesn't expect.

I will approve with the current approach of $onInit / $onDestroy

However, an alternative approach could be to use the $on('$locationChangeStart') with an eye to registering the handler when the location is changed to /editor/entry and unregister the handler when the location is changed to something else.

megahirt (changes requested):
There are some console errors that need fixing, per our standup discussion

megahirt:
@palmtreefrb please also edit the PR template, specifically filling out the summary description, type of change and then deleting the screenshots section since there aren't any

src/angular-app/languageforge/lexicon/editor/editor.component.ts
longrunningprocess: > However, an alternative approach could be to use the $on('$locationChangeStart') with an eye to registering the handler when the location is changed to /editor/entry and unregister the handler when the location is changed to something else.

💡  I get it now...this is a trip wire for the editor where pg-up/pg-dn are relevant.  It's different than relying on the presence/absence of a particular DOM element.  If I'm understanding this properly, this does sound perfect!

palmtreefrb:
Narrowed the scope when loading the listener using viewContentLoaded and added unload listener and scope destroy to the existing $window.onbeforeunload method.

src/angular-app/languageforge/lexicon/editor/editor.component.ts
```diff
@@ -128,6 +148,9 @@ export class LexiconEditorController implements angular.IController {
       if (this.hasUnsavedChanges()) {
         this.saveCurrentEntry();
       }
+      // destroy listeners when leaving editor page
+      angular.element(window).unbind('keyup', (e: Event) => {});
+      this.$scope.$destroy();
```
megahirt, line 153: Unfortunately `$window.onbeforeunload()` is what is called when a tab or windows is closed.  So this code doesn't do anything helpful before the event handler will be destroyed as the window/tab is closed anyway.

`this.$scope.$destroy();` seems unnecessary to me.  AngularJS should be calling this automatically and we shouldn't need to explicitly call it.

Keeping with the $onInit() lifecycle hook, we should use the complementary `$onDestroy()` hook to clean up the event listener.  So the `element.unbind()` should be done inside the `$onDestoy()`.

megahirt (dismissed):

src/angular-app/languageforge/lexicon/editor/editor.component.ts
palmtreefrb: I did look at $onDestroy() and it does not get called on just leaving and navigating to another page. I thought we discussed not having the key handler listening after leaving the editor page. this.$scope.$destroy(); was just an assurance everything was cleaned up. I can remove it.

longrunningprocess: @palmtreefrb I thought Chris' idea about the locationChange was pretty good, did you find it not to be a good option?

palmtreefrb: Not a good option because it doesn't get called on page load. You have to click on a navigation button to execute the event.

longrunningprocess: ok, yeah that makes sense.  Thank you.

src/angular-app/languageforge/lexicon/editor/editor.component.ts
```diff
@@ -128,6 +148,8 @@ export class LexiconEditorController implements angular.IController {
       if (this.hasUnsavedChanges()) {
         this.saveCurrentEntry();
       }
+      // destroy listeners when leaving editor page
```
longrunningprocess, line 151: I still don't think this will fire if someone is just navigating around...unless this event is handled specially by Angular, it would only apply for a full page refresh, not a location change in a SPA.

longrunningprocess (changes requested):
unless my reading is wrong, this will not give you what you actually want.

src/angular-app/languageforge/lexicon/editor/editor.component.ts
palmtreefrb: Pull it down and add a brakepoint there. It does actually give the behavior I was intending.

longrunningprocess: ok, good enough for me.  I'll approve.  Thanks.

longrunningprocess (approved):

longrunningprocess:
@palmtreefrb unless you want Robin and/or Chris as well, you can remove them and merge.

src/angular-app/languageforge/lexicon/editor/editor.component.ts
megahirt: that's too bad that `$locationChangeStart` doesn't get fired on first page load :(

palmtreefrb (commented):
Can I merge this now? Merge is still blocked and waiting for approval.

longrunningprocess:
since you've requested Chris to approve and the last thing he did was request changes, you either need to remove him from the reviewers or make the changes he requested (if they're still relevant)

src/angular-app/languageforge/lexicon/editor/editor.component.ts
palmtreefrb: > since you've requested Chris to approve and the last thing he did was request changes, you either need to remove him from the reviewers or make the changes he requested (if they're still relevant)

Not an option I can find to clear this up to complete the merge. Maybe my user level does not have these options.

rmunn:
@longrunningprocess wrote:
> since you've requested Chris to approve and the last thing he did was request changes, you either need to remove him from the reviewers or make the changes he requested (if they're still relevant)

I don't think someone who is *not* a repo admin can remove someone else's "change requested" review.

longrunningprocess:
ok then, based upon @megahirt 's WhatsApp stand-up message, I'm going to go ahead and get this one in.

PR #1159, opened 2021-09-16
Fixed `make dev` for local CSS development

longrunningprocess:
## Description

`make dev` was always meant to allow developers to make changes locally and see those changes reflected in his/her running app.  However, CSS development was overlooked in the initial creation of the infrastructure, i.e., local CSS changes would not show up in the locally running app unless the dev cleaned volumes and rebuilt the ui image from scratch.

Fixes an untracked issue

### Type of Change

## Screenshots

N/A

## How Has This Been Tested?

- [x] change an `.scss` file under `src/sass/`
- [x] change an `.scss` file under `src/Site/views/languageforge/theme/default/sass/`
- [x] change an `.scss` file under `src/angular-app/`

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "build:dev:watch": "npm-run-all -p -l sass:watch webpack:dev:watch",
     "build:prd": "npm-run-all -p -l sass webpack:prd",
     "sass": "sass src/sass:src/dist/css",
-    "sass:watch": "npm run sass && sass src/sass:src/dist/css --watch --recursive",
+    "sass:watch": "sass --watch src/sass:src/dist/css src/Site/views/languageforge/theme/default/sass:src/dist/css src/angular-app:src/dist/css",
     "webpack:dev": "webpack --config webpack-dev.config.js",
     "webpack:dev:watch": "webpack -w --config webpack-dev.config.js",
     "webpack:prd": "webpack --config webpack-prd.config.js",
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -19,8 +19,10 @@ services:
       - ../package.json:/data/package.json
       - ../package-lock.json:/data/package-lock.json
 
-      # needed this volume mapping so changes to typescript would be reflected in running app, actually rebundled and output to dist which is then shared to the app container.
+      # needed these volume maps so changes to typescript/scss would be reflected in running app, actually rebundled and output to dist which is then shared to the app container.
       - ../src/angular-app:/data/src/angular-app
+      - ../src/sass:/data/src/sass
+      - ../src/Site/views/languageforge/theme/default/sass:/data/src/Site/views/languageforge/theme/default/sass
 
   app:
     build:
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
     "build:dev:watch": "npm-run-all -p -l sass:watch webpack:dev:watch",
     "build:prd": "npm-run-all -p -l sass webpack:prd",
     "sass": "sass src/sass:src/dist/css",
-    "sass:watch": "npm run sass && sass src/sass:src/dist/css --watch --recursive",
+    "sass:watch": "sass --watch src/sass:src/dist/css src/Site/views/languageforge/theme/default/sass:src/dist/css src/angular-app:src/dist/css",
```
megahirt, line 11: I am trying to figure out why the extra path mappings are needed.  My thinking is there shouldn't be a need for the second and third path mappings that you've included on the command line, since they are included as import statements in [`src/sass/languageforge.scss`](https://github.com/sillsdev/web-languageforge/blob/develop/src/sass/languageforge.scss#L1) 

First line in `languageforge.scss`:
```
@import "../Site/views/languageforge/theme/default/sass/languageforge_default.scss";
```
The other sass files in angular-app are explicitly imported one line at a time.

It seems fine to remove the `--recursive` since that doesn't actually seem necessary, as well as the redundant `npm run sass`.  Thanks for fixing that up.

```suggestion
    "sass:watch": "sass --watch src/sass:src/dist/css",
```

Shouldn't it work with the simplified suggestion above?  I haven't tested it.

This comment originally appeared over at https://github.com/sillsdev/web-languageforge/pull/1160#discussion_r712823597 until I realized it should actually be on this PR.  Sorry for the confusion.

megahirt (approved):
I approve of this PR, @longrunningprocess especially since you've waited unnecessarily long for a review :( I have a non-blocking suggestion regarding extra path mappings in sass compilation. Obviously it works as is since you tested it, but it seems there are some redundant paths being compiled.

package.json
longrunningprocess: > I am trying to figure out why the extra path mappings are needed. My thinking is there shouldn't be a need for the second and third path mappings that you've included on the command line, since they are included as import statements in src/sass/languageforge.scss

`sass` only watches changes in a directory (or a file)...not, "watch this file and any of its dependencies, i.e., anything it `@import`s as well".  See https://sass-lang.com/documentation/cli/dart-sass.  Additionally, `@import` is deprecated anyway...typically you'd see something like, `@use '.../some_partial_scss'`.  I didn't take the time to test whether having the newer syntax in there would've helped the watch work more simply but it's always something we could come back to if it's too bothersome in this current state.  Just keep in mind, migrating to`~@use` will require some additional changes to accommodate the namespacing that comes in with the newer syntax.

> It seems fine to remove the --recursive since that doesn't actually seem necessary
That option was available in the previous `node-sass` but is not an option for `sass`

megahirt: Thank you for the explanation @longrunningprocess .  I didn't know about upgrading `@import` statements to `@use` statements.  It also makes sense that the sass watch needs to have different arguments than the vanilla compile step since the watch parameters are for what *files* it should watch, not which files are entry points for compilation.

PR #1160, opened 2021-09-16
Allow dev to test locally running app from any device on same network

longrunningprocess:
## Description

There are times when a dev needs to access his/her locally running app from another device on the same network.  Before these changes, this was not really feasibly with our reverse-proxy configuration.

Fixes an untracked issue

### Type of Change

## Screenshots

N/A

## How Has This Been Tested?

- [x] I ran the app locally and accessed it from my phone via WiFi

## Checklist:

- [x] I have rebased off `develop` after #1159 has been merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,8 @@ Other useful resources:
 > ```
 > 
 > then hit `http://192.168.161.99` from your phone or other device on the same network.
+>
+> NOTE: disabling cache on your device may not be trivial, you'll either need to wipe the site settings on your device's browser or you'll need some do it via USB debugging.
 
 ### Running E2E Tests
 
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ Other useful resources:
 > 
 > then hit `http://192.168.161.99` from your phone or other device on the same network.
 >
-> NOTE: disabling cache on your device may not be trivial, you'll either need to wipe the site settings on your device's browser or you'll need some do it via USB debugging.
+> NOTE: disabling cache on your device may not be trivial, you'll either need to wipe the site settings on your device's browser or you'll need to do it via USB debugging.
 
 ### Running E2E Tests
 
--- a/README.md
+++ b/README.md
@@ -82,6 +82,18 @@ Other useful resources:
 1. You should see a landing page, click "Login"
 1. Use `admin` and `password` to get in
 
+> Sometimes there may be a need to hit the locally running app from a device other than the machine the app is running on.  In order to do that, you'll need to do the following:
+> 1. Figure out your local ip address
+> 1. Access the app via http at that address
+> 
+> On a Mac for example:
+> ```
+> ifconfig | grep broadcast
+> 	inet 192.168.161.99 netmask 0xfffffc00 broadcast 192.168.163.255
+> ```
+> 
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
     "build:dev:watch": "npm-run-all -p -l sass:watch webpack:dev:watch",
     "build:prd": "npm-run-all -p -l sass webpack:prd",
     "sass": "sass src/sass:src/dist/css",
-    "sass:watch": "npm run sass && sass src/sass:src/dist/css --watch --recursive",
+    "sass:watch": "sass --watch src/sass:src/dist/css src/Site/views/languageforge/theme/default/sass:src/dist/css src/angular-app:src/dist/css",
```
megahirt, line 11: I am trying to figure out why the extra path mappings are needed.  I didn't see any comment in the PR about changes to sass compilation and my thinking is there shouldn't be a need for the second and third path mappings that you've included on the command line, since they are included as import statements in [`src/sass/languageforge.scss`](https://github.com/sillsdev/web-languageforge/blob/develop/src/sass/languageforge.scss#L1) 

First line in `languageforge.scss`:
```
@import "../Site/views/languageforge/theme/default/sass/languageforge_default.scss";
```
The other sass files in angular-app are explicitly imported one line at a time.

It seems fine to remove the `--recursive` since that doesn't actually seem necessary, as well as the redundant `npm run sass`.  Thanks for fixing that up.

```suggestion
    "sass:watch": "sass --watch src/sass:src/dist/css",
```

Shouldn't it work with the simplified suggestion above?  I haven't tested it.

megahirt (approved):
I approve of this PR, @longrunningprocess especially since you've waited unnecessarily long for a review :)  I have a non-blocking suggestion regarding extra path mappings in sass compilation.  Obviously it works as is since you tested it, but it seems there are some redundant paths being compiled.

package.json
megahirt: sorry - please disregard this suggestion as it refers to #1159.

Unfortunately I reviewed this PR before #1159 and didn't realize that this PR included unrelated changes.  In the future please ensure PR change sets include only what is intended and not other PRs.  This requires a bit of git foo to arrange branches in certain ways.  There are some cases in which PRs must depend upon changes, however in this case these changes were unrelated (caddy vs sass) and could have been separated.

longrunningprocess:
> I approve of this PR, @longrunningprocess especially since you've waited unnecessarily long for a review :(
> 
> Adding the feature to open port 80 on the network for the developer machine seems like a good idea for the local device testing scenario. I was wondering if we should make it more clear somehow in our docks that running docker compose will allow anyone on your network to access your machine on port 80 - in some ways this might fit better as an "opt in" feature rather than default to open port. Thoughts?

Yeah, I like your thinking for the "opt-in" approach.  The options coming immediately to my mind would be a little cumbersome but it might be worthwhile to pursue more later.  As for adding a warning to the docs, I'd be happy to do that if I thought anyone would read them 😆 .   I'm open to your suggestions on where to put that and how to call it out though.

package.json
longrunningprocess: yep, you're right.  Taking the PR's in order is probably another good way to avoid the confusion.  Thanks.

PR #1167, opened 2021-09-22
Silence Dart Sass warnings with -q

### LINKED ISSUE #1152: Deprecation warnings from sass
**Describe the bug**
Sass is undergoing some deprecations and those warnings are starting to show in our UI builds now.

**To Reproduce**
Doing anything that requires the `ui-builder` be built will reflect the warnings.

**Expected behavior**
We'd really like to keep the noise down on our builds.

**Additional context**
many of the following:

```sh
[sass       ] DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.
[sass       ]
[sass       ] Recommendation: math.div($spacer, 2)
...
...
```

megahirt:
## Description

The current sass deprecation warnings are related to our version of bootstrap, and we don't know how much work it would be to upgrade, so I decided the easiest thing to move forward is to silence those warnings.

This PR adds the `-q` flag to the `sass` command to suppress all warning messages.

Fixes #1152

### Type of Change

## Screenshots

Before this change:
image

After this change:
image

## Checklist:


rmunn:
As I mentioned in https://github.com/sillsdev/web-languageforge/issues/1152#issuecomment-924711958, we'll want to make a note somewhere to not upgrade to Sass 2 while these warnings are silenced, and we can't do so in package.json.

rmunn (approved):

Issue #1172, opened 2021-09-24
bug: textarea resize fixes


megahirt:
**Describe the bug**
Textarea resize fixes:
- border color should be grey, the same as what is on languageforge.org
- when text lines are deleted, resizing doesn't appear to work properly.  So if you delete two lines the space persists, and will slowly shrink if you tap arrow keys back and forth

**Expected behavior**
The new auto-sizing text areas should look identical to what is already shipped.  No color or shape difference.

Text areas should shrink to the proper height when deleting lines.  If you delete all content the box should look identical to how it started with no content.

**Screenshots**

Screenshot on staging, demonstrating black outline and textarea that does not shrink when content is removed.
![image](https://user-images.githubusercontent.com/3444521/134749372-384d9edc-adf3-40b0-a0af-b54da9dccfae.png)

Screenshot on prod, showing original grey outline color
![image](https://user-images.githubusercontent.com/3444521/134749448-67aa1f05-a2f0-40a3-b1ca-1cb0ebb9cf6c.png)

**Additional context**
applies to current staging environment

megahirt:
shipped in v1.9

PR #1185, opened 2021-09-30
remove black border on dc-text textarea

megahirt:
## Description

This changes the textarea border from black to grey to be consistent with all other inputs

See #1172 

### Type of Change

## Screenshots

Unwanted black borders before this change
image

Grey borders after this change
image

## Checklist:


longrunningprocess (approved):
great catch, that is much better

PR #1199, opened 2021-10-03
reduce vertical padding/margin on form controls in edit view

megahirt:
## Description

The intent of reducing vertical padding/margin on edit controls is to increase the usable space in the editor which increases information density for the user.

With this change, the user can see more fields in the entry, as well as see more entries in the compact list on the left side.

I consider this to be an improvement over the existing margins, etc. but one could also argue against this change because it could also be harder for users with a touch screen to click on the right item/field.  To make these adjustments, I had to override the sensible default Bootstrap margins on form controls.

In the side list, this change increases the visible items from 11 to 13 (on my desktop).
In this example entry, this change increases the visible fields in the editor from 10 to 12 (on my desktop).

This PR is not about "redesigning the UI" but rather making small, incremental improvements that stakeholders agree is a move in the right direction.

### Type of Change
- UI change

## Screenshots

Before:

image

After:
image

## Checklist:

### DIFF
--- a/src/angular-app/languageforge/lexicon/editor/_editor.scss
+++ b/src/angular-app/languageforge/lexicon/editor/_editor.scss
@@ -323,6 +323,10 @@ dc-entry .card {
   #compactEntryListContainer {
     height: calc(100vh - 400px);
     overflow: auto;
+
+    .list-group-item {
+      padding: .25rem 1.25rem;
+    }
   }
   .word-definition-title {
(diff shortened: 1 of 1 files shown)


alex-larkin:
Looks good! I'll approve. Thanks Chris!

alex-larkin (approved):

longrunningprocess (approved):
lgtm, nice idea

PR #1200, opened 2021-10-03
Make rendered entry scrollable on mobile

### LINKED ISSUE #1112: Hide top entry paragraph while keyboard visible on small mobile devices
When editing a long entry on mobile, the top portion of the page that shows all of the entry's data takes up a disportionate amount of screen space, making it difficult to see what the user is editing.

I would recommend hiding that part of the page when the keyboard pops into view, on small mobile devices. 

See the attached screen recording:

https://user-images.githubusercontent.com/82162128/131802828-3f8285c3-a8b2-47c8-9877-2de752c8a172.mp4

Thanks!

megahirt:
## Description

Per discussion below, make the "rendered entry" on mobile scrollable with the entry editor.  The rendered-entry on desktop remains sticky at the top of the surface.

Fixes #1112 

### Type of Change

- UI change

## Video Demo

https://user-images.githubusercontent.com/3444521/137492485-5ec18636-be8c-4fcc-9418-2772f4e874ae.mp4

See how the desktop version keeps the rendered entry sticky with the toolbar on top.
See how the mobile version scrolls the rendered entry with the editor fields

## Checklist:

### DIFF
--- a/src/angular-app/languageforge/lexicon/editor/field/_dc-rendered.scss
+++ b/src/angular-app/languageforge/lexicon/editor/field/_dc-rendered.scss
@@ -26,7 +26,12 @@
   }
 }
 @include media-breakpoint-down(sm) {
-  .dc-rendered-container {
+  .dc-rendered-on-desktop {
+    display: none;
+  }
+}
+@include media-breakpoint-up(md) {
+  .dc-rendered-on-mobile {
     display: none;
   }
 }
--- a/src/angular-app/languageforge/lexicon/editor/field/_dc-rendered.scss
+++ b/src/angular-app/languageforge/lexicon/editor/field/_dc-rendered.scss
@@ -25,6 +25,11 @@
     font-style: normal;
   }
 }
+@include media-breakpoint-down(sm) {
+  .dc-rendered-container {
+    display: none;
+  }
+}
 .dc-rendered-overflow{
   display: -webkit-box;
   -webkit-line-clamp: 2;
--- a/src/angular-app/languageforge/lexicon/editor/editor-entry.view.html
+++ b/src/angular-app/languageforge/lexicon/editor/editor-entry.view.html
@@ -95,7 +95,7 @@
             <div id="scrolling-editor-container" class="col-md-7 col-lg-8 entry-primary-container container-scroll">
                 <div>
                     <div class="entryItemView" data-ng-if="$ctrl.entryLoaded()">
-                        <div id="editor-title-text" class="word-definition-title dc-rendered-on-mobile">
+                        <div class="word-definition-title dc-rendered-on-mobile">
                             <dc-rendered config="$ctrl.lecConfig.entry" global-config="$ctrl.lecConfig"
                                         model="$ctrl.currentEntry" option-lists="$ctrl.lecOptionLists"></dc-rendered>
                         </div>
--- a/src/angular-app/languageforge/lexicon/editor/editor-entry.view.html
+++ b/src/angular-app/languageforge/lexicon/editor/editor-entry.view.html
@@ -95,6 +95,10 @@
             <div id="scrolling-editor-container" class="col-md-7 col-lg-8 entry-primary-container container-scroll">
                 <div>
                     <div class="entryItemView" data-ng-if="$ctrl.entryLoaded()">
+                        <div id="editor-title-text" class="word-definition-title dc-rendered-on-mobile">
+                            <dc-rendered config="$ctrl.lecConfig.entry" global-config="$ctrl.lecConfig"
+                                        model="$ctrl.currentEntry" option-lists="$ctrl.lecOptionLists"></dc-rendered>
+                        </div>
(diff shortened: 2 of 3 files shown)


alex-larkin:
Thanks so much for your work on this. 

Just to be sure l understand correctly, does this change completely remove the rendered entry on mobile devices? 

I do think that the rendered entry is useful on mobile for referencing or skimming purposes. But when the user wants to switch to the task of editing, visibility of the text fields becomes the priority. Additionally, available screen space shrinks with the appearance of the keyboard. 

My original idea was to use keyboard visibility as a kind of indicator of what "mode" the user was in. 

Some quick googling indicated to me that while there is no direct way to determine if the keyboard is present, there are some [indirect methods](https://www.geeksforgeeks.org/how-to-detect-virtual-keyboard-using-javascript/). Perhaps we could use a resize listener. 

Granted, there are some other, possibly "edge", cases too. Maybe the user is in landscape mode, or maybe they are in split screen mode. Even so, in these cases I think it would be appropriate to hide the rendered entry. 

So now I'm thinking we shouldn't be thinking so much about keyboard visibility, but rather about amount of space available to the editing window. 

I guess, LF could run a calculation on page load and every time a resize event is triggered. If the editing window is less than, say, 600px high, hide the rendered entry window. If another resize entry is triggered, repeat the test, using the recorded rendered entry window height. If the test fails, show the rendered entry window and run the test one more time (in case the user added a bunch of data, making the rendered entry window bigger than it was before).  

What do you think?

megahirt:
This change hides the rendered entry on the edit screen on mobile.  It is
still visible on mobile if you go back to the list.

The article you linked to is helpful in thinking through our options.  I
would prefer a CSS solution to this issue if we need to make more changes,
but I first want to understand exactly what you're asking for.  Let's try
and talk about this when you're next available.

longrunningprocess:
The way I'm beginning to understand this request is that we would only hide the rendered entry if both the user is viewing on a mobile device **AND** the keyboard is open.  If that's the case, I believe we might accommodate the latter by simply firing an event when an input field gets focus.  I agree on having a design discussion on this one though.

src/angular-app/languageforge/lexicon/editor/field/_dc-rendered.scss
```diff
@@ -25,6 +25,11 @@
     font-style: normal;
   }
 }
+@include media-breakpoint-down(sm) {
+  .dc-rendered-container {
+    display: none;
```
longrunningprocess, line 30: this might need to be conditional...let's discuss further

longrunningprocess (dismissed):
Mind if we clarify this one before we approve?   I'd like to make sure we can get away with always hiding this on mobile or whether we need to only hide it when the keyboard is showing.  Thanks.

alex-larkin:
Thinking about this some more, I would like to know if the rendered view really needs to be in its own window (or is it called a "container"?) and always visible. Does the user always need to have it in view? 

Maybe the easiest solution is to simply put the rendered view in the same window as the text fields, on the top. When the user wants to skim or look at the data in a consolidated form, it's right there. When he wants to edit, he just scrolls down, and the rendered view slides out of sight. The user has maximum vertical space to edit, regardless of what device he is on, regardless of keyboard configuration. 

What do you all think?

longrunningprocess:
> Thinking about this some more, I would like to know if the rendered view really needs to be in its own window (or is it called a "container"?) and always visible. Does the user always need to have it in view?
> 
> Maybe the easiest solution is to simply put the rendered view in the same window as the text fields, on the top. When the user wants to skim or look at the data in a consolidated form, it's right there. When he wants to edit, he just scrolls down, and the rendered view slides out of sight. The user has maximum vertical space to edit, regardless of what device he is on, regardless of keyboard configuration.
> 
> What do you all think?

yeah, that's a great question, it certainly would be easier to put it in the same container as the input fields without all the conditional concern and just let it scroll out of view.  It's worth discussing for sure!

megahirt:
Great conversation, guys.  @alex-larkin here's the timeline on this:
- The rendered entry was originally scrollable on v1.7 (both mobile and desktop)
- v1.8 shipped a change to make the rendered entry along with the toolbar sticky at the top.  This was at my request (Rick implemented it) and is documented in #1004 .  I don't think we paid enough attention to the UX on mobile for this change, which is why you raised this issue.  I think the correct behavior is as you say: on mobile, make the rendered entry scrollable.  On desktop, my opinion is it makes sense to keep the rendered entry sticky since we have enough vertical space to display it (and we now truncate with ellipsis at 3 lines I believe).

longrunningprocess:
> If you are agreeable @alex-larkin and @longrunningprocess , I will modify this PR so that the rendered entry scrolls (not sticky on top) on mobile, as can be seen in v1.7. I will leave the desktop version unchanged.

I like this idea, you definitely have my support @megahirt

alex-larkin:
Ok, I just created #1210 to discuss the rendered entry use cases and design requirements.

My only concern is that if we ever decide to show the entire rendered entry without truncation on desktop, that for larger entries, editing space will be reduced. 

But if we keep the rendered entry truncated, it should be fine on desktop.

PR #1204, opened 2021-10-04
Delta updates to minimize edit conflicts

### LINKED ISSUE #1051: Entry updates should send diffs, not whole entries
Currently lexical entry updates are sent as whole entries which get saved to Mongo. This is simple, but can end up causing users to lose data due to race conditions if multiple people are editing the same entry at once. The race condition is rather simple to trigger, because there are two separate timers running when someone is editing a lexical entry:

* The `pollUpdateTimer` timer in LexiconSendReceiveService is running every 32 seconds (search for `POLL_UPDATE_INTERVAL` to find where that number is hardcoded). When it triggers, it calls an arbitrary callback that can be defined by other code. In the editor component, the callback does this:
  ```js
  private pollUpdateSuccess = (): void => {
    if (this.hasUnsavedChanges()) {
      if (this.sendReceive.isInProgress()) {
        this.cancelAutoSaveTimer();
        this.warnOfUnsavedEdits(this.currentEntry);
        this.resetEntryLists(this.currentEntry.id, angular.copy(this.pristineEntry));
      }
    } else {
      this.setCurrentEntry(this.entries[this.editorService.getIndexInList(this.currentEntry.id, this.entries)]);
    }
  }
  ```
* In the editor component, there is an autosave timer which runs every 5 seconds. It does exactly what you'd expect.

The intent of the `if (this.hasUnsavedChanges())` line is to prevent data loss, but there's a race condition. Person A and person B are editing the same entry, and both make a change. Say A edits a definition (adding a new writing system to an existing field) and B removes a gloss. Now if the timers go off as follows, then A's change can be lost:

* Time T: the `pollUpdateTimer` fires on A's client. Its next firing is set to go off at T+32 seconds.
* Time T+3: the `pollUpdateTimer` fires on B's client. Its next firing is set to go off at T+35 seconds.
* Time T+5: A finishes adding a new definition.
* Time T+10: The autosave timer goes off on A's clent, and A's new definition is saved to MongoDB.
* Time T+21: B finishes removing a gloss.
* Time T+26: The autosave timer goes off on B's clent, and B's gloss deletion is saved to MongoDB.
* Time T+32: the `pollUpdateTimer` fires on A's client... and A's new definition *disappears*!

Now, at time T+3, B's `pollUpdateTimer` went off and loaded the state of the entry as it was at that time, which did not include the new definition that A had written. So B's copy of the project had the definition in English, but not in Thai. Meanwhile, at time T+10 the MongoDB entry was updated with A's new definition, so that that point MongoDB had the English *and* Thai definitions available. But then at time T+26, B's autosave timer went off, causing B's entry to be saved to MongoDB. B's entry included a gloss deletion which was desired, but B's entry also included a definition in English but not Thai, because that's what was loaded from MongoDB at time T+3. So the state of the entry in Mongo becomes a copy of what B had on-screen: gloss removed, and definition lacking Thai. So then at T+32, A's client loads the current state of the entry from Mongo, and it includes B's gloss deletion but not A's Thai sees his entry update, and the definition he wrote 27 seconds ago has vanished!

What should happen is that updates need to be sent to the server as diffs, rather than as entire entries. It wouldn't work to send the entire entry and do the diffing server-side, because what would happen in the scenario above is that B's entry (which lacks a Thai definition) would be diffed against the MongoDB contents (which contain a Thai definition), and the diff that would be produced and applied would contain "gloss deleted, Thai definition deleted".

At some point in the future we'll move to ShareDB, which uses operational transforms to deal with this issue and can handle it for us. But for a solution that works *right now* without rewriting our backend, we'll need to change the client to send diffs rather than entire entries. The client will preserve a pristine copy of what is loaded from Mongo, and when the autosave is triggered, the current state of the entry will be compared against the pristine copy and a diff will be generated. (I suggest using [jsondiffpatch](https://github.com/benjamine/jsondiffpatch) for this.) Then the entry update code on the server can take the diff and apply it to what is loaded from Mongo (we'd have to parse [jsondiffpatch's delta format](https://github.com/benjamine/jsondiffpatch/blob/master/docs/deltas.md), but looking at its spec that doesn't look hard to parse in PHP). Then either we can write a Mongo update from that, or else we can apply those changes to a PHP `LexEntry` object and use the exisiting `$entry->write()` mechanism. I plan on doing the latter since that's the simplest code change that will work.

rmunn:
## Description

When a user edits an entry, the changes are now sent as deltas instead of full entries. This means that instead of replacing the entry entry when the user edits the English definition, that edit will turn into a Mongo update containing just "set definition.en.value = 'new text'". This should greatly reduce the number of times that user edits conflict with each other. When two people work on the same entry at the same time, the only way that their edits can conflict with each other is if they edit the same writing system in the same field.

However, for technical reasons, adding or removing a new Meaning or Example causes a full-entry update rather than a delta update. So if user A adds a new example sentence, this can cause the work that user B had done on the definition to be lost. This should hopefully be a rare scenario, however; we expect that more than 90% of edit conflicts will be solved by this PR.

Fixes #1051 

### Type of Change

## Screenshots

## How Can This Be Tested?

To test this, two or three people should edit the same entry in the same project. Sit where you can see each others' screens if possible, and watch the edits that the other person is making. Verify the following scenarios:

- [ ] One person edits the lexeme, the other person edits a definition. Both edits go through.
- [ ] One person edits the definition in English, the other person edits the same definition in Spanish. Both edits go through.
- [ ] One person edits the definition, the other edits an example sentence. Both edits go through.
- [ ] One person edits the definition, the other adds a **new** example sentence. This will cause a race condition where sometimes, both edits go through, while at other times, the definition is lost because the example sentence update "squashed" the definition update. (This will usually trigger if the definition update happens first and the new example sentence is saved second, but due to the different speed of the save and refresh timers, it's not always possible to guarantee triggering the race condition). This is a known issue, and would require an unfeasible amount of work to fix.

## Checklist:


longrunningprocess (commented):
Moving the symfony patch and `Exception` -> `Error` work to a different PR (as well as the `xdebug` config) might help keep the diff's in this PR a little more targeted to the goal of delta updates.  It's a little nit-picky perhaps but we discussed trying to make PR's a little more cohesive at our last retrospect.  This is not a blocker for me personally, just a suggestion to meet some of our process improvement goals.


The code reads well to me but I'd prefer to get @megahirt 's eyes on this one as well, therefore I won't approve.

rmunn:
> Moving the symfony patch and `Exception` -> `Error` work to a different PR (as well as the `xdebug` config) might help keep the diff's in this PR a little more targeted to the goal of delta updates.

Whoops. That wasn't supposed to go in here, though I do plan to make a PR for it as the Exception/Error thing trips me up every time I have a WIP that throws an exception, so it will be useful. I'll rebase and re-push this. Good catch.

rmunn:
Split the errant code into two PRs, #1206 and #1208.

megahirt:
@rmunn I apologize I wasn't able to review this today.  I hope to tomorrow.  Thanks for your patience.

megahirt:
@rmunn I have looked through a decent chunk of this PR and in general I see high quality work and changes.  Per our phone conversation, I had in my mind that delta updates would be applied via the existing JsonDecoder function, as I expected that to handle partial updates without code changes.  As it turns out, it handles deltas for simple properties, but for Array objects, it doesn't (yet).  I haven't tested mapOf types.  So I have started a draft PR in #1214  to create tests and also changed DecodeArrrayOf to fix it up for array updates.  My tests that I created are passing, however other unit tests are failing.  I need a few more hours to work on this but I don't have the time this weekend or before I take some vacation time.  So I think this will have to wait until either you or I pick this up.

src/Api/Model/Shared/DeepDiff/EditedDiff.php
```diff
@@ -0,0 +1,44 @@
...
+        $path = $this->toMongoPath();
+        return [ '$set' => [ $path => $this->newData ] ];
+    }
+
+    public function getValue() {
+        return $this->newData;
+    }
+
+    public function toActivityLogDifference() {
+        // TODO: Implement. Probably look for '.ws.value' at end of path, or rather ['ws', 'value'], and strip off the "value" part.
```
megahirt, line 33: what is this `TODO: implement` doing here?

src/angular-app/languageforge/lexicon/editor/editor.component.ts
```diff
@@ -345,10 +350,23 @@ export class LexiconEditorController implements angular.IController {
         isNewEntry = true;
         newEntryTempId = entryToSave.id;
         entryToSave.id = ''; // send empty id to indicate "create new"
+      } else {
```
megahirt, line 353:
```suggestion
```
Delete empty `else` statement

src/Api/Model/Shared/DeepDiff/EditedDiff.php
rmunn: Artifact of a code path I thought I was going to need, but that it turned out I could do without. I'll delete it.

src/Api/Model/Languageforge/Lexicon/Command/LexEntryCommands.php
```diff
@@ -138,7 +146,12 @@ public static function updateEntry($projectId, $params, $userId, $mergeQueuePath
             $differences = null; // TODO: Do we want differences even on a brand-new, added, entry?
         }
 
-        $entry->write();
+        if ($isDeltaUpdate) {
+            $update = DeepDiffDecoder::toMongoUpdate($deepDiff);
+            $entry->writeDiff($update, $userId, $now);
```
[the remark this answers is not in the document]
rmunn: Done, and the tests still run correctly! :confetti_ball:

PR #1206, opened 2021-10-05
Fix PHP exception handling in Symfony

### LINKED ISSUE #1205: PHP exceptions are swallowed by outdated exception handler
**Describe the bug**
The Symfony library that Language Forge uses for session handling has an exception handler that is based on obsolete versions of PHP. Specifically, it expects exceptions to be based on the Exception base class, but in PHP 7 that was changed to be based on the Error class instead. Both Exception and Error inherit from Throwable, but Error does not inherit from Exception. So any time the server encouters an error, what appears in the log message is not the original exception, but a new exception from Symfony about expecting an Exception instance but receiving an Error instance instead. This effectively "swallows" the original exception and makes bug-hunting in the server code a very tedious job.

I'm about to create a PR to fix this issue.

rmunn:
## Description

Fixes #1205 by patching the Symfony code to handle PHP 7 exceptions (Error instances instead of Exception instances), so that the original exception will be handled instead of a new "Expected Exception, got Error instead" exception being thrown, which was hiding the original exception.

This PR works by patching the Symfony code in our `vendor` folder after Composer has installed it. That's because upgrading our Symfony version to one that is actually based on PHP 7 would involve a LOT of updates, and this is a much lower-risk change. If we do end up upgrading Symfony at some point, this PR can be easily backed out by simply removing the patch from the Dockerfile.

### Type of Change

## How Has This Been Tested?

Once I installed this patch in my Dockerfile, I could actually see the exceptions being thrown by my syntax errors as I worked in the guts of the server's PHP code.

## Checklist:

### DIFF
--- a/docker/app/Dockerfile
+++ b/docker/app/Dockerfile
@@ -125,6 +125,10 @@ COPY docker/app/entrypoint.sh /
 
 COPY --from=composer-builder /composer/vendor /var/www/html/vendor
 
+# patch exception handling from Symfony to actually show exceptions instead of swallowing them
+COPY docker/app/symfony-exceptions.patch /
+RUN patch -p4 -i /symfony-exceptions.patch
+
 RUN echo "${BUILD_VERSION}" > /var/www/html/build-version.txt \
  && sed -i /var/www/html/version.php -e "s/^\\(define('VERSION', '\\).*;\$/\\1${BUILD_VERSION}'\\);/"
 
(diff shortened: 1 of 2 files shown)


longrunningprocess (approved):
very slick way to address this, thank you!  I'm assuming there's nothing different with the way we configure this or pull it in based on whether we're building a "dev" vs. "prod" image.

PR #1212, opened 2021-10-07
Establish v1.9.2-rc

longrunningprocess:
## Description
This will form the v1.9.2rc

### Type of Change

- Release candidate


rmunn (approved):
This looks like it has a lot of commits, but it's just a fast-forward merge: if you do it in Git, Git will default to a fast-forward merge and you'd have to use `--no-ff` if you want a "real" merge commit. GitHub doesn't allow fast-forward merges of PRs, so there are two choices:

1. Merge this PR with a merge commit, so that we don't see this situation again where there is a large number of unexpected commits. Downside: a merge commit in the repo history where one side of the merge is empty.
2. Merge this as a fast-forward at the Git command line (`git merge --ff-only`) and push that, then close this PR unmerged.

I suggest we discuss this at the next standup and decide how we want to handle this situation.

PR #1215, opened 2021-10-09
Feature/entry number for display and navigation

palmtreefrb:
## Description
This PR will allow users to enter an index number to move focus to an entry. With the below features.
 Should display the current entry's index number (1-based)
 Should be editable
 Should protect against invalid input, including numbers out of range (revert to current entry number)
 Upon user typing valid entry number, should navigate to that entry
 For desktop and mobile.

### Type of Change

- UI change

## Screenshots

Peek 2021-10-09 08-12

## How Has This Been Tested?
- [ ] Enter an entry index
- [ ] Change filter
- [ ] Change focus in list

## Checklist:

### DIFF
--- a/src/angular-app/languageforge/lexicon/editor/editor.component.ts
+++ b/src/angular-app/languageforge/lexicon/editor/editor.component.ts
@@ -316,6 +316,12 @@ export class LexiconEditorController implements angular.IController {
     return (mod.filterBy && mod.filterBy.option) || mod.sortBy.value !== 'default' || mod.sortReverse || mod.wholeWord || mod.matchDiacritic
   }
 
+  filterSortIsOn() {
+    const mod = this.entryListModifiers;
+    let on = mod.filterBy != null && mod.filterBy.option !=null || mod.sortBy.value !== 'default' || mod.sortReverse;
+    return on;
+  }
+
   shouldShowFilterReset() {
     const modifiers = this.entryListModifiers;
     return modifiers.filterActive() || modifiers.sortBy.value !== 'default' || modifiers.sortReverse
--- a/src/angular-app/languageforge/lexicon/editor/editor.component.ts
+++ b/src/angular-app/languageforge/lexicon/editor/editor.component.ts
@@ -469,6 +469,13 @@ export class LexiconEditorController implements angular.IController {
     this.goToEntry(id);
   }
 
+  gotoToEntry(index: number, isValid: boolean) {
+    if (isValid) {
+      let id = this.editorService.getIdInFilteredList(Number(index));
+      this.editEntryAndScroll(id);
+    }
+  }
+
   canSkipToEntry(distance: number): boolean {
     const i = this.editorService.getIndexInList(this.currentEntry.id, this.visibleEntries) + distance;
     return i >= 0 && i < this.visibleEntries.length;
--- a/src/angular-app/languageforge/lexicon/editor/editor.component.ts
+++ b/src/angular-app/languageforge/lexicon/editor/editor.component.ts
@@ -476,6 +476,12 @@ export class LexiconEditorController implements angular.IController {
     }
   }
 
+  entryIndex(): number {
+    let id = this.currentEntry.id;
+    let index = this.editorService.getIndexInList(id, this.entries)
+    return index + 1;
+  }
+
   canSkipToEntry(distance: number): boolean {
     const i = this.editorService.getIndexInList(this.currentEntry.id, this.visibleEntries) + distance;
     return i >= 0 && i < this.visibleEntries.length;
--- a/src/angular-app/languageforge/lexicon/editor/editor.component.ts
+++ b/src/angular-app/languageforge/lexicon/editor/editor.component.ts
@@ -473,6 +473,8 @@ export class LexiconEditorController implements angular.IController {
     if (isValid) {
       let id = this.editorService.getIdInFilteredList(Number(index));
       this.editEntryAndScroll(id);
+      let gotoElement = (<HTMLInputElement>document.getElementById('goto'));
+      gotoElement.value = '';
     }
   }
 
--- a/src/angular-app/languageforge/lexicon/editor/editor.component.ts
+++ b/src/angular-app/languageforge/lexicon/editor/editor.component.ts
@@ -321,12 +321,6 @@ export class LexiconEditorController implements angular.IController {
     return (mod.filterBy && mod.filterBy.option) || mod.sortBy.value !== 'default' || mod.sortReverse || mod.wholeWord || mod.matchDiacritic
   }
 
-  filterIsOn() {
-    const mod = this.entryListModifiers;
-    let on = mod.filterBy != null && mod.filterBy.option !=null;
-    return on;
-  }
-
   shouldShowFilterReset() {
     const modifiers = this.entryListModifiers;
     return modifiers.filterActive() || modifiers.sortBy.value !== 'default' || modifiers.sortReverse
--- a/src/angular-app/languageforge/lexicon/editor/editor.component.ts
+++ b/src/angular-app/languageforge/lexicon/editor/editor.component.ts
@@ -49,6 +49,11 @@ export class LexiconEditorController implements angular.IController {
 
   lastSavedDate = new Date();
   currentEntry: LexEntry = new LexEntry();
+
+  currentIndex = {
+    index: -1
+  }
+
   commentContext = {
     contextGuid: ''
   };
@@ -316,9 +321,9 @@ export class LexiconEditorController implements angular.IController {
     return (mod.filterBy && mod.filterBy.option) || mod.sortBy.value !== 'default' || mod.sortReverse || mod.wholeWord || mod.matchDiacritic
   }
 
-  filterSortIsOn() {
+  filterIsOn() {
     const mod = this.entryListModifiers;
-    let on = mod.filterBy != null && mod.filterBy.option !=null || mod.sortBy.value !== 'default' || mod.sortReverse;
+    let on = mod.filterBy != null && mod.filterBy.option !=null;
     return on;
   }
 
@@ -479,15 +484,14 @@ export class LexiconEditorController implements angular.IController {
     if (isValid) {
       let id = this.editorService.getIdInFilteredList(Number(index));
       this.editEntryAndScroll(id);
-      let gotoElement = (<HTMLInputElement>document.getElementById('goto'));
-      gotoElement.value = '';
     }
   }
 
   entryIndex(): number {
     let id = this.currentEntry.id;
     let index = this.editorService.getIndexInList(id, this.entries)
-    return index + 1;
+    this.currentIndex.index = index + 1;
+    return this.currentIndex.index;
   }
 
   canSkipToEntry(distance: number): boolean {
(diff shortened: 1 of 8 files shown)


src/angular-app/languageforge/lexicon/editor/editor.component.ts
```diff
@@ -441,6 +441,21 @@ export class LexiconEditorController implements angular.IController {
...
   }
 
+  gotoToEntry(index: number, isValid: boolean) {
+    if (isValid) {
+      let id = this.editorService.getIdInFilteredList(Number(index));
+      this.editEntryAndScroll(id);
+      let gotoElement = (<HTMLInputElement>document.getElementById('goto'));
+      gotoElement.value = '';
+    }
+  }
```
megahirt, line 451: It feels like the function name should be `goToEntry() instead of `gotoToEntry`  Can you rename?

The if statement around validation isn't very nice, since I believe AngularJS has built-in validation as part of a form using ngSubmit.  e.g. you could wrap the input in a <form ngSubmit="goToEntry()"> and I think the min / max validation will happen for free.  No need to pass in the form validation state.

That's what I'm thinking in my head, however I haven't tried it out.  Let me know how you get on.

src/angular-app/languageforge/lexicon/editor/field/dc-entry.component.html
```diff
@@ -13,7 +13,7 @@
             <div class="comment-bubble-entry">
                 <comment-bubble control="$ctrl.control" field="$ctrl.fieldName" model="$ctrl.model" parent-context-guid="$ctrl.contextGuid"></comment-bubble>
             </div>
-            &nbsp;Entry
+            &nbsp;Entry, Number: {{$ctrl.entryIndex}}
```
megahirt, line 16: Is this addition intentional?

megahirt (changes requested):
Nice work Rick!  This will be a nice improvement, especially for Margarete who is having difficulty navigating her dictionary.  I would like to know if the <form ngSubmit> idea is feasible.  If it's not then let me know and I will approve as is.  Thanks for checking that out.  Let me know if you have questions.

src/angular-app/languageforge/lexicon/editor/editor.component.ts
palmtreefrb: There already is a function with that name private goToEntry(entryId: string)

palmtreefrb: For the ngSubmit suggestion to work would require an input type of "submit" the text box is input type of "numeric"

src/angular-app/languageforge/lexicon/editor/field/dc-entry.component.html
palmtreefrb: This was discussed in Slack, Alex liked this suggestion to add the entry number to the card.
https://sil-language-software.slack.com/archives/CGDKZ7NTV/p1633621109496300

rmunn:
I was testing this feature today, to approve it, and found a bug. Point 2 of https://github.com/sillsdev/web-languageforge/pull/1215#issuecomment-944314582 says: 

> The entry number is editable when there are no search filters in place. If a filter is active, the entry number input and the back/forward buttons are greyed out. Rationale: back/forward and entry number navigation are useful for browsing the dictionary as a whole. They are not useful in navigating a results subset.

If the entry number input is edited and you press Enter, while there is no filter in place, this works correctly. Typing in 5 and pressing Enter jumped me to entry number 5, as it should.

However, **if there is a filter in place**, then typing in the entry number input and pressing Enter does not behave correctly. I set a filter that showed only two items, then I typed 5 into the input and pressed Enter.

**Expected result**: I get jumped to entry number 5 (the fifth entry in the unfiltered list).
**Actual result**: I get taken out of the editor view and into the index view, with the filter in place so I only see two items.

This is good work, and I'm looking forward to getting it merged in once all the bugs are fixed.

rmunn:
Looked at it again, and I missed the part of the design where it says that the **entry number input** should also be grayed out along with the Previous/Next buttons. So forget what I said about expected behavior when typing into the entry number input with a filter in place. It's actually supposed to be disabled in that circumstance. So the bug report is just "entry number input wasn't disabled when Previous/Next buttons were disabled". This was in Firefox on Linux, in case that matters.

megahirt:
@palmtreefrb I've been out of action for several weeks and I think once you address @rmunn 's bug, and he tests it, then we can merge this.  Looks like a rebase is in order too.

palmtreefrb:
> @palmtreefrb I've been out of action for several weeks and I think once you address @rmunn 's bug, and he tests it, then we can merge this. Looks like a rebase is in order too.

The bug is fixed and I have rebased this branch with develop.

rmunn:
I pulled the latest version of the branch and built it, and I can still type into the entry number input when there's an active filter. It's not actually disabled yet.

Also, I've found a bug: when you type a search expression into the filter, the entry number can change, without any apparent cause. For example, here's what happens when there is no filter and I have the sixth item selected:

image

Now I type `ab` into the filter, and the following happens:

image

It seems very unintuitive to have the entry number change from 6 to 8 when I added a filter expression. (After I typed the letter `a` the entry number was still 6. It only became 8 when I typed the `b`). With different filter expressions, sometimes the entry number stayed 6, once it became 7, and once it became 15. So the results can vary, but there's definitely a bug somewhere.

Actually, the bug might not even be in your code, @palmtreefrb. The logic around the filtered list view is complex enough that we could actually have some other bugs in there, that your code is merely exposing but not causing. But before we can merge this PR, we certainly do need to find and fix those bugs, because if the user sees the entry number jump around it will be very confusing.

rmunn:
Found another behavior I'm not entirely happy with. When you type in the filter list view, the entry number input becomes red once it's out of range. E.g. you're viewing entry number 6, but you've filtered the list view so it only shows 2 items. The entry input box shows 6 and has a red border.

I don't believe we should show a red border around the entry input field when it's disabled. The red border (validation failed) should only be displayed when the input is enabled *and* incorrect. So if the field is being correctly disabled when a filter is applied, then there's no need to change the validation logic. You'll just need to have the red border show up when the field is enabled AND the validation has failed.

I think I would add a couple more requirements to the list that Chris made above:

- [ ] An invalid or out-of-range input box shows a red border if **and only if** the field is editable. (So if there's a filter and the field has become uneditable, the red border will never display).
- [ ] The entry number **only** changes when one of the following occurs:
  - The user clicks the Previous or Next buttons.
  - The user types in the entry number box and presses Enter.
  - The user clicks on a different item in the list. (**See below for more about this one**)

In particular, the entry number should *not* change when filtering items. It will end up showing a number that's outside the range, but if I was looking at entry number 6 in the unfiltered list, filtering the list should not tell me I'm looking at entry number 8.

Now for the "clicks on a different item in the list" part. In my testing, I filtered the list by typing `ab` and *then* clicked on one of the list items (the first one in the *filtered* list). When I clicked on it, the entry list number showed "1". Then I removed the fitler (by clicking the X button next to the filter search text) and the entry number jumped to 10. I know why that's happening: when I clicked on the entry in the filtered list, it was the 1st item (index 0) in the filtered list but the 10th (index 9) of the whole dictionary. And so the existing code that's supposed to look up an item's position in the list is using the *filtered* list to look it up. But what we *should* be using is the *unfiltered* list here.

In other words, there's another new requirement coming out of my testing:

- [ ] The entry index number always shows the index of that entry in the *complete*, unfiltered dictionary.

This might be a new requirement that you weren't aware of, in which case I apologize for the extra work that it will take to meet this requirement. But I'm pretty sure this is always how we intended the feature to work.

rmunn:
Found one reason for the entry input not being disabled. You've created a `filterIsOn` function to detect whether the filter is active, but its logic is only checking the option list and not the filter text. So when I type something into the filter text input, `filterIsOn()` returns false when it should return true.

Fortunately, the fix is simple. There's already a `filterActive()` method on the `this.entryListModifiers` object. So all you need to do is call `entryListModifiers.filterActive()` instead of `filterIsOn()` and you'll have the right logic already, instead of duplicating that logic somewhere else.

In other words, this should fix that part of the issue.

```diff
diff --git a/src/angular-app/languageforge/lexicon/editor/editor.component.ts b/src/angular-app/languageforge/lexicon/editor/editor.component.ts
index 597dc7b65..d8b2dcacc 100644
--- a/src/angular-app/languageforge/lexicon/editor/editor.component.ts
+++ b/src/angular-app/languageforge/lexicon/editor/editor.component.ts
@@ -317,8 +317,7 @@ export class LexiconEditorController implements angular.IController {
 
   filterIsOn() {
     const mod = this.entryListModifiers;
-    let on = mod.filterBy != null && mod.filterBy.option !=null;
-    return on;
+    return mod.filterActive();
   }
 
   shouldShowFilterReset() {
```

I've tested this and it does disable the entry number input correctly. The input still shows red when it's disabled, and it shouldn't, but that's a different issue I don't have time to track down right now.

rmunn:
> Robin, this was by design.The requirement was to disable the index search when a filter was applied. Shouldn't we allow a search if only sorting the list?

The `filterActive()` method is defined like this:

```js
  filterActive = () => !!(this.filterText() || this.filterBy && this.filterBy.option);
```

That won't consider the sort at all, but it *will* consider the `filterText`, which the current code is not considering.

Sorting is an issue I hadn't considered. Do we want the entry number to change when the user reverses the sort order, or not? The way the code is right now, if I click on entry 1 of my 30-entry project and reverse the sort order, the entry number field becomes 30. If I click on entry 2 and reverse the sort order, it becomes 29. My gut feeling is this is *not* what users are expecting, and that we should make the entry number be "the entry's order in the *normally-sorted* order of the dictionary", not the current sort selection. But I could see an argument either way for it, and that's something we can discuss at our next standup meeting.

At the moment, though, the `filterIsOn` function should be using `filterActive` so that it takes `filterText` into consideration as well. And the entry number needs to *not* be based on the filtered list, but on the whole, unfiltered list. We do not want the behavior where I click on the first item of the filtered list and see "entry number 1", then unfliter the list and the entry number jumps to 10. It should have shown "10" even though that entry was the first item in the filtered list, because that's the entry number in the whole dictionary.

rmunn:
After further review of the LF code, it seems that some of the issues with the entry number jumping around are difficult to solve and aren't something that this PR introduces. For example, sorting entries actually changes the `this.entries` array and not just `this.visibleEntries`. (Look at the `sortEntries` function in editor-data.service.ts). So it's going to be difficult to implement this feature as originally envisioned, because we're going to have to make some fundamental changes to how we handle the entry filter and sort feature if we want the entry number to look stable.

palmtreefrb:
> After further review of the LF code, it seems that some of the issues with the entry number jumping around are difficult to solve and aren't something that this PR introduces. For example, sorting entries actually changes the `this.entries` array and not just `this.visibleEntries`. (Look at the `sortEntries` function in editor-data.service.ts). So it's going to be difficult to implement this feature as originally envisioned, because we're going to have to make some fundamental changes to how we handle the entry filter and sort feature if we want the entry number to look stable.

I had come to this very same conclusion after diving deeper into the workings of the filter and sort implementation code and wanted to discus on the next stand-up.

megahirt (approved):

megahirt:
[This latest commit](https://github.com/sillsdev/web-languageforge/pull/1215/commits/a646fe8e4d75b9e8be19653b0a22918c9303c8c0) fixes the misbehaving red box behavior on blur and hides the control completely when the filter is active, thereby avoiding the wonky behavior discussed earlier in this PR.

Here is a demo of the control working after this latest commit.
Animation3

@palmtreefrb and @longrunningprocess wanted to get your review on this before merging.

megahirt:
@josephmyers and I paired on this PR today and also had a bit of a design discussion on moving the entry index navigation closer to the entry list UI, which is captured in #1274

PR #1224, opened 2021-10-15
Adjustable update timer

### LINKED ISSUE #1054: Allow poll update timer interval to be adjustable in project settings
Along with the race condition mentioned in https://github.com/sillsdev/web-languageforge/issues/1051, the poll update timer interval (currently hard-coded to 32 seconds) is something that we might want to adjust in the project configuration somewhere. Maybe in an "Advanced" tab with a "Don't touch this unless you know what you're doing!" warning (might even be a checkbox that says "I promise I know what I'm doing and I won't complain if this makes something break" *grin*). The default would be 32 seconds, same as now, but could be set shorter.

Would require that the LexiconSendReceiveService method `startPollUpdateTimer` take a parameter: the time to set the timer to, in milliseconds. If the parameter is omitted then it will default to the current hardcoded value. Then when the setting is chagned in the project, the project will call `cancelPollUpdateTimer()` followed by `startPollUpdateTimer(timerSeconds * 1000)`.

rmunn:
## Description

This PR creates an "Advanced Configuration" tab in the project configuration, which currently contains only one setting: the ability to change the poll update timer from its default of 32 seconds. This will cause everyone logged in to the project to make much more frequent requests to the server, so we don't want to enable it for everyone. But the language team whose needs caused us to create #1054 will benefit from setting this to 5 seconds instead of 32.

Fixes #1054

### Type of Change

- UI change

## Screenshots

image

## How Has This Been Tested?

- [X] Updated the setting in a test project on my local Docker install of LF, and verified that it sent poll requests much more frequently after I updated the setting.
- [X] Loaded that project with a non-manager user, and verified that the project settings were being loaded and applied correctly (the non-manager user was also polling more often)

## Checklist:

### DIFF
--- a/test/app/languageforge/lexicon-traversal.e2e-spec.ts
+++ b/test/app/languageforge/lexicon-traversal.e2e-spec.ts
@@ -69,16 +69,6 @@ describe('Lexicon E2E Page Traversal', () => {
       await configurationPage.unifiedPane.example.fieldSpecificInputSystemCheckbox('Translation', 0).click();
       await configurationPage.unifiedPane.fieldSpecificButton('Translation').click();
     });
-
-    it('Input Systems tab', async () => {
-      await configurationPage.tabs.inputSystems.click();
-      await configurationPage.inputSystemsPane.moreButton.click();
-    });
-
-    it('Option List tab', async () => {
-      await configurationPage.tabs.optionlists.click();
-      // There is no model of option list tab - Mark W 2018-01-14
-    });
   });
 
   describe('Explore editor page', () => {


rmunn:
E2E tests are running incredibly slowly, and the same tests on the `develop` branch are running normally. I think that's an indication of a problem, so I'm not going to merge until I've figured it out. I can think of one possible reason for the issue, which is that somehow the timer interval is getting loaded as 0 in some circumstances and therefore triggering as fast as possible, which would slow the server down to a crawl if this got merged. I'm therefore setting this PR to draft mode until I've identified the issue, to make sure it won't get merged by mistake.

rmunn:
Looks like that fixed the slowdown on the E2E tests. I'll take this out of draft mode and, when the E2E tests complete, merge it.

PR #1243, opened 2021-11-09
Ignore diacritics by default when searching

### LINKED ISSUE #1114: feat: ignore diacritics as new default search behavior; add "match diacritics" search option for "exact match"
Consider enabling approximate string matching when doing word searches, as in DAB.

https://user-images.githubusercontent.com/82162128/131817635-293f5f03-093b-4839-9e74-bc8b92a8ef9b.mp4

https://user-images.githubusercontent.com/82162128/131817667-d000bcd4-bfd2-443b-af3f-881f197eeaf7.mp4

This would be a very helpful feature, as the primary LF user is most likely still learning how to write in his language.

(Note: I like how DAB gives the option of matching diacritics, and that it is disabled be default. This is helpful because sometimes there can be two or more words that are only differencible by their diacritics. (Example in Panao Quechua: ćhaqui vs chaqui.) Adding this search option to LF I think is a separate decision from adding approximate string matching. We need to assess if the complication it adds to the GUI is worth it. My hunch of to not add the option.)

Thanks!

longrunningprocess:
## Description

This feature will change the default search to include "diacritic-agnostic" results.  It will also add a new advanced option allowing users to narrow results based on diacritics.

Fixes #1114 

### Type of Change

- UI change

## Screenshots

Provided with test cases below

## How Has This Been Tested?

Project used for testing (5491 entries):  [Alex's project.zip](https://github.com/sillsdev/web-languageforge/files/7412915/Alex.s.project.zip)

### Functional testing (should work the same in both list and entry view)

- [x] Searching for `chaca` or `CHACA` should result in 124 matches with no advanced options selected

image

- [x] Searching for `chaca` or `CHACA` should result in 90 matches when "matching diacritics"

image

- [x] Searching for `chacä` or `CHACÄ` should result in 124 matches with no advanced options selected

image

- [x] Searching for `chacä` or `CHACÄ` should result in 9 matches when "matching diacritics"

image

- [x] Searching for `chacå` or `CHACÅ` should result in 124 matches with no advanced options selected

image

- [x] Searching for `chacå` or `CHACÅ` should result in 0 matches when "matching diacritics"

image

- [x] Checking the "Match diacritics" option should cause the "Options" changed indicator to appear

image

- [x] see #1229 for more regression tests

## Checklist:

### DIFF
--- a/src/angular-app/bellows/core/offline/editor-data.service.ts
+++ b/src/angular-app/bellows/core/offline/editor-data.service.ts
@@ -475,7 +475,7 @@ export class EditorDataService {
         if (isSemanticDomain && this.semanticDomainsMatch(val, rawQuery)) {
           found = true;
         } else {
-          const normalizedValue = this.entryListModifiers.matchDiacritic ? val : this.normalizeDiacritics(val)
+          const normalizedValue = this.entryListModifiers.matchDiacritic ? val : this.removeDiacritics(val)
 
           if (queryRegex.test(normalizedValue)) {
             found = true;
--- a/src/angular-app/bellows/core/offline/editor-data.service.ts
+++ b/src/angular-app/bellows/core/offline/editor-data.service.ts
@@ -451,18 +451,38 @@ export class EditorDataService {
     return input.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'); // $& means the whole matche
   }
 
+  private normalizeDiacritics(input: string) {
+    // refs:
+    // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize
+    // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Unicode_Property_Escapes
+    // https://unicode.org/reports/tr44/#Diacritic
+
+    // https://stackoverflow.com/a/37511463/10818013
+
+
+    return input.normalize('NFD').replace(/\p{Diacritic}/gu, '')
+  }
+
   private entryMeetsFilterCriteria(config: any, entry: LexEntry): boolean {
     if (this.entryListModifiers.filterText() !== '') {
-      const query = this.escapeRegex(this.entryListModifiers.filterText());
-      const queryRegex = new RegExp(this.entryListModifiers.wholeWord ? `\\b${query}\\b` : query, 'i');
+      const rawQuery = this.entryListModifiers.filterText()
+      const normalizedQuery = this.entryListModifiers.matchDiacritic ? rawQuery : this.normalizeDiacritics(rawQuery);
+      const regexSafeQuery = this.escapeRegex(normalizedQuery);
+      const queryRegex = new RegExp(this.entryListModifiers.wholeWord ? `\\b${regexSafeQuery}\\b` : regexSafeQuery, 'i');
       let found = false;
-      
+
       this.walkEntry(config.entry, entry, (val, isSemanticDomain) => {
-        if (queryRegex.test(val) || (isSemanticDomain && this.semanticDomainsMatch(val, query))) {
-          found = true
+        if (isSemanticDomain && this.semanticDomainsMatch(val, rawQuery)) {
+          found = true;
+        } else {
+          const normalizedValue = this.entryListModifiers.matchDiacritic ? val : this.normalizeDiacritics(val)
+
+          if (queryRegex.test(normalizedValue)) {
+            found = true;
+          }
         }
       });
-      
+
       if (!found) return false;
     }
 
--- a/src/angular-app/bellows/core/offline/editor-data.service.ts
+++ b/src/angular-app/bellows/core/offline/editor-data.service.ts
@@ -45,6 +45,7 @@ class EntryListModifiers {
   sortOptions: SortOption[] = [];
   sortReverse = false;
   wholeWord = false;
+  matchDiacritic = false;
   filterBy: {
     text: string;
     option: FilterOption;
--- a/src/angular-app/bellows/core/offline/editor-data.service.ts
+++ b/src/angular-app/bellows/core/offline/editor-data.service.ts
@@ -456,9 +456,7 @@ export class EditorDataService {
     // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize
     // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Unicode_Property_Escapes
     // https://unicode.org/reports/tr44/#Diacritic
-
     // https://stackoverflow.com/a/37511463/10818013
-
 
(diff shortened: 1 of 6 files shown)


longrunningprocess:
I still need to double-check the browser sorry for the character class I'm using and I also want to check this other ” combination" diacritic class that seems more specific.

src/angular-app/bellows/core/offline/editor-data.service.ts
```diff
@@ -450,18 +451,38 @@ export class EditorDataService {
     return input.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'); // $& means the whole matche
   }
 
+  private normalizeDiacritics(input: string) {
```
megahirt, line 454:
```suggestion
  private removeDiacritics(input: string) {
```
Suggest a function rename since the returned string actually has no diacritics, not just normalized to NFD.

src/angular-app/bellows/core/offline/editor-data.service.ts
```diff
@@ -450,18 +451,38 @@ export class EditorDataService {
...
+
+    return input.normalize('NFD').replace(/\p{Diacritic}/gu, '')
+  }
+
   private entryMeetsFilterCriteria(config: any, entry: LexEntry): boolean {
     if (this.entryListModifiers.filterText() !== '') {
-      const query = this.escapeRegex(this.entryListModifiers.filterText());
-      const queryRegex = new RegExp(this.entryListModifiers.wholeWord ? `\\b${query}\\b` : query, 'i');
+      const rawQuery = this.entryListModifiers.filterText()
+      const normalizedQuery = this.entryListModifiers.matchDiacritic ? rawQuery : this.normalizeDiacritics(rawQuery);
```
megahirt, line 469:
```suggestion
      const normalizedQuery = this.entryListModifiers.matchDiacritic ? rawQuery : this.removeDiacritics(rawQuery);
```

src/angular-app/bellows/core/offline/editor-data.service.ts
```diff
@@ -450,18 +451,38 @@ export class EditorDataService {
...
       let found = false;
-      
+
       this.walkEntry(config.entry, entry, (val, isSemanticDomain) => {
-        if (queryRegex.test(val) || (isSemanticDomain && this.semanticDomainsMatch(val, query))) {
-          found = true
+        if (isSemanticDomain && this.semanticDomainsMatch(val, rawQuery)) {
+          found = true;
+        } else {
+          const normalizedValue = this.entryListModifiers.matchDiacritic ? val : this.normalizeDiacritics(val)
```
megahirt, line 478:
```suggestion
          const normalizedValue = this.entryListModifiers.matchDiacritic ? val : this.removeDiacritics(val)
```

megahirt (approved):
Approved with a ⭐ !
I'm impressed with your independent work to take advantage of research resources (C Hubbard) and to design and implement the solution.  You also took care to make sure that the UX is consistent across products and consulted Alex as power user to make sure we are getting it right.

This is great work and will be immediately helpful to users when it is shipped.

Your PR description and test scenarios are first class - well done.

megahirt:
I would like to add an additional test case that demonstrates diacritics keyed in both NFC and NFD form.  The test should demonstrate that diacritics are ignored regardless of NFD/NFC input.  I snagged some images from the Unicode docs on normalization forms to show some sample characters to search for with their codepoints, where these diacritics are safely ignored in search.  How you key in these characters depends on your OS and keyboard software.

image

image


As I write this, I realize that "exact match" might not work in certain edge cases where the input query is not in the same normalized form (NFC/NFD) as the data being searched.  The fix is of course to normalize the query and data before comparison.  This would be a separate issue to look into, not part of this PR.

src/angular-app/bellows/core/offline/editor-data.service.ts
```diff
@@ -450,18 +451,38 @@ export class EditorDataService {
...
+  private normalizeDiacritics(input: string) {
+    // refs:
+    // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize
+    // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Unicode_Property_Escapes
+    // https://unicode.org/reports/tr44/#Diacritic
+
+    // https://stackoverflow.com/a/37511463/10818013
+
+
+    return input.normalize('NFD').replace(/\p{Diacritic}/gu, '')
```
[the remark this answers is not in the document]
longrunningprocess: great, I'm glad you're pleased with it, thanks for reviewing.  Just to confirm, you're good with the browser support and you think `\p{Diacritic}` is good, no need to further investigate `InCombiningDiacriticalMarks`?

src/angular-app/bellows/core/offline/editor-data.service.ts
longrunningprocess: yep, good call!

longrunningprocess:
> I would like to add an additional test case that demonstrates diacritics keyed in both NFC and NFD form. The test should demonstrate that diacritics are ignored regardless of NFD/NFC input. I snagged some images from the Unicode docs on normalization forms to show some sample characters to search for with their codepoints, where these diacritics are safely ignored in search. How you key in these characters depends on your OS and keyboard software.
> 
> image
> 
> image
> 
> As I write this, I realize that "exact match" might not work in certain edge cases where the input query is not in the same normalized form (NFC/NFD) as the data being searched. The fix is of course to normalize the query and data before comparison. This would be a separate issue to look into, not part of this PR.

ok, thanks for thinking of some additional use cases, I've felt like I was on shaking ground with my limited testing.  I'll see if I can figure out if I can doing something in the browser console to test this.

longrunningprocess:
FYI @megahirt a brief NFD/NFC test:

good (except for highlighting) in the default search mode:
image


no good when attempting to match diacritics:
image

longrunningprocess:
I'm going to go ahead and merge this under the assumption you did check the browser support and were ok as well as anticipating `InCombiningDiacriticalMarks` is probably irrelevant.

megahirt:
@longrunningprocess in my testing this, I realized that we didn't address the highlighting that matches a string.  I'm pretty sure you knew about this, but I'm just commenting about it explicitly for future reference.

The [highlightMatches() method](https://github.com/sillsdev/web-languageforge/blob/staging/src/angular-app/languageforge/lexicon/editor/editor.component.ts#L661) would need to be rewritten or modified to take into a account the part about ignoring diacritics.

When I search for "chu" in the default mode, only the exact matches for "chu" are selected, even though the results show those that include diacritics, see below.

image
(whole word is selected as an option to reduce hits)

longrunningprocess:
> @longrunningprocess in my testing this, I realized that we didn't address the highlighting that matches a string. I'm pretty sure you are already aware of it, and that's fine. I'm mostly just commenting about it explicitly for future reference.

I wasn't even looking at it, I'm glad you put fresh eyes on it.  I'll convert your finding to a new issue.

PR #1246, opened 2021-11-15
cleaned up some old livereload remnants

longrunningprocess:
## Description

I came across some old livereload deps that aren't used anymore.

### Type of Change

- cleaning up repo

## Screenshots

N/A

## How Has This Been Tested?

- [x] `make dev` still works locally

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,11 +1,11 @@
 {
-  "name": "x-forge",
+  "name": "web-languageforge",
   "version": "1.0.0",
   "lockfileVersion": 2,
   "requires": true,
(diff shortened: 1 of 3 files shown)


palmtreefrb (approved):
Yes, now that we have "make dev" live reload is not needed nor is useful in Docker environment?

longrunningprocess:
> Yes, now that we have "make dev" live reload is not needed nor is useful in Docker environment?

well, it's certainly possible to make it work I just don't think it's practical or worth it honestly.

PR #1247, opened 2021-11-16
update service-worker implementation

longrunningprocess:
## Description

We were using a very old version of workbox to generate our service worker as well as some outdated techniques.  This PR strives to modernize the implementation in the hopes it will resolve the problems noticed in #1233.

Fixes #1233 

### Type of Change

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Screenshots

N/A

## How Has This Been Tested?

- [x] I ensured the service worker gets registered
- [x] I reviewed some of the assets built into the precache by workbox
- [x] I made sure I was presented with an option to install as a PWA (in staging) on my desktop
- [x] I made sure I was presented with an option to install as a PWA (in staging) on my phone
- [x] I made an update to the app's css code (see https://github.com/sillsdev/web-languageforge/pull/1247/commits/7a59b63e77dd527169619deea6edae304c69cf23 for CSS variations) and made sure the PWA reflected that new styling

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have reviewed the title/description of this PR which will be used as the squashed PR commit message
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #1249, opened 2021-11-19
Optimize app Dockerfile with npm cache

megahirt:
## Description

This PR is intended to be merged after Billy's service-worker PR.  It probably won't pass tests until it is merged.

This PR optimizes the build time of `docker-compose build app` in the following ways:

1. It utilizes an npm cache image to reduce/eliminate having to download npm packages from the internet.  As we update our dependencies, this npm cache image will slowly provide less performance value.  For now, it provides about a 50% speed increase in `npm install` time.
2. It remove unnecessary COPY lines that aren't used by webpack
3. It targets the COPY src/Site/views to only copy sass files

## Performance Metrics on Chris's machine from Thailand
`docker-compose build app`

### Before This PR

package.json file change 2:47
sass file, html file, typescript file change 0:47
PHP file change 0:03
TWIG file change 0:47

### After This PR
package.json file change **1:53**
sass file, html file, typescript file change 0:47
PHP file change 0:03
TWIG file change **0:09**

## Checklist:

TODO:
- publish the npm-cache image to sillsdev so it's not published under megahirt.  At this time I don't have access to push and I need help getting push access.

### DIFF
--- a/docker/app/Dockerfile
+++ b/docker/app/Dockerfile
@@ -11,7 +11,7 @@ RUN mkdir -p /data
 WORKDIR /data
 
 # copy npm cache directory; as optimization for npm install
-COPY --from=megahirt/web-languageforge:npm-cache /root/.npm /root/.npm
+COPY --from=sillsdev/web-languageforge:npm-cache /root/.npm /root/.npm
 
 # unsafe-perm true is required to work around an npm bug since we are running as root, have a git+HTTPS repo source and it has a `prepare` script (perfect storm).
 # see https://github.com/npm/npm/issues/17346
--- a/docker/app/Dockerfile
+++ b/docker/app/Dockerfile
@@ -10,6 +10,9 @@ RUN npm config set unsafe-perm true && npm install -g npm@7.6.3
 RUN mkdir -p /data
 WORKDIR /data
 
+# copy npm cache directory; as optimization for npm install
+COPY --from=megahirt/web-languageforge:npm-cache /root/.npm /root/.npm
+
 # unsafe-perm true is required to work around an npm bug since we are running as root, have a git+HTTPS repo source and it has a `prepare` script (perfect storm).
 # see https://github.com/npm/npm/issues/17346
 COPY package.json package-lock.json ./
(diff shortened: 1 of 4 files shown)


docker/app/Dockerfile
```diff
@@ -10,6 +10,9 @@ RUN npm config set unsafe-perm true && npm install -g npm@7.6.3
 RUN mkdir -p /data
 WORKDIR /data
 
+# copy npm cache directory; as optimization for npm install
+COPY --from=megahirt/web-languageforge:npm-cache /root/.npm /root/.npm
```
megahirt, line 14: Once I know how to publish to sillsdev, we can change this away from the megahirt repo

docker/app/Dockerfile
```diff
@@ -20,13 +23,11 @@ COPY typings ./typings/
...
 
 # copy in src local files
+# Note: *.html files in src/angular-app aren't necessary for webpack compilation, however changes to HTML files will invalidate this layer
 COPY src/angular-app ./src/angular-app
-COPY src/appManifest ./src/appManifest
-COPY src/js ./src/js
-COPY src/json ./src/json
 COPY src/sass ./src/sass
-COPY src/service-worker ./src/service-worker
-COPY src/Site/views ./src/Site/views
```
megahirt, line 29: Removing the appManifest and service-worker lines will only work after #1247 is merged

longrunningprocess: I merged it, you may rebase now

docker/app/Dockerfile
longrunningprocess: I think this is promising, I'd like to suggest this be done via a GHA triggered off any push to `develop` and producing a `npm-cache-develop` tag.

Additionally, I'd like to consider/discuss the implications this may have when producing the "production" version of this image.

longrunningprocess (changes requested):
Every little bit helps, thanks 🚀 

I believe this needs a little more discussion, let me know if/when we can talk.

docker/app/Dockerfile
megahirt: I don't see any issues with using this npm-cache in a production build.  It's a cache that npm should be able to intelligently use at it's discretion, but should not affect the product output or final app image in any way (thanks to multi-stage builds!).  The cache folder (`~/.npm`) will be generated either way.  Saving it as a docker image instead of having to rebuild it on every run should save time in both prod and dev image builds.  That's my thinking anyway.

megahirt: I like the idea of of publishing a GHA triggered npm-cache-develop image after we have proved out the concept.

longrunningprocess: yep, I was not thinking about it properly, thanks for reminding me this is transient ad won't impact our app integrity across versions.

longrunningprocess:
@megahirt I had a moment to add the GHA and seed Docker Hub.  If you're good with these commits and there's nothing else you'd like to do I'm good with the merge.  Please feel free to squash-merge even though we've both worked in this one.

Additionally, we may want to think about our `make clean` targets because we have nothing in our `Makefile` at this point that would clean this new image...I might suggest a line in the `powerwash` target that will explicitly remove this image.

longrunningprocess (approved):
once this goes into `develop` it would be nice to ensure the button shows up for run the workflow on demand.

I made one suggestion related to cleaning things up when a dev requests a `powerwash` but I'll leave it to you whether you want to add the extra line or not.

longrunningprocess:
:-D, nm, I went ahead and added the clean up did after I thought about it

PR #1251, opened 2021-11-19
update to MongoDB 5.0.4

megahirt:
## Description

Our current version of MongoDB, version 4.0 was released Jule 2018 and is therefore over 3 years old.  This PR upgrades us to the latest MongoDB version 5.0.4 and pins the Docker image.  I read a bit about how an upgrade works, but I think the only way to be sure that the upgrade is successful to to test this with a full production copy on our staging server.

Note that in v5.0 the mongo shell has been renamed from `mongo` to `mongosh`

links:

https://en.wikipedia.org/wiki/MongoDB
https://docs.mongodb.com/manual/release-notes/5.0-compatibility/
### Type of Change

- infrastructure change

## How Has This Been Tested?

If the following prove successful, then we can have some confidence that it is safe to move to MongoDB v5

- [ ] E2E tests pass
- [ ] We do extensive testing in Staging with a full production dataset

### DIFF
--- a/docker/deployment/db-deployment.yaml
+++ b/docker/deployment/db-deployment.yaml
@@ -49,7 +49,7 @@ spec:
     spec:
       containers:
       - name: db
-        image: mongo:4.0
+        image: mongo:5.0.4
         volumeMounts:
         - mountPath: /data/db
           name: data
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -105,7 +105,7 @@ services:
     command: sh -c "postconf -e 'default_transport = retry:no outbound email allowed' && /run.sh"
 
   db:
-    image: mongo:4.0
+    image: mongo:5.0.4
     container_name: lf-db
     ports:
       # exposed this to host for admin tools


longrunningprocess:
everything seems to be ok locally so I'm guessing this has something to do with TC...  the db doesn't spin up quick enough:

```
Starting with configuration:
20:18:23    - Hosts to be waiting for: [db:27017, mail:25, selenium:4444, app-for-e2e:80]
20:18:23    - Timeout before failure: 30 seconds
20:18:23    - TCP connection timeout before retry: 5 seconds
20:18:23    - Sleeping time before checking for hosts availability: 0 seconds
20:18:23    - Sleeping time once all hosts are available: 0 seconds
20:18:23    - Sleeping time between retries: 1 seconds
20:18:23   --------------------------------------------------------
20:18:23   Checking availability of db:27017
20:18:24   Host db:27017 not yet available...
20:18:25   Host db:27017 not yet available...
20:18:27   Host db:27017 not yet available...
20:18:33   Host db:27017 not yet available...
20:18:39   Host db:27017 not yet available...
20:18:45   Host db:27017 not yet available...
20:18:47   Host db:27017 not yet available...
20:18:53   Host db:27017 not yet available...
20:18:53   Timeout! After 30 seconds some hosts are still not reachable
20:18:53   1
```

megahirt:
I'm not up for spending more time on this at the moment, so closing.

PR #1252, opened 2021-11-28
Go to most recently viewed entry

### LINKED ISSUE #1161: When arriving at a project, go to most recently-viewed entry
#1016 suggests going directly to a project on sign-in, but we have learned that it would also be good to go directly to the most-recently-used entry.

Any time the user clicks on a project, in the projects view or the breadcrumbs or by logging in and being redirected to the project, it's a good idea to take them directly to the most-recently-used entry as well. Only when you click on the "←List" button should you actually see the entry list.

Not doing this and leaving the status quo.

**Additional context**
Also related: #1129.

palmtreefrb:
## Description

When arriving at a project in the entry editor, go to most recently-viewed entry. If there is no recently viewed entry, set it to the first entry in the list. 

 100% AngularJS / Indexed DB
 handle multiple projects
 Loading a project always goes to the editor (avoid going to the list by default)
 Load the most recently used entry if exists in the cache
 If no cache is available, load the first entry
 clicking on the "<-- list" button or the "list" breadcrumb explicitly still takes the user to the list

Fixes #1161 

### Type of Change

## Screenshots
Screenshot from 2021-11-29 16-54-16-2

## QA Testing

- [ ] Loading a project will now navigate directly to the entry editor page (entry list view is no longer the default view)
- [ ] The editor will navigate to the last seen entry, if one exists
- [ ] If one doesn't exist, it will navigate to the first entry
- [ ] Can load a project by clicking on the project link from the project menu
- [ ] Can load a project by logging in (go directly to project)
- [ ] Create a brand new project, goes to first entry (thanks @longrunningprocess )
- [ ] Create a brand new project with no entries, goes to list view
- [ ] Delete the project, no side effects

## Checklist:

### DIFF
--- a/src/angular-app/bellows/core/offline/offline-cache-utils.service.ts
+++ b/src/angular-app/bellows/core/offline/offline-cache-utils.service.ts
@@ -29,6 +29,18 @@ export class OfflineCacheUtilsService {
     return this.offlineCache.setObjectsInStore('projects', this.sessionService.projectId(), [obj]);
   }
 
+  getProjectMruEnrtyData(): angular.IPromise<any> {
+    return this.offlineCache.getOneFromStore('projectsmru', this.sessionService.projectId());
+  }
+
+  updateProjectMruEnrtyData(mruEntryId: string): angular.IPromise<any> {
+    const obj = {
+      id: this.sessionService.projectId(),
+      mruEntryId: mruEntryId
+    };
+    return this.offlineCache.setObjectsInStore('projectsmru', this.sessionService.projectId(), [obj]);
+  }
+
   getInterfaceLanguageCode(): angular.IPromise<any> {
     return this.$q.when(this.interfaceStore.getItem<string>(this.INTERFACE_KEY_LANGUAGE_CODE));
   }
--- a/src/angular-app/bellows/core/offline/offline-cache-utils.service.ts
+++ b/src/angular-app/bellows/core/offline/offline-cache-utils.service.ts
@@ -15,7 +15,7 @@ export class OfflineCacheUtilsService {
   constructor(private readonly $q: angular.IQService, private readonly sessionService: SessionService,
               private readonly offlineCache: OfflineCacheService) { }
 
-  getProjectData(): angular.IPromise<any> {
+  getProjectData(): Promise<any> {
     return this.offlineCache.getOneFromStore('projects', this.sessionService.projectId());
   }
 
@@ -29,11 +29,11 @@ export class OfflineCacheUtilsService {
     return this.offlineCache.setObjectsInStore('projects', this.sessionService.projectId(), [obj]);
   }
 
-  getProjectMruEnrtyData(): angular.IPromise<any> {
+  getProjectMruEntryData(): Promise<any> {
     return this.offlineCache.getOneFromStore('projectsmru', this.sessionService.projectId());
   }
 
-  updateProjectMruEnrtyData(mruEntryId: string): angular.IPromise<any> {
+  updateProjectMruEntryData(mruEntryId: string): angular.IPromise<any> {
     const obj = {
       id: this.sessionService.projectId(),
       mruEntryId: mruEntryId
--- a/src/angular-app/bellows/core/offline/offline-cache.service.ts
+++ b/src/angular-app/bellows/core/offline/offline-cache.service.ts
@@ -31,10 +31,19 @@ export class OfflineCacheService {
     }));
   }
 
-  getOneFromStore(storeName: string, key: string): angular.IPromise<any> {
-    return this.$q.when(this.getStore(storeName).getItem(key).then(item => {
+  /*
+  @returns the item if found, otherwise null
+  */
+  async getOneFromStore(storeName: string, key: string): Promise<any> {
+    let item;
+    try {
+      item = await this.getStore(storeName).getItem(key);
       return OfflineCacheService.removeProjectId(item);
-    }));
+
+    } catch (err) {
+      item = null;
+    }
+    return this.$q.when(item);
   }
 
   setObjectsInStore(storeName: string, projectId: string, items: any[]): angular.IPromise<any[]> {
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -31,23 +31,23 @@ jobs:
           github_token: ${{ github.token }}
           files: docker/PhpUnitTests.xml
 
-  e2e-tests:
-    runs-on: ubuntu-latest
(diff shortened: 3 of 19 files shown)


alex-larkin:
I'm really excited about this feature! It should be especially helpful for teams like Margarita's, that are working methodically through the entire lexicon. 

Just to be certain, this will show each user the most recently viewed entry of that user, correct? For example, if I look at "wasi" and sign off, and then Palo looks at "chacra" in the same project, when I get back on I'll see "wasi," right?

palmtreefrb:
> I'm really excited about this feature! It should be especially helpful for teams like Margarita's, that are working methodically through the entire lexicon.
> 
> Just to be certain, this will show each user the most recently viewed entry of that user, correct? For example, if I look at "wasi" and sign off, and then Palo looks at "chacra" in the same project, when I get back on I'll see "wasi," right?

Alex, the functionality here is saving mrv entry "per project" to the "browser cache". Each project will have a mrv in browser cache. This browser cache means saving per PC not User.

alex-larkin:
Ah okay, sounds good. Since typically each user has their own device, this is essentially a "per-user" feature. Two users working on the same device probably wouldn't be deterred if they got taken to their colleague's last-viewed word. And in such cases, it's probable that they would be just operating under a single account anyways. 

Sounds good. 

However this does mean that if someone starts on their phone and switches to their computer, the computer won't load the phone's most recent entry. I do not think this is an issue, though.

Did I understand this right?

Thanks!

palmtreefrb:
> Ah okay, sounds good. Since typically each user has their own device, this is essentially a "per-user" feature. Two users working on the same device probably wouldn't be deterred if they got taken to their colleague's last-viewed word. And in such cases, it's probable that they would be just operating under a single account anyways.
> 
> Sounds good.
> 
> However this does mean that if someone starts on their phone and switches to their computer, the computer won't load the phone's most recent entry. I do not think this is an issue, though.
> 
> Did I understand this right?
> 
> Thanks!

Yes, your understanding is correct.

longrunningprocess:
Seems like there should be some more tests listed out but I trust you all have done all the edge-case testing locally

longrunningprocess (approved):
the code makes sense to me, I'd just like to make sure we have good test coverage for whomever comes behind to testing in staging.

longrunningprocess:
Some possible edge-cases:

- First time user with no projects yet
- A deleted project situation, e.g., a user's MRU project is now gone

megahirt:
> Some possible edge-cases:
> 
> * First time user with no projects yet
> * A deleted project situation, e.g., a user's MRU project is now gone

cool, I updated the QA Testing section with your suggestions

megahirt:
Upon further investigation, the E2E test failure for
```
1) Bellows E2E Projects List app Lexicon E2E Project Access Admin added to project when accessing without membership
  - Failed: Wait timed out after 3011ms
```
is a real failure, which I can reproduce locally.  I will investigate this tomorrow when I'm working.  In general, the tests will need to be modified since we now have new application behavior (go directly to the entry view, not to the list view)

longrunningprocess (approved):

src/angular-app/bellows/core/offline/offline-cache-utils.service.ts
```diff
@@ -29,6 +29,18 @@ export class OfflineCacheUtilsService {
     return this.offlineCache.setObjectsInStore('projects', this.sessionService.projectId(), [obj]);
   }
 
+  getProjectMruEntryData(): Promise<any> {
```
josephmyers, line 32: Feel free to disregard, but it's pretty inconsistent that we've changed a couple of these method signatures. Ideally they're all the same, right? Or we can just not!

src/angular-app/bellows/core/offline/offline-cache.service.ts
```diff
@@ -31,10 +31,19 @@ export class OfflineCacheService {
...
+  */
+  async getOneFromStore(storeName: string, key: string): Promise<any> {
+    let item;
+    try {
+      item = await this.getStore(storeName).getItem(key);
       return OfflineCacheService.removeProjectId(item);
-    }));
+
+    } catch (err) {
+      item = null;
```
josephmyers, line 44: Are we pretty okay with this pattern? If the caller's aren't prepared for null, we will get more exceptions. (Of course, it's better than it was.) Mostly wondering at a broader level if we have an established way of handling "error" cases.

megahirt:
Sadly, I have made the difficult decision to stop further runs
of the E2E tests on GHA.  They pass on my machine but
fail on GHA, and so they cannot be relied on.  We will hope
that the re-write in Cypress will pay off down the road.

src/angular-app/bellows/core/offline/offline-cache.service.ts
megahirt: This is a good point.  I am OK with this pattern given that this is the recommended way to use the 3rd party library.  I would like to understand more about what options we have to handle these situations and what we would recommend.

src/angular-app/bellows/core/offline/offline-cache-utils.service.ts
megahirt: I agree, thank you for pointing this out!  Since the plan is to abandon this code, I mostly only interested in fixing code that is broken, instead of refactoring code that is apparently working.

src/angular-app/bellows/core/offline/offline-cache.service.ts
megahirt: I am going to resolve this thread since it is not requesting changes and is not intended to block the PR.

src/angular-app/bellows/core/offline/offline-cache-utils.service.ts
megahirt: I am going to resolve this thread since it is not requesting changes and is not intended to block the PR.

palmtreefrb:
I have tested in QA, every check item passed.

longrunningprocess:
I ended up with an empty entry box after deleting an entry, closing the tab and re-opening the tab...is that ok?

I was expecting some errorhandling to take over and load an entry.

image

palmtreefrb:
Billy, had you ever encountered this error before?

longrunningprocess:
Not that I can recall but I also don't do a lot of testing around deleting entries.

megahirt:
@longrunningprocess this is definitely a bug in the current code.  It should not be possible to reach this state.  The check "If one doesn't exist, it will navigate to the first entry" should apply here.  Thank you for your testing which found this!

PR #1253, opened 2021-11-29
Correct scroll sync issues

### LINKED ISSUE #1014: entry list does not auto-scroll properly on page load when deep linking to entry
In Language Forge, when viewing an entry, the URL reflects a link to display that particular entry.  This makes it possible to share a link to a particular entry in Language Forge.

Here is an example URL:

`https://languageforge.org/app/lexicon/[projectid]/#!/editor/entry/[entryid]`

On loading of the editor, the item list attempts to scroll automatically to the place in the list where the entry is, however the scroll feature is flaky and often does not result in the selected entry being in the viewable area of the scroll list.

The item is selected in blue in the list below but it is off the visible area of the scrollable div, thus confusing users.

image
(no entry appears selected, even though it is)

The auto-scroll logic is buggy and doesn't seem to work anymore.  The relevant methods `scrollListToEntry()` and `scrollDivToId()` can be found in [editor.component.ts](https://github.com/sillsdev/web-languageforge/blob/master/src/angular-app/languageforge/lexicon/editor/editor.component.ts#L1158)

Suggestion: rewrite these two methods from scratch.

A non-functional item-list auto-scroll affects several features:
- highlight entry in list when deep linking to entry (this issue)
- highlight entry in list when clicking "show all" after a filtered list

### LINKED ISSUE #1130: Keep selected term visible on word list
When a user searches for a word, clicks on it, and clears the search query, the list on the left goes back to the very top.  This is not useful for navigating to other entries near the word in question. This is important for those who want to browse in a certain part of the dictionary, but don't want to scroll from the top. 

Demonstration on desktop:
https://user-images.githubusercontent.com/82162128/132067253-45c79b95-db9e-47f8-a0aa-5fe1431d464d.mp4

Demonstration on mobile:
https://user-images.githubusercontent.com/82162128/132068982-fcf42d23-339f-4455-b6a8-cd9b4df5515c.mp4

The current behavior is:
1. user searches for term
2. user selects term
3. user hits the "X" (or back arrow on mobile)
**4. The list of entries goes back to the very top**

The desired behavior is:
1. user searches for term
2. user selects term
3. user hits the "X" (or back arrow on mobile)
**4. the entry that the user selected stays exactly in the same position as it was in step 2. The words that are before and after it alphabetically show up above and below it.** 

So in my desktop example above, this is what I would expect to see after hitting the X: 
image

### LINKED ISSUE #1133: scrollListToEntry() isn't working reliably
In [`editor.component.ts` there is a private function called `scrollListToEntry()`](https://github.com/sillsdev/web-languageforge/blob/9f6519e93d9617f10cc9f689fa32f6bb0dddf603/src/angular-app/languageforge/lexicon/editor/editor.component.ts#L1158) that isnt' working as intended.  The purpose of the method is to scroll the item list so that the selected entry is visible.  A selected entry can easily scroll off the list or become not visible at all whenever searches are finished, are cleared, or whenever new entries are added (either by the user or by others).  This method should reliably manipulate the scrollable area so that the highlighted entry is always in view after the method returns.

This method is implemented in a certain way, however there may be a simpler way to implement this using more modern CSS tricks.  I am not sure.  My recommendation is to understand what this and its calling methods do, and then rewrite the method from scratch.  This may be the best way to make it more reliable than it currently is.

This method is complicated by the infinite scrolling feature whereby the list does not actually contain all items of the dictionary at startup. So it must "loadMoreEntries()" until the entry has been loaded into the list for it to be scrollable.

The main work of scrollListToEntry() is done in [scrollDivToId()](https://github.com/sillsdev/web-languageforge/blob/9f6519e93d9617f10cc9f689fa32f6bb0dddf603/src/angular-app/languageforge/lexicon/editor/editor.component.ts#L1158).

Having a reliable scrollListToEntry() should fix #1130 and also #1014 

There is also the possibility that we can make this method simpler if the research in #1127 proves that we can simply load all entries for most dictionaries.

longrunningprocess:
## Description

There are a number of scenarios where the user can be left disoriented by the fact that the entry that is selected is not visible in the full list of entries on the left.  Some of the use cases are outlined in the comments of https://github.com/sillsdev/web-languageforge/issues/1133

Fixes #1133 Fixes #1130 Fixes #1014 

### Type of Change

## Screenshots

Video's demoing the fixes are already captured in the issue comments.

## How Has This Been Tested?

- [ ] list scrolls to entry when PgUp / PgDn keyboard shortcuts are used
- [ ] list scrolls to entry when Prev / Next buttons are used
- [ ] list scrolls to entry when refreshing page on an entry that is down the list (keeping editor state)
- [ ] When clearing the filter via "show all" button, list scrolls to selected entry
- [ ] When clearing text in the search box, list scrolls to selected entry

## Checklist:

### DIFF
--- a/src/angular-app/languageforge/lexicon/editor/editor.component.ts
+++ b/src/angular-app/languageforge/lexicon/editor/editor.component.ts
@@ -451,7 +451,7 @@ export class LexiconEditorController implements angular.IController {
 
   editEntryAndScroll(id: string): void {
     this.editEntry(id);
-    this.scrollListToEntry(id, 'middle');
+    this.scrollListToEntry(id);
   }
 
   editEntry(id: string): void {
@@ -477,7 +477,7 @@ export class LexiconEditorController implements angular.IController {
   skipToEntry(distance: number): void {
     const i = this.editorService.getIndexInList(this.currentEntry.id, this.visibleEntries) + distance;
     this.editEntry(this.visibleEntries[i].id);
-    this.scrollListToEntry(this.visibleEntries[i].id, 'middle');
+    this.scrollListToEntry(this.visibleEntries[i].id);
   }
 
   newEntry(): void {
@@ -1218,9 +1218,7 @@ export class LexiconEditorController implements angular.IController {
     });
   }
 
-  private scrollListToEntry(id: string, position: string): void {
-    // const posOffset = (position === 'top') ? 274 : 487;
-
+  private scrollListToEntry(id: string, alignment: string = 'center'): void {
     const entryDivId = '#entryId_' + id;
     const listDivId = '#compactEntryListContainer';
     let index = this.editorService.getIndexInList(id, this.filteredEntries);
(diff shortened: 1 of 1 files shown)


megahirt (approved):
This is a fantastic PR that squashes some bugs that have been out there for years.  Many thanks for also finding some methods that were async but not explicitly set as such.  This should help with weird behavior as well.  Great job!

megahirt:
merging per conversation with @longrunningprocess

longrunningprocess:
> This is a fantastic PR that squashes some bugs that have been out there for years.  Many thanks for also finding some methods that were async but not explicitly set as such.  This should help with weird behavior as well.  Great job!

Thanks @megahirt , it definitely gets the job done.  I would've preferred not to compromise on the solution but the `reload` is reliable.

megahirt:
@longrunningprocess This is such a great improvement that I wanted to get this in with the current deployment that we are testing on QA, so I did a bit of extra git to move it into staging.  I'm very excited about this.

In my testing, I found another use case that isn't current working that I can reproduce.  

I've added it to the "tested in staging checkbox above"
- [ ] List scrolls to entry when first starting from list view, e.g. entry near bottom of list

scroll-error

Here are the repro steps:

1) go to my projects and click on the project name
2) in the list view, click on an entry that is at least 15 down in the list (so it must scroll)
3) see that it doesn't scroll and there is a console log error
image

This only appears to happen when the app hasn't loaded before, so if you go to the list a second time without leaving the project, the scroll to entry works as designed.

longrunningprocess:
good find, thank you!  I'll look at that today and plan on creating a PR against staging for your review.

longrunningprocess:
bummer, I'm having a hard time reproducing it locally at least...mine is with a non-S/R type project.  I'll add that same project you used and see if the problem has something to do with S/R projects.

longrunningprocess:
having some trouble getting S/R to work locally but I was not able to reproduce in QA either...maybe I'm missing something in the use case set up?

https://user-images.githubusercontent.com/4412848/144298695-8a066850-2b21-4e81-8da7-4a76c93c308c.mov

longrunningprocess:
ok, I'm getting it now.  I had to throttle the network down to a fast 3G and that does it.

longrunningprocess:
made an attempt at this in here:  https://github.com/sillsdev/web-languageforge/commit/05270d180eecd603dda80294fdc03dc552ae0362

All other scenarios still test out fine.  I couldn't find a situation where that `is(':visible')` check was ever true so I just removed it.  If you want to give this some more rigorous testing we can do that.  Just let me know how you want to proceed @megahirt .

PR #1258, opened 2021-11-30
Utilize GHA for unit tests

longrunningprocess:
## Description

I'd like to utilize GHA instead of TeamCity for unit tests since GHA proves to be more reliable.

### Type of Change

- engineering

## Screenshots

N/A

## How Has This Been Tested?

- [x] Opening a pull request runs the workflow
- [x] Pushing new commits to an open PR runs the workflow
- [ ] Merging the PR to develop runs the workflow

## Checklist:

- [x] I have turned off any TC integrations with this repo
- [x] I have disabled the TC unit test job and any associated configs


.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,20 @@
+name: Run unit tests
+
+# https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#on
+on:
+  pull_request:
+
+  push:
+    branches:
+      - develop
```
megahirt, line 9:
```suggestion
      - develop
      - staging
```
Also run tests on PRs into staging

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  unit-tests:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Run unit tests
+        working-directory: ./docker/
+        run: make unit-tests
```
megahirt, line 20: Add Unit Test Reporter (copied this from my own GHA branch)

```suggestion
        run: make unit-tests
        
# https://github.com/EnricoMi/publish-unit-test-result-action
      - name: Publish Test Results
        uses: docker://ghcr.io/enricomi/publish-unit-test-result-action:v1
        if: always()
        with:
          github_token: ${{ github.token }}
          files: docker/PhpUnitTests.xml
```

megahirt (approved):
Approving as is, with the suggestion to also include staging branch and a Test Reporter (from my old GHA branch)

megahirt (approved):

.github/workflows/unit-tests.yml
longrunningprocess: 👍🏻

.github/workflows/unit-tests.yml
longrunningprocess: seems harmless enough...if this ends up breaking any builds, i.e., false-negative, it won't be worth keeping it.

longrunningprocess:
I pulled the reporting piece back out @megahirt , looks like it still needs more work to get the placement of that file correct...not sure, but we definitely don't want that unit test report to end up in our final image we deploy :-). I've extracted it in https://github.com/sillsdev/web-languageforge/pull/1258/commits/16154cbb02c7425478ba4ce57168e2794eff84cb, please feel free to incorporate it later.

megahirt:
> I pulled the reporting piece back out @megahirt , looks like it still needs more work to get the placement of that file correct...not sure, but we definitely don't want that unit test report to end up in our final image we deploy :-). I've extracted it in https://github.com/sillsdev/web-languageforge/pull/1258/commits/16154cbb02c7425478ba4ce57168e2794eff84cb, please feel free to incorporate it later.

It sounds like the reporter didn't work.  Probably the wrong working directory.

> we definitely don't want that unit test report to end up in our final image we deploy :-).

This doesn't make sense to me.  The PHP test image never gets published or deployed.  It's purpose is to run tests and then be discarded.  The test output XML is ephemeral.

longrunningprocess:
> It sounds like the reporter didn't work. Probably the wrong working directory.

yeah, I'm sure it's something simple to fix

> This doesn't make sense to me. The PHP test image never gets published or deployed. Its purpose is to run tests and then be discarded. The test output XML is ephemeral.

I thought I had seen that output hang around before and then I saw the volume mapping in this PR...lemme check real quick and make sure it doesn't make it to any of our deployed images

longrunningprocess:
ok, I checked, my bad, I've probably just seen it in local development then, sorry for bringing it up.

PR #1264, opened 2021-12-07
Release 1.11.1

### LINKED ISSUE #1133: scrollListToEntry() isn't working reliably

longrunningprocess:
## Description

Fixes #1133 

### Type of Change

## Screenshots

See #1261 

## How Has This Been Tested?

See #1261 

## Checklist:

### DIFF
--- a/src/angular-app/languageforge/lexicon/editor/editor.component.ts
+++ b/src/angular-app/languageforge/lexicon/editor/editor.component.ts
@@ -1219,13 +1219,7 @@ export class LexiconEditorController implements angular.IController {
   }
 
   private scrollListToEntry(id: string, alignment: string = 'center'): void {
-    const entryDivId = '#entryId_' + id;
-    const listDivId = '#compactEntryListContainer';
     let index = this.editorService.getIndexInList(id, this.filteredEntries);
-
-    // make sure the item is visible in the list
-    // todo implement lazy "up" scrolling to make this more efficient
-
     // only expand the "show window" if we know that the entry is actually in
     // the entry list - a safe guard
     if (index != null) {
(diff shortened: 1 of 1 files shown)


megahirt (approved):
Approved.  You have my permission to merge staging into master and cut a release when there is new stuff to release.  No PR necessary.  This code has already been reviewed.  Full confidence in you :)

PR #1272, opened 2022-01-07
normalize search string to NFC before comparison

### LINKED ISSUE #1244: Convert search terms to NFC
There may be languages or scenarios where a user will enter a search term and either match/miss entries because of incompatible UNICODE encodings, e.g., NFC or NFD.  I'm not 100% sure whether this problem exists in LF or not but thought it would be good to research this possibility more in case there are users out there working in languages and having a lot of difficulty as a result of LF not "normalizing" inputs against the data being searched against.

This thought and a possible situation demonstrating a failing scenario first came up in discussions here:  https://github.com/sillsdev/web-languageforge/pull/1243#issuecomment-964789221

megahirt:
## Description

Normalize the search string to NFC since all data in LF is normalized to NFC on disk.  This allows for exact match or ignore diacritic queries to work regardless of form or language, e.g. Korean.

A note about this fix:
- All data is normalized to NFC in the database on write.  It's been this way for years.
- @longrunningprocess 's addition in #1243 normalized the query to NFD for the purposes of removing diacritics from the data and query.  This a fine approach.
- This PR could have chosen to normalize all data to NFD for comparison under all circumstances given the second point above, however I chose to stick with NFC since that is what the data is underneath.  Either way works.

Fixes #1244

### Type of Change

## Tests and Test Data

Consider the single Korean character below:

**감=감**

To the left of the equals sign is the NFC single composed character.  To the right is the NFD decomposed form (3 code points).  They are canonically equivalent and should display identically where Korean is properly supported.  Interestingly, my Windows machine isn't rendering the NFD portion correctly, as seen in the character identifier screenshot below.  My web browser displays it just fine, as you see it in this PR description.
image

### Test 1 - Query match with "match diacritics"
Steps:
1. Paste the NFC character 감 (left side) into a LF entry data field
2. Do a search using the NFC character 감 to verify a match
3. Do a search using the NFD characters 감 to verify a match  (**currently this fails on production**)

### Test 2 - Query match with "ignore diacritics" (default behavior)
Steps:
1. Paste the NFC character 감 (left side) into a LF entry data field
2. Do a search using the NFC character 감 to verify a match
3. Do a search using the NFD characters 감 to verify a match

### Screencast demo from this branch
Animation2

## Checklist:

- [x] The tests above demonstrate my fix is effective or that my feature works

### DIFF
--- a/src/angular-app/bellows/core/offline/editor-data.service.ts
+++ b/src/angular-app/bellows/core/offline/editor-data.service.ts
@@ -457,7 +457,7 @@ export class EditorDataService {
     // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Unicode_Property_Escapes
     // https://unicode.org/reports/tr44/#Diacritic
     // https://stackoverflow.com/a/37511463/10818013
-
+// Convert to NFD in order to remove any code points with the property 'Diacritic', then convert back to NFC for comparison
     return input.normalize('NFD').replace(/\p{Diacritic}/gu, '').normalize('NFC');
   }
 
--- a/src/angular-app/bellows/core/offline/editor-data.service.ts
+++ b/src/angular-app/bellows/core/offline/editor-data.service.ts
@@ -458,12 +458,12 @@ export class EditorDataService {
     // https://unicode.org/reports/tr44/#Diacritic
     // https://stackoverflow.com/a/37511463/10818013
 
-    return input.normalize('NFD').replace(/\p{Diacritic}/gu, '')
+    return input.normalize('NFD').replace(/\p{Diacritic}/gu, '').normalize('NFC');
   }
 
   private entryMeetsFilterCriteria(config: any, entry: LexEntry): boolean {
     if (this.entryListModifiers.filterText() !== '') {
-      const rawQuery = this.entryListModifiers.filterText()
+      const rawQuery = this.entryListModifiers.filterText().normalize('NFC');
       const normalizedQuery = this.entryListModifiers.matchDiacritic ? rawQuery : this.removeDiacritics(rawQuery);
       const regexSafeQuery = this.escapeRegex(normalizedQuery);
       const queryRegex = new RegExp(this.entryListModifiers.wholeWord ? `\\b${regexSafeQuery}\\b` : regexSafeQuery, 'i');


src/angular-app/bellows/core/offline/editor-data.service.ts
```diff
@@ -458,12 +458,12 @@ export class EditorDataService {
     // https://unicode.org/reports/tr44/#Diacritic
     // https://stackoverflow.com/a/37511463/10818013
 
-    return input.normalize('NFD').replace(/\p{Diacritic}/gu, '')
+    return input.normalize('NFD').replace(/\p{Diacritic}/gu, '').normalize('NFC');
```
longrunningprocess, line 461: if changing the first `normalize('NFD')` to `'NFC'` would eliminate the need for your extra call here, I think it would be easier to understand "the why" if we come back to this code for any reason 6 months from now.  If it won't and we need both calls, I think we should either put a link to the discussion in a comment here as an answer to "why?" this code required both.

longrunningprocess (approved):
one little thought to simplify things if possible.

src/angular-app/bellows/core/offline/editor-data.service.ts
megahirt: We have to normalize to NFD first in order to get at the diacritic as a separate code point.  If you normalize to NFC first, then the `/\p{Diacritic}/` will never match anything (unless someone types a diacritic in isolation, which doesn't make sense).

Here is a working example from the JS console:
image

And here is the character split up using Character Identifier:
image

`/\p{Diacritic}/` matches U+0308 because it has the Unicode property 'Diacritic'.  When in NFC form, the single character does NOT have the Unicode property 'Diacritic'.

src/angular-app/bellows/core/offline/editor-data.service.ts
```diff
@@ -458,12 +458,12 @@ export class EditorDataService {
     // https://unicode.org/reports/tr44/#Diacritic
     // https://stackoverflow.com/a/37511463/10818013
```
megahirt, line 460:
```suggestion
// Convert to NFD in order to remove any code points with the property 'Diacritic', then convert back to NFC for comparison
```

src/angular-app/bellows/core/offline/editor-data.service.ts
longrunningprocess: ok, thanks for explaining that nuance!  This stuff is a little crazy 🤪

longrunningprocess (approved):

PR #1284, opened 2022-01-18
Implement CD for production

### LINKED ISSUE #1277: bug: cache-busting not working in staging
**Describe the bug**

Requests for assets in the staging environment should have a version referenced in the request that pertains to the image tag but they are currently `v9.9.9`.

**To Reproduce**
1. Go to https://qa.languageforge.org/auth/login
2. Open DevTools and ensure the cache is not enabled
3. Refresh the page
4. You should see requests to assets with the `v9.9.9` querystring on them.

**Expected behavior**
In this specific case, these should've been something along the lines of `v20220113-8faf84a3c51d757231d060ef329bc0a51c3aed5c`

**Screenshots**
image

### LINKED ISSUE #1279: feat:  begin automating deployment to production
Human/manual involvement in the delivery of our production software can be prone to missteps and errors.

We testing out an automated deployment for staging and have been satisfied with it.

N/A

**Additional context**
This effort should not only establish the automation but it should also update the README and should consider/document the process of reverting to a previous version of the app if needed.

longrunningprocess:
## Description

This PR aims to automate deployment of the LF app upon the creation of a tag (`v#`-type convention) on `master` typically.

Fixes #1279 
Fixes #1277 

### Type of Change

- Engineering

## Testing on your branch

- [x] Simulated tests run against staging being triggered by a `billy*` tag
- [ ] Merging to `develop` will be the first real test for staging.
- [ ] Tagging a commit on `master` with a `v*` tag will be the first real test for production.

## Checklist

- [x] Ensure the addition of `wait` to the production image is discussed with team
- [x] change yml to trigger off of `v*` tag
- [x] ensure the README is updated
- [x] Ensure the process of reverting to a previous version of the app is documented


longrunningprocess:
@megahirt  I pulled out the layer caching on https://github.com/sillsdev/web-languageforge/pull/1284/commits/2fd7d73db1a28ba7676678e7f9610d2223e7aba8 after checking his stuff a little closer.  He's still below a major release and already had some breaking changes.  Additionally, between his job run and the post run, it's really not saving us that much time.  I'd like to keep it out but let's talk if you want it back in.

docker/app/Dockerfile
```diff
@@ -53,6 +53,7 @@ FROM sillsdev/web-languageforge:base-php AS production-app
 RUN rm /usr/local/bin/install-php-extensions
 RUN apt-get remove -y gnupg2
 RUN mv $PHP_INI_DIR/php.ini-production $PHP_INI_DIR/php.ini
+COPY --from=sillsdev/web-languageforge:wait-latest /wait /wait
```
megahirt, line 56: I guess this is an unfortunate requirement to make the e2e tests run off a production image.  I think it's a reasonable compromise, but we should add a comment in the Dockerfile stating why we need "wait" in production.

.github/workflows/production.yml
```diff
@@ -0,0 +1,19 @@
+name: Deploy to production
+
+# https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#on
+on:
+  push:
+    tags:
+      - v*
```
megahirt, line 7: I am concerned that this trigger is too permissive.   Would this allow anyone with repo push access to be able to deploy to production?  What we want is a trigger for tags which are pushed only to master.  We should be able to protect master branch sufficiently.

I could not figure out how to boolean AND workflow events (e.g. tag AND branch = master) but this SO describes one approach:  https://stackoverflow.com/questions/59874409/trigger-github-workflow-only-if-pushed-to-specific-branch-and-tag-exists

Something like (untested):
```
 - name: Exit if not on master branch
        if: github.ref != 'refs/heads/master')
        run: exit -1
```

docker/app/Dockerfile
```diff
@@ -62,8 +63,7 @@ RUN mv $PHP_INI_DIR/php.ini-development $PHP_INI_DIR/php.ini
 COPY --from=sillsdev/web-languageforge:wait-latest /wait /wait
 
 FROM ${ENVIRONMENT}-app AS languageforge-app
-ARG BUILD_VERSION=9.9.9
-ENV BUILD_VERSION=${BUILD_VERSION}
```
megahirt, line 66: Is the ENV BUILD_VERSION no longer needed?

megahirt (changes requested):
@longrunningprocess this is really nice work and I LOVE how you have refactored the GHA and moved the common pieces into "integrate and deploy" making the staging and production YML files very slim.

I would like to come to a resolution on the permissive trigger before approving this.  We can talk through it if you have questions about what I wrote.  Thanks again.

docker/app/Dockerfile
longrunningprocess: yep, there might be a better way to address this but I can't think of attack vectors this presents so let's go with it.  I'll push a commit with the comment.

.github/workflows/production.yml
longrunningprocess: yeah, I presume we'd be protected in our repo setting but I much prefer the idea of an extra check here, I'll look into more and if I can't find something conventional from GH's docs, I'll use this exit approach.

docker/app/Dockerfile
longrunningprocess: not as far as I could tell.  I'm not 100% sure it would've been there in the first place, maybe you have some history on it?

.github/workflows/production.yml
longrunningprocess: we'd have to make sure to get the property right, I just did a little test over in my sandbox and there might be a prop to go after that would be the one to check:  https://github.com/longrunningprocess/gha-sandbox/runs/4861490865?check_suite_focus=true#step:2:212

longrunningprocess: I opened a ticket with GH just to make sure there isn't a built-in option but I'll go ahead and put in a fix for this today.

longrunningprocess: I tested this approach out in my own sandbox so it should be good but we should plan on keeping an eye on it the next release we cut.  Fix in https://github.com/sillsdev/web-languageforge/pull/1284/commits/56308a9192616f08bd7e1f8887fd80c901c99bd2

docker/app/Dockerfile
megahirt: ok thanks.

.github/workflows/production.yml
megahirt: this looks good, thanks!

megahirt (approved):
Looks good!  The syntax for limiting to master branch is cleaner than what I found

Issue #1189, opened 2021-10-01
feat: automatically send an email to issues@languageforge.org whenever a project goes on HOLD


megahirt:
**Is your feature request related to a problem? Please describe.**
Projects on HOLD aren't getting enough developer attention.  Firing off an email is one way to make sure projects don't silently fail.

**Describe the solution you'd like**
The system should send an email whenever a project goes on hold.
An even better solution would be to use the github api to create an issue automatically :)
https://docs.github.com/en/rest/reference/issues#create-an-issue

megahirt:
This issue is actually more about keeping on top of projects that get stuck, since we don't have a LFMerge dashboard to display the status of these projects

PR #1285, opened 2022-01-20
chore: lfmerge-copy-state Makefile target

megahirt:
## Description

Add a Makefile target that copies all Send/Receive state files to the developer's machine, and then copy projects ONHOLD into a separate folder for further inspection.

This PR is intended to commit utility code that I used to conveniently get state files onto my machine for local inspection.  It is not intended as a long-term fix or implementation.  We should instead [write an email handler for whenever a project goes on HOLD](https://github.com/sillsdev/web-languageforge/issues/1189).

### DIFF
--- a/docker/deployment/.gitignore
+++ b/docker/deployment/.gitignore
@@ -1,1 +1,3 @@
 *-current.yaml
+all_projects
+on_hold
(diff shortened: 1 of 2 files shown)


longrunningprocess (approved):
looks like some real script🥋  going on there.  I'll trust it worked for you.

PR #1300, opened 2022-02-08
chore: move custom base-php files to proper location

### LINKED ISSUE #1299: Config files located in wrong place
**Describe the bug**
There was some work to refactor our `base-php` image and some files were accidentally left in the wrong folder.  See https://github.com/sillsdev/web-languageforge/pull/1273

**Expected behavior**
`docker/app/customizations.php.ini` and `docker/app/000-default.conf` should be located in the `docker/base-php` folder and `docker/base-php/Dockerfile` should updated to grab those locally.

megahirt:
## Description

Move the files referenced by the base-php image to the base-php folder.  They used to be part of the app image, but when we moved to a base-php image build, we forgot to move the custom files.  This commit puts them in the appropriate place.

Note, I do not believe that it is necessary to rebuild the base-php image as this change does not involve any functional difference to the existing image.

Fixes #1299

## Checklist


longrunningprocess (approved):
I agree, no need to rebuild.  Thanks for catching that lfmerge script too!

PR #1306, opened 2022-02-14
re-enable e2e tests

megahirt:
Re-enable e2e tests, but we won't require them.  We can still run them against the last-know-e2e tag.  I also won't use the e2e test reporter, since it is expected to report failures.


josephmyers (approved):

megahirt:
OK, now that I see the red X next to my PR even though the status check isn't required, I repent of this PR and will be closing without merging

PR #1315, opened 2022-02-28
Run lfmerge in separate container

rmunn:
## Description

Docker-compose and Kubernetes setups for creating a separate lfmerge container.

Docker-compose setup has been tested. Kubernetes setup is still undergoing testing.

## Checklist

### DIFF
--- a/docker/deployment/app-deployment.yaml
+++ b/docker/deployment/app-deployment.yaml
@@ -192,6 +192,22 @@ spec:
             value: production
           - name: LFMERGE_LOGGING_DEST
             value: syslog
+          - name: LFMERGE_BASE_DIR
+            value: /var/lib/languageforge/lexicon/sendreceive
+          - name: LFMERGE_WEBWORK_DIR
+            value: webwork
+          - name: LFMERGE_TEMPLATES_DIR
+            value: Templates
+          - name: LFMERGE_MONGO_HOSTNAME
+            value: db
+          - name: LFMERGE_MONGO_PORT
+            value: 27017
+          - name: LFMERGE_MONGO_MAIN_DB_NAME
+            value: scriptureforge
+          - name: LFMERGE_MONGO_DB_NAME_PREFIX
+            value: sf_
+          - name: LFMERGE_VERBOSE_PROGRESS
+            value: true
           - name: MONGODB_CONN
             valueFrom:
               secretKeyRef:
--- a/docker/deployment/app-deployment.yaml
+++ b/docker/deployment/app-deployment.yaml
@@ -179,8 +179,7 @@ spec:
                 key: LD_TRUST_TOKEN
                 name: ld-trust-token
       - name: lfmerge
-        # TODO: Decide on a verion-numbering policy for LfMerge releases
-        image: ghcr.io/sillsdev/lfmerge:latest
+        image: ghcr.io/sillsdev/lfmerge:2.0.98
         imagePullPolicy: Always
         volumeMounts:
           - mountPath: /var/lib/languageforge/lexicon/sendreceive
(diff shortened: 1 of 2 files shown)


docker/deployment/app-deployment.yaml
```diff
@@ -178,3 +178,37 @@ spec:
               secretKeyRef:
                 key: LD_TRUST_TOKEN
                 name: ld-trust-token
+      - name: lfmerge
+        # TODO: Decide on a verion-numbering policy for LfMerge releases
+        image: ghcr.io/sillsdev/lfmerge:latest
+        imagePullPolicy: Always
+        volumeMounts:
```
longrunningprocess, line 185: if there are strict volume permissions, we'll need to make sure we have an init container to get those correct.

longrunningprocess (changes requested):
I'm really excited about this, can't wait to see it all working.

I'd also like to see individualized scripts in the deployment Makefile to for lfmerge... you should be able to see the convention in there, if not just ping me.

docker/deployment/app-deployment.yaml
```diff
@@ -178,3 +178,37 @@ spec:
               secretKeyRef:
                 key: LD_TRUST_TOKEN
                 name: ld-trust-token
+      - name: lfmerge
+        # TODO: Decide on a verion-numbering policy for LfMerge releases
+        image: ghcr.io/sillsdev/lfmerge:latest
```
longrunningprocess, line 183: excellent!  How did you like ghcr as a repo?

docker/deployment/app-deployment.yaml
```diff
@@ -178,3 +178,37 @@ spec:
               secretKeyRef:
                 key: LD_TRUST_TOKEN
                 name: ld-trust-token
+      - name: lfmerge
+        # TODO: Decide on a verion-numbering policy for LfMerge releases
```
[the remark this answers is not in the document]
rmunn: Finally got back to this. I've updated the LfMerge build process so that it tags the Docker images with a version number that increments with every push to the `live` branch (the LfMerge branch scheme is currently `master` / `live` where Language Forge uses `develop` / `master`; I'll rename those branches soon, but haven't done so yet). The LfMerge build process also tags the Git repo with the same tags, so that it will be easy to tell which commit the Docker image was built from.

With that in place, I can finally get rid of this TODO. I've changed the lfmerge image tag in the Kubernetes deployment to lock it to a specific version. When we want to deploy a new lfmerge version, we'll just need to edit the Kubernetes deployment file and re-run `kubectl apply`. At some point I'll put that into the Makefile, but I haven't yet reached a decision on how to handle the Makefile commands (e.g. whether to use an `LFMERGE_VERSION` env var or just edit the Kubernetes file each time, and if we use the env var then how should it be calculated, that sort of thing).

But now this PR can move forward.

docker/deployment/app-deployment.yaml
rmunn: Agreed. The init container we currently have should work fine, but I'll work out the issues in the staging environment if it doesn't.

docker/deployment/app-deployment.yaml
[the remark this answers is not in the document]
rmunn: Done in commit dec2d08.

docker/deployment/app-deployment.yaml
rmunn: Very easy to use from GHA workflows. And the fact that you can delete packages with less than 1,000 downloads means I don't have to worry about cluttering up the repo with failed builds if I end up creating any, which is nice.

rmunn:
@longrunningprocess I finally was able to make the changes you requested over a month ago. Please see how you like this now, and update your review. Thanks!

longrunningprocess (approved):

PR #1321, opened 2022-03-09
Create set-backlog-fields.yml

josephmyers:
## Description

Workflow task for automatically adding issues to our project (and Backlog board) on creation.

See my forked repo [commits ](https://github.com/josephmyers/web-languageforge/commits/develop)and [issues ](https://github.com/josephmyers/web-languageforge/issues)for the individual changes, if interested.

### DIFF
--- a/.github/workflows/set-backlog-fields.yml
+++ b/.github/workflows/set-backlog-fields.yml
@@ -1,0 +1,60 @@
+name: set-backlog-fields
+
+on:
+  issues:
+    types: [ opened ] ############ exclude pull requests
+
+jobs:
+  steve:
+    runs-on: ubuntu-latest
+    steps:
+      - name: get backlog field ID
+        run: |
+          echo issue $NUMBER created
+          
+          gh api graphql -f query='
+            query($org: String!) {
+              organization(login: $org){
+                projectNext(number: 1) {
+                  id
+                  fields(first:20) {
+                    nodes {
+                      id
+                      name
+                      settings
+                    }
+                  }
+                }
+              }
+            }'
+        env:
+          NUMBER: ${{ github.event.issue.number }}
+          FIELD: ${{ ??????????????????????????????? }}
+        
+      - name: set fields
+        run: |
+          gh api graphql -f query='
+            mutation(
+              $project:ID!
+              $item:ID!
+              $status_field: ID!
+              $status_value: String!)
+            {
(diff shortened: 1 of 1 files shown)


.github/workflows/set-backlog-fields.yml
```diff
@@ -0,0 +1,103 @@
+name: Add Issue to Project
+
+on:
+  issues:
+    types: [ opened ]
+
+jobs:
+  steve:
```
megahirt, line 8:
```suggestion
  add-issue-to-board:
```
"steve" doesn't sound like the right job name to me :)

.github/workflows/set-backlog-fields.yml
```diff
@@ -0,0 +1,103 @@
...
+          GITHUB_TOKEN: ${{ steps.generate_token.outputs.token }}
+          ISSUE_NUMBER: ${{ github.event.issue.number }}
+          REPO: ${{ toJson(github.event.repository.name) }}
+          ORGANIZATION: sillsdev
+        run: |
+          echo issue $ISSUE_NUMBER added to $REPO
+          
+          gh api graphql -f query='
+            query($issue_number:Int!, $org:ID!) {
+              repository(name:"web-languageforge", owner:$org) {
```
megahirt, line 30:
```suggestion
              repository(name:"${REPO}", owner:$org) {
```
I'm not actually sure this will work given the single and double quote acrobatics here, but I thought it was worth a suggestion.

megahirt (approved):
A few non-blocking suggestions.

You've worked really hard on this PR - Looking forward to seeing it in action!

rmunn (approved):
LGTM.

.github/workflows/set-backlog-fields.yml
rmunn: That's a good idea, since I might want to copy this workflow into the LfMerge repo.

.github/workflows/set-backlog-fields.yml
rmunn: I dunno, I think that job name would apple, I mean apply, to a lot of situations.

.github/workflows/set-backlog-fields.yml
josephmyers: Yeah, it should definitely be noted that the task in its current state applies to "web-languageforge" project "1" and nothing else right now. I plan to change the repo for the final state, but the project number will be slightly more tricky if we start incrementing that.

PR #1328, opened 2022-03-13
Move npm install outside of Docker

### LINKED ISSUE #1323: Refactor npm outside of Docker

megahirt:
## Description

This changes the build to require node on the developer's machine and moves the `npm install` step outside docker (to take advantage of npm caching).  This should result in faster Docker builds on the developer's machine and possibly also on GHA (due to node_module caching on GHA)

Additional notes:
- I needed to rebuild node-sass in Docker since the Docker environment is likely different than the developer host machine
- I removed the `npm run build:dev` from the test-e2e Dockerfile because I discovered it's not needed

### Important Note to Developers
This changes the build in a significant way.  Once this PR is merged, the developer machine must have the appropriate version of node and npm installed locally.  Use Node 14.16.1 and npm 7.6.3  

Fixes #1323 

### Type of Change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist


.github/workflows/build-and-publish-adhoc.yml
```diff
@@ -18,6 +18,15 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      -
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14.16.1'
```
rmunn, line 24: Is there a reason not to use Node 16? E.g., some developer's machine where it's hard to get Node 16 but easy to get 14?

Note that if we do move to Node 16.14.0 specifically (which is the latest Node 16.x at the moment), https://nodejs.org/en/blog/release/v16.14.0/#notable-changes lists something we'll need to be aware of if we import .json files (which right now we only do in Playwright tests AFAIK). Typescript 4.5 or later can handle that syntax: https://devblogs.microsoft.com/typescript/announcing-typescript-4-5/#import-assertions — and since we're using current versions of Typescript in Playwright tests, we should be fine there. So I don't see a drawback to moving to the latest Node version, rather than 14.

P.S. Same comment applies to the other places where we specify the Node version.

docker/test-e2e/Dockerfile
```diff
@@ -7,14 +7,12 @@ RUN npm config set unsafe-perm true && npm install -g npm@7.6.3
 RUN mkdir -p /data
 WORKDIR /data
 
-# unsafe-perm true is required to work around an npm bug since we are running as root, have a git+HTTPS repo source and it has a `prepare` script (perfect storm).
-# see https://github.com/npm/npm/issues/17346
-COPY package.json package-lock.json ./
-RUN npm config set unsafe-perm true && npm install --legacy-peer-deps
+COPY package.json ./
+COPY node_modules ./node_modules/
```
rmunn, line 11: Here the comment about `npm rebuild node-sass` does not apply, so it might be possible to volume-mount `node_modules` here and reduce image size. OTOH, this is an image we don't deploy, so if this image is large it doesn't cost us much. And since this image is going to go away eventually, paying a complexity cost by changing it like that doesn't make as much sense.

rmunn (approved):
Would like to discuss a couple of these items (specifically, the Node version and whether `node_modules` can be volume-mounted) before merging, but I'm okay with this being merged as-is if necessary.

.github/workflows/build-and-publish-adhoc.yml
megahirt: Good point.  I have another branch where I bump Node to 16.14.0 (latest LTS) and that PR will follow as soon as this PR is merged.  Glad to hear about the improvements.

.github/workflows/build-and-publish-adhoc.yml
```diff
@@ -18,6 +18,15 @@ jobs:
...
+          cache: 'npm'
```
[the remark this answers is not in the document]
megahirt: yes, I am pretty sure this will lower build time over the long term.

docker/app/Dockerfile
```diff
@@ -10,16 +10,15 @@ RUN npm config set unsafe-perm true && npm install -g npm@7.6.3
...
-# copy npm cache directory; as optimization for npm install
-COPY --from=sillsdev/web-languageforge:npm-cache /root/.npm /root/.npm
-
 # Copy in files needed for compilation, located in the repo root
 COPY typings ./typings/
-COPY package.json package-lock.json webpack.config.js webpack-dev.config.js webpack-prd.config.js tsconfig.json tslint.json ./
+COPY package.json webpack.config.js webpack-dev.config.js webpack-prd.config.js tsconfig.json tslint.json ./
+
+# Use built node_modules from host
+COPY node_modules ./node_modules/
```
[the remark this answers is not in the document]
megahirt: Because this is a multi-stage build, the `node_modules` folder is actually not included in the final app image build.  This line is part of the ui-builder stage which is discarded.  Only the `dist` folder resulting from ui-builder is persisted in the app image.

docker/test-e2e/Dockerfile
megahirt: Yes, at this point, this image is only built on developer machines and is neither built nor published on GHA.

megahirt (commented):
Thanks for the review, Robin!

longrunningprocess (commented):
great work @megahirt , I'll trust you guys that the `make` and `make dev` still work from a "power-washed" env.  I take it you had fun doing this 😊 future devs will never know how much faster things are now!

PR #1338, opened 2022-03-16
Update set-backlog-fields.yml

josephmyers:
## Description

Trying the SHA, not from the silly Github tutorial, but from [the github-app-token releases page](https://github.com/tibdex/github-app-token/releases)


rmunn (approved):
Double-checked the SHA this time, and it's their latest release. And FWIW, I approve of using the SHA rather than the version number for something with security implications like this.

PR #1342, opened 2022-03-21
Update set-backlog-fields.yml

josephmyers:
## Description

Fixed syntax error and changed applicable board view from Backlog to ProductOwner


megahirt (approved):
LGTM.

Can you also change the job name from "steve" to "add_issue_to_project" ?

PR #1344, opened 2022-03-21
Update set-backlog-fields.yml

josephmyers:
## Description

Adding github token authentication at each step. This wasn't necessary for whatever reason on the private repo.


megahirt (approved):
LGTM

PR #1354, opened 2022-03-21
Optimize creation of login storage state files

rmunn:
We now skip creating the file if it already exists and has a creation date less than 1 year old (since that's the default lifetime of the "Remember Me" cookies that Silex creates). This means that the slowness of logging in five times through the UI (which takes about 20 seconds on a relatively fast developer machine) will only happen once per machine.

## Checklist

### DIFF
--- a/test/e2e/utils/globalSetup.ts
+++ b/test/e2e/utils/globalSetup.ts
@@ -5,6 +5,7 @@ import { testControl } from './jsonrpc';
 import constants from '../../app/testConstants.json';
 import { loginAs } from './login';
 import { usersToCreate } from './userFixtures';
+import * as fs from 'fs';
 
 function createUser(request: APIRequestContext, baseName: string) {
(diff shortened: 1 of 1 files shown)


megahirt (approved):
LGTM thanks Robin.

Per our conversation, can you use the [path variable on line 50 ](https://github.com/sillsdev/web-languageforge/pull/1354/files#diff-63da123e527ab3af470209573ada9e7e512073906773ccd069b95ca3b53be337R50) that you define on line 42?

rmunn:
> LGTM thanks Robin.
> 
> Per our conversation, can you use the [path variable on line 50 ](https://github.com/sillsdev/web-languageforge/pull/1354/files#diff-63da123e527ab3af470209573ada9e7e512073906773ccd069b95ca3b53be337R50) that you define on line 42?

Done.

PR #1358, opened 2022-03-23
Change password playwright tests 

JeanneSon:
## Description
The following tests have been converted to playwright

Bellows E2E Change Password app
✓ setup: login as user, go to change password page
✓ refuses to allow form submission if the confirm input does not match
✓ allows form submission if the confirm input matches
[previously commented out] should not allow a password less than 7 characters
✓ can successfully changes user's password after form submission

For details/remarks regarding converting the above tests, refer to the spreadsheet; https://docs.google.com/spreadsheets/d/1w9wjfm-02QHLhv0ObPAKKlqnDWeag379GIXwOcRU4W0/edit#gid=0

### Type of Change
E2E test

## Checklist

### DIFF
--- a/test/e2e/utils/fixtures.ts
+++ b/test/e2e/utils/fixtures.ts
@@ -4,11 +4,20 @@ import type { Browser, Page } from '@playwright/test';
 import type { usernamesForFixture } from './userFixtures';
 import constants from '../testConstants.json';
 
-export type UserTab = Page & {
+export type UserDetails = {
   username: string,
   password: string,
   name: string,
   email: string,
+}
+
+export type UserTab = Page & UserDetails;
+
+function setupUserDetails(obj: any, username: usernamesForFixture) {
+  obj.username = constants[`${username}Username`] ?? username;
+  obj.name = constants[`${username}Name`] ?? username;
+  obj.password = constants[`${username}Password`] ?? 'x';
+  obj.email = constants[`${username}Email`] ?? `${username}@example.com`;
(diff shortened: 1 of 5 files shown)


rmunn (commented):
LGTM, but I believe I should let Chris review this too before it's merged. Because I've been looking at this code for days, and if there's a flaw in it, I'll miss it because of being too familiar with it by this point.

test/e2e/utils/fixtures.ts
```diff
@@ -4,32 +4,74 @@ import type { Browser, Page } from '@playwright/test';
...
-  tab.name = constants[`${username}Name`] ?? username;
-  tab.password = constants[`${username}Password`] ?? 'x';
-  tab.email = constants[`${username}Email`] ?? `${username}@example.com`;
+  setupUserDetails(tab, username);
   await use(tab);
-  await tab.close();
-  await context.close();
 }
 
 // Extend basic test by providing a "todoPage" fixture.
```
megahirt: we need a different comment here to explain the purpose of the enhanced test fixture functionality

megahirt (approved):
Great job @JeanneSon and @rmunn - you are paving the way for our E2E tests and in this PR we have established a good pattern: 
- we have a beforeAll() which opens the tab
- each test() will re-use the existing tab
- each playwright file will be a unit.  Multiple files should be able to be run in parallel
- test fixtures provide setup data for the application and ease of accessing that metadata

PR #1361, opened 2022-03-29
LFNext infrastructure and change password migration

### LINKED ISSUE #1309: Establish infrastructure for LFNext app
work occurring in https://github.com/sillsdev/web-languageforge/tree/next/changepassword

longrunningprocess:
## Description

This PR introduces a brand new infrastructure and architecture to the LF app.  It enables a capability by capability migration from the existing Angularjs impl to a modern stack in Svelte and Sveltekit.

Fixes #1309 

## Type of Change

## Testing on your branch

- [ ] Change password
1. Log in to the app
1. Choose to change your password
1. Walk through the change password pages
1. Attempt to log back in with a new, different password

- [ ] Register for an account
- [ ] Create a non-S/R project
- [ ] Work with a S/R project
- [ ] Attach an image to an entry (< 60 MB)
- [ ] Attach an audio file to an entry (< 60 MB)
- [ ] Attach a file to an entry (> 60 MB)

## TODO In This PR

- [x] Turn off dark mode
- [x] Deploy out to staging (manually)
- [x] Build out automated staging deployment
- [x] finalize the `TODO:`'s in `staging.yml`
- [x] Determine deployment strategy
- [x] Build out automated production deployment
- [ ] think about whether these adhoc builder actions are needed, e.g., `build-and-publish-next.yml`
- [x] Need a way “back” into the legacy app from the next app
- [x] Test the mobile UI ([screenshots below](https://github.com/sillsdev/web-languageforge/pull/1361#issuecomment-1083261967)).
- [ ] Fix up change password tests to suit
- [x] Use self-signed SSL cert in proxy (local dev)
- [x] update config, move @apply out, move class into elements
- [x] turn e2e tests back on in `integrate-and-deploy.yml`?
- [x] verify whether autoprefixer and the postcss.config are still needed or not.

## Production deployment strategy

1. take note of current images in prod
1. merge to `master`
1. set kube context to `prod` from local machine
1. `make delete` from `deployment` dir
1. cut the release
1. monitor the "Deploy to production" action

## For future PR's

- [ ] Implement the user-logged-in menu
- [ ] mock up new header design and navigation
- [ ] require a strong password?
- [ ] consolidate `make dev` and `make next-dev`

## Checklist

### DIFF
--- a/.github/workflows/build-and-publish-next.yml
+++ b/.github/workflows/build-and-publish-next.yml
@@ -4,6 +4,10 @@ name: Publish LFNext images to Docker Hub on-demand
 on:
   workflow_dispatch:
 
+  # TODO: remove branch trigger before merging into `develop`
+  push:
+    branches:
+      - next/changepassword
 jobs:
   build-n-publish:
     runs-on: ubuntu-latest
--- a/.github/workflows/build-and-publish-next.yml
+++ b/.github/workflows/build-and-publish-next.yml
@@ -4,10 +4,6 @@ name: Publish LFNext images to Docker Hub on-demand
 on:
   workflow_dispatch:
 
-  # TODO: remove branch trigger before merging into `develop`
-  push:
-    branches:
-      - next/changepassword
 jobs:
   build-n-publish:
     runs-on: ubuntu-latest
--- a/.github/workflows/integrate-and-deploy.yml
+++ b/.github/workflows/integrate-and-deploy.yml
@@ -127,6 +127,6 @@ jobs:
       -
   #       run: kubectl --context ${{ secrets.kube-context }} set image deployment/app app=${{ needs.integrate.outputs.IMAGE }}
           # kubectl --context ${{ secrets.kube-context }} set image deployment/app app={{ needs.integrate.outputs.APP }}
-        run: |
-          kubectl --context ${{ secrets.kube-context }} set image deployment/next-proxy app={{ needs.integrate.outputs.IMAGE_PROXY }}
-          kubectl --context ${{ secrets.kube-context }} set image deployment/next-app app={{ needs.integrate.outputs.IMAGE_NEXT_APP }}
+        run: kubectl --context ${{ secrets.kube-context }} set image deployment/next-proxy app={{ needs.integrate.outputs.IMAGE_PROXY }}
+      -
+        run: kubectl --context ${{ secrets.kube-context }} set image deployment/next-app app={{ needs.integrate.outputs.IMAGE_NEXT_APP }}
--- a/.github/workflows/integrate-and-deploy.yml
+++ b/.github/workflows/integrate-and-deploy.yml
@@ -92,12 +92,12 @@ jobs:
     #       docker-compose -f docker/docker-compose.yml up -d app-for-playwright
     #       npx playwright install
     #       npx playwright test
-    #   -
-    #     name: Log in to Docker Hub
-    #     uses: docker/login-action@v1
-    #     with:
-    #       username: ${{ secrets.image-repo-username }}
-    #       password: ${{ secrets.image-repo-password }}
+      -
+        name: Log in to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.image-repo-username }}
+          password: ${{ secrets.image-repo-password }}
       -
         name: Tag images
     #     run: docker tag lf-app ${{ steps.image.outputs.NAMESPACE }}:${{ steps.image.outputs.TAG }}
--- a/.github/workflows/integrate-and-deploy.yml
+++ b/.github/workflows/integrate-and-deploy.yml
@@ -52,41 +52,30 @@ jobs:
           echo ::set-output name=TAG_PROXY::${{ inputs.image-tag-proxy }}
           echo ::set-output name=TAG_NEXT_APP::${{ inputs.image-tag-next-app }}
       -
+        uses: actions/setup-node@v2
+        with:
+          node-version: '16.14.0'
+          cache: 'npm'
+      -
         run: npm ci
       -
-        name: Build app
-        run: docker-compose build --build-arg ENVIRONMENT=production --build-arg BUILD_VERSION=${{ steps.image.outputs.TAG }} app
-      -
-        run:  docker-compose run --rm app cat /var/www/html/build-version.txt /var/www/html/version.php
-      -
-        name: Unit Tests
-        run: make unit-tests-ci
-      -
-        name: Playwright E2E Tests
-        working-directory: .
+        name: Build legacy app
         run: |
-          echo ${{ steps.image.outputs.NAMESPACE }}
-          echo ${{ steps.image.outputs.TAG_APP }}
-          echo ${{ steps.image.outputs.TAG_PROXY }}
-          echo ${{ steps.image.outputs.TAG_NEXT_APP }}
-    #     run: npm ci
-    #   -
-    #     name: Build legacy app
-    #     run: docker-compose build --build-arg ENVIRONMENT=production --build-arg BUILD_VERSION=${{ steps.image.outputs.TAG }} app
-    #   -
-    #     run:  docker-compose run --rm app cat /var/www/html/build-version.txt /var/www/html/version.php
+          docker-compose build --build-arg ENVIRONMENT=production --build-arg BUILD_VERSION=${{ steps.image.outputs.TAG }} app
+          docker-compose run --rm app cat /var/www/html/build-version.txt /var/www/html/version.php
       -
         name: Build "next" images
         run: make build-next
       -
         name: Unit Tests
         run: make unit-tests-ci
-      # -
-      #   name: Playwright E2E Tests
-      #   run: |
-      #     docker-compose up -d app-for-playwright
-      #     npx playwright install
-      #     npx playwright test
+      -
+        name: E2E Tests
+        working-directory: .
+        run: |
+          docker-compose -f docker/docker-compose.yml up -d app-for-playwright
+          npx playwright install
+          npx playwright test
       -
         name: Log in to Docker Hub
         uses: docker/login-action@v1
@@ -120,6 +109,6 @@ jobs:
         uses: sillsdev/common-github-actions/install-kubectl@v1
       -
         run: |
-          kubectl --context ${{ secrets.kube-context }} set image deployment/app app={{ needs.integrate.outputs.IMAGE_APP }}
+          kubectl --context ${{ secrets.kube-context }} set image deployment/app app=${{ needs.integrate.outputs.IMAGE_APP }}
           kubectl --context ${{ secrets.kube-context }} set image deployment/next-proxy next-proxy=${{ needs.integrate.outputs.IMAGE_PROXY }}
           kubectl --context ${{ secrets.kube-context }} set image deployment/next-app next-app=${{ needs.integrate.outputs.IMAGE_NEXT_APP }}
--- a/.github/workflows/integrate-and-deploy.yml
+++ b/.github/workflows/integrate-and-deploy.yml
@@ -81,12 +81,12 @@ jobs:
       -
         name: Unit Tests
         run: make unit-tests-ci
-      -
-        name: Playwright E2E Tests
-        run: |
-          docker-compose up -d app-for-playwright
-          npx playwright install
-          npx playwright test
+      # -
+      #   name: Playwright E2E Tests
+      #   run: |
+      #     docker-compose up -d app-for-playwright
+      #     npx playwright install
+      #     npx playwright test
       -
         name: Log in to Docker Hub
         uses: docker/login-action@v1
--- a/.github/workflows/integrate-and-deploy.yml
+++ b/.github/workflows/integrate-and-deploy.yml
@@ -107,9 +107,9 @@ jobs:
           docker tag lf-next-app ${{ steps.image.outputs.NAMESPACE }}:${{ steps.image.outputs.TAG_NEXT_APP }}
       -
         run: docker images
-    #   -
-    #     name: Publish image
-    #     run: docker push ${{ steps.image.outputs.NAMESPACE }}:${{ steps.image.outputs.TAG }}
+      -
+        name: Publish images
+        run: docker push --all-tags ${{ steps.image.outputs.NAMESPACE }}
 
     # outputs:
     #   IMAGE: ${{ steps.image.outputs.NAMESPACE }}:${{ steps.image.outputs.TAG }}
--- a/.github/workflows/integrate-and-deploy.yml
+++ b/.github/workflows/integrate-and-deploy.yml
@@ -51,11 +51,11 @@ jobs:
           echo ::set-output name=TAG_APP::${{ inputs.image-tag-app }}
           echo ::set-output name=TAG_PROXY::${{ inputs.image-tag-proxy }}
           echo ::set-output name=TAG_NEXT_APP::${{ inputs.image-tag-next-app }}
-      -
-        uses: actions/setup-node@v2
-        with:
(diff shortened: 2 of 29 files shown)


.github/workflows/integrate-and-deploy.yml
```diff
@@ -71,14 +83,21 @@ jobs:
...
-        name: Publish image
-        run: docker push ${{ steps.image.outputs.NAMESPACE }}:${{ steps.image.outputs.TAG }}
+        name: Publish images
+        run: docker push --all-tags ${{ steps.image.outputs.NAMESPACE }}
 
     outputs:
-      IMAGE: ${{ steps.image.outputs.NAMESPACE }}:${{ steps.image.outputs.TAG }}
+      IMAGE_APP: ${{ steps.image.outputs.NAMESPACE }}:${{ steps.image.outputs.TAG_APP }}
+      IMAGE_PROXY: ${{ steps.image.outputs.NAMESPACE }}:${{ steps.image.outputs.TAG_PROXY }}
+      IMAGE_NEXT_APP: ${{ steps.image.outputs.NAMESPACE }}:${{ steps.image.outputs.TAG_NEXT_APP }}
```
megahirt, line 100: I wasn't able to delete the `outputs:` line in my suggestion :( 
```suggestion
```

I'm going on a hunch here that these variables are already in the job output, as defined above.  If they are not, then, I suppose this would simply need to be changed to e.g.
`IMAGE_APP: ${{ steps.image.outputs.IMAGE_APP}}`
It would be nice if that wasn't necessary, but I don't actually know

docker/next-app/Dockerfile.next-app
```diff
@@ -0,0 +1,22 @@
+# Build the app
+FROM node:alpine AS builder
+
+WORKDIR /app
+
+COPY jsconfig.json package-lock.json package.json postcss.config.cjs svelte.config.js tailwind.config.cjs /app
+COPY src /app/src
+COPY static /app/static
+
+RUN npm install
```
megahirt, line 10: after this PR is merged, we can move npm install out of docker

next-app/.npmrc
```diff
@@ -0,0 +1 @@
+engine-strict=true
```
megahirt, line 1: nice!  We should also put this in the root folder to prevent the same sort of issues with wrong node version

megahirt (approved):
Wow, a HUGE PR with a lot of groundwork laid.  Thanks for your work on this - excited to see Svelte and SvelteKit land in the LF project!  Just one non-blocking suggestion about refactoring variable names in the GA yaml, but I haven't tested it so I don't know if it works.

megahirt:
@rmunn and I were having a good laugh at the "good first issue" label you put on here, @longrunningprocess :)

.github/workflows/integrate-and-deploy.yml
```diff
@@ -41,7 +48,9 @@ jobs:
         id: image
         run: |
           echo ::set-output name=NAMESPACE::sillsdev/web-languageforge
-          echo ::set-output name=TAG::${{ inputs.image-tag }}
+          echo ::set-output name=TAG_APP::${{ inputs.image-tag-app }}
+          echo ::set-output name=TAG_PROXY::${{ inputs.image-tag-proxy }}
+          echo ::set-output name=TAG_NEXT_APP::${{ inputs.image-tag-next-app }}
```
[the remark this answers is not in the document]
longrunningprocess: yeah, I like this.  I appreciate you trying to do it for me in the suggestion but your namespace is in the wrong place so I'll just push a commit with these changes.

.github/workflows/integrate-and-deploy.yml
```diff
@@ -71,14 +83,21 @@ jobs:
           username: ${{ secrets.image-repo-username }}
           password: ${{ secrets.image-repo-password }}
       -
-        name: Tag image
-        run: docker tag lf-app ${{ steps.image.outputs.NAMESPACE }}:${{ steps.image.outputs.TAG }}
+        name: Tag images
+        run: |
+          docker tag lf-app ${{ steps.image.outputs.NAMESPACE }}:${{ steps.image.outputs.TAG_APP }}
+          docker tag lf-next-proxy ${{ steps.image.outputs.NAMESPACE }}:${{ steps.image.outputs.TAG_PROXY }}
+          docker tag lf-next-app ${{ steps.image.outputs.NAMESPACE }}:${{ steps.image.outputs.TAG_NEXT_APP }}
```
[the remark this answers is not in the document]
longrunningprocess: see above

.github/workflows/integrate-and-deploy.yml
longrunningprocess: yes, see above

.github/workflows/integrate-and-deploy.yml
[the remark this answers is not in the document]
longrunningprocess: well, this may not be as straight-forward as I was thinking, it might require looking at a different location to make those things available but we also need to be mindful of the fact that those are incoming workflow args...maybe we should pair on this one.

docker/next-app/Dockerfile.next-app
longrunningprocess: I don't think so...we publish this image and then utilize it in our runtime so everything needs to be in the image.

next-app/.npmrc
longrunningprocess: yeah, that's probably a good idea, I like the idea of doing that when we move the Makefile to the root of the project, are you ok with giving yourself a reminder for this one?

.github/workflows/integrate-and-deploy.yml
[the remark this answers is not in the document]
megahirt: I defer to your experience and that you tried to make this work.  satisfied for now.

.github/workflows/integrate-and-deploy.yml
megahirt: satisfied that you tried to make this work.

docker/next-app/Dockerfile.next-app
megahirt: this is just a comment really.  I think we need to discuss our npm and Docker strategy more so we're on the same page.

next-app/.npmrc
megahirt: yup, I'm looking forward to the challenge :)

next-app/src/lib/forms/Input.svelte
```diff
@@ -0,0 +1,36 @@
...
+	function randomId() {
+		return Math.random().toString(36).substring(2, 7)
+	}
+
+	function autofocusIfRequested() {
+		autofocus && input.focus()
+	}
+
+	// works around "svelte(invalid-type)" warning, i.e., can't have a dynamic type AND bind:value...keep an eye on https://github.com/sveltejs/svelte/issues/3921
+	function typeWorkaround(node) {
```
rmunn, line 24: I see the benefit here for using `type=password` yet having just one Input component instead of two. I think for things like checkboxes and radio buttons we'll want components named Checkbox and Radio, and so on, rather than re-using the Input component.

The one thing I'd want to check is whether this can cause a flash of unstyled content, and more importantly, whether there are any browsers that won't take the `type = 'password'` assignment after DOM rendering, and would therefore cause this to act like a regular text input where the password is visible. I don't *think* there are any such browsers, but I'd like some certainty.

The alternative, where we create a duplicate component called PasswordInput that functions almost identically to an Input component, is code duplication, and I'm fine with this solution **if** no browsers exist where it could expose someone's password to a shoulder-surfer. But I don't know if all mobile browsers (mobile Opera, for instance) would do that.

next-app/src/routes/password/changed.svelte
```diff
@@ -0,0 +1,34 @@
+<script context=module>
+	export const prerender = true
```
rmunn, line 2: There might be an issue with prerendering pages like this one; the `getSession` hook might not run when the user clicks on links to the rest of the site. See https://github.com/sveltejs/kit/issues/4426. Though that bug will probably be solved by the time we deploy this.

next-app/src/lib/fetch/server.js
```diff
@@ -0,0 +1,63 @@
...
+				'content-type': 'application/json',
+				cookie,
+			},
+			body: JSON.stringify(body),
+		})
+	} catch {
+		// these only occur for network errors, like these:
+		//	request made with a bad host, e.g., //httpbin
+		//	the host is refusing connections
+		throwError('NETWORK ERROR', 500)
```
rmunn, line 55: At least for the deployment to QA, I think I'd like this to be a slightly more informative error message, i.e. `catch (e)` and then have `throwError` take an optional third parameter that would be the original exception. We might have code in `throwError` that only passes on that exception if the environment is "dev" or "qa", since we don't want to expose possibly-sensitive exception info on prod. But since I expect that the precise exception thrown would be useful in diagnosing any network errors, I'd like to keep it around in some form unless we're in prod.

next-app/src/lib/forms/Button.svelte
```diff
@@ -0,0 +1,14 @@
...
+
+<!-- https://daisyui.com/components/button -->
+<button on:click class:danger class='btn btn-primary'>
+	<slot />
+</button>
+
+<style>
+.danger { @apply
+	btn-error;
+}
```
rmunn, line 13: Does the `btn-error` style override `btn-primary`?

Also, would we want `export let primary` in the future, and/or a few other colors-as-props?

rmunn (approved):
LGTM, with specific points of discussion below. None of them are important enough to hold up merging this to develop, but I'd like to at least have a conversation about these before deploying this to production.

next-app/src/lib/progress/index.js
```diff
@@ -0,0 +1,21 @@
...
+
+export const stop = id => {
+  const i = pending.findIndex(anId => anId === id)
+
+  if (i >= 0) {
+    pending.splice(i,1)
+
+    loading.set(!!pending.length)
+  }
+}
```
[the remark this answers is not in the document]
megahirt: agreed

next-app/src/routes/password/changed.svelte
megahirt: @rmunn nice job with your Svelte issue repro and getting the attention of the original dev! 💯

.github/workflows/build-and-publish-next.yml
```diff
@@ -0,0 +1,41 @@
...
+
+    defaults:
+      run:
+        working-directory: docker
+
+    steps:
+      -
+        uses: actions/checkout@v2
+      -
+        run: make build-next
```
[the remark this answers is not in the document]
longrunningprocess: yeah, I'm definitely open to this idea... I'd like to simplify things if possible.

.github/workflows/integrate-and-deploy.yml
```diff
@@ -50,15 +59,18 @@ jobs:
...
-        name: Build app
-        run: docker-compose build --build-arg ENVIRONMENT=production --build-arg BUILD_VERSION=${{ steps.image.outputs.TAG }} app
+        name: Build legacy app
+        run: |
+          docker-compose build --build-arg ENVIRONMENT=production --build-arg BUILD_VERSION=${{ steps.image.outputs.TAG }} app
+          docker-compose run --rm app cat /var/www/html/build-version.txt /var/www/html/version.php
       -
-        run:  docker-compose run --rm app cat /var/www/html/build-version.txt /var/www/html/version.php
+        name: Build "next" images
+        run: make build-next
```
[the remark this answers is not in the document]
longrunningprocess: yep, got it

next-app/src/lib/progress/index.js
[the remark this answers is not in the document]
longrunningprocess: yeah, I could see a need for that.   Those are called "determinate" progress indicators and we should be able to extend our current "indeterminate" API and component to accept and utilize something like what you've described.

next-app/src/lib/forms/Input.svelte
longrunningprocess: > I see the benefit here for using type=password yet having just one Input component instead of two. I think for things like checkboxes and radio buttons we'll want components named Checkbox and Radio, and so on, rather than re-using the Input component.

yeah, those might make sense to do for the same reasons we wouldn't include `input type=button` in here but there will be others like `email`, `search` that are similar enough not write a separate comp so it will boil down to a case-by-case basis.  I assume we'll at least do wrappers for the corresponding UI comps from Daisy:
image

> The one thing I'd want to check is whether this can cause a flash of unstyled content

If I'm thinking about the scenario correctly, this would only be a problem if we were to prefill the (password) input with a value which I don't think we ever would.  If there's another scenario I'm not thinking of let me know and we'll REPL it to make sure.

> whether there are any browsers that won't take the type = 'password' assignment after DOM rendering

I'm not familiar with this possibility but if it turns out to be a real concern we can look at mitigating it.

> The alternative, where we create a duplicate component...
yeah, just putting a conditional around the two concrete inputs would be another possible solution.  

We can look at solutions to mitigate your concerns if you find they are still valid.

next-app/src/routes/password/changed.svelte
longrunningprocess: I haven't really put any integrity checks in place on this page to ensure the user came from the correct previous page because if someone is fiddling with url's they can expect some weirdness :-) like a page that says they changed their password when they didn't...so maybe your warning about `getSession` might prove to be relevant for other future pages?

next-app/src/lib/fetch/server.js
longrunningprocess: we may need to discuss this more... the exception block you noted here wouldn't get tripped for a `500` from the api call... we wouldn't know about it until we inspected the response.  If it were a 500 we'd need to take a closer look at the shape of the data in that response.  As far as I could tell the only valuable data from `sf` failures was in the fields I pass along...

I did not shut the db down to see what the error response looks like in that case, maybe it's diff.  If so, we'll need to potentially look it various places for the error info.

next-app/src/lib/forms/Button.svelte
longrunningprocess: > Does the btn-error style override btn-primary?

Yes it does, you can see that demonstrated on the `Index.svelte` page if you just `npm run dev` and hit `http://localhost:3000`

> Also, would we want export let primary in the future, and/or a few other colors-as-props?

Potentially as needed, `secondary` or `accent` perhaps.

next-app/src/lib/fetch/server.js
longrunningprocess: fwiw, shutting the db down locally and trying a change password results in the following:
image

.github/workflows/build-and-publish-next.yml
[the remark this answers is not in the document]
rmunn: Resolving this conversation so that we won't be blocked from merging the PR. If you have more to say, @longrunningprocess, I'd be happy to carry on the conversation(s) in regular comments.

next-app/src/routes/password/changed.svelte
rmunn: Yes, it probably will... but I hope https://github.com/sveltejs/kit/issues/4426 will be resolved before we need to care about that.

As with the other threads, I'm going to resolve this conversation so it doesn't block our ability to merge the PR. That doesn't mean I don't want to continue discussing this, but it's probably better to do so in primary PR comments so that the PR merge isn't blocked. (Or we could consider whether the "All conversations must be resolved" setting is worth the extra hassle of doing this.)

longrunningprocess: I still kind of like the resolve rules we have in place.  I'm adding the discussion items to our team meeting agenda

megahirt:
@longrunningprocess I'm good with you merging this.

PR #1365, opened 2022-03-29
New TestControl API, rename testControl.ts to testSetup

rmunn:
New API entries in testSetup.ts:

* initTestProject to create new, empty project
* addCustomField to add a custom field to a project
* addLexEntry to add a lexical entry to a project
* getProjectJson for help with logging or expect()ing project config
* changePassword API moved from testControl.ts to testSetup.ts

Also added some examples in example.spec.ts showing how to use the
testSetup API. Each example is in a `test.skip` block so that they will
not actually run by default.

### DIFF
--- a/test/e2e/utils/TestControl.php
+++ b/test/e2e/utils/TestControl.php
@@ -2,15 +2,27 @@
 
 namespace Api\Service;
 use Api\Model\Shared\Mapper\MongoStore;
+use Api\Model\Shared\Mapper\ArrayOf;
 use Api\Model\Shared\Command\UserCommands;
 use Api\Model\Shared\ProjectModel;
 use Api\Model\Shared\UserModel;
 use Api\Model\Shared\UserModelWithPassword;
+use Api\Model\Languageforge\Lexicon\LexEntryModel;
 use Api\Model\Languageforge\Lexicon\LexProjectModel;
+use Api\Model\Languageforge\Lexicon\Config\LexConfig;
+use Api\Model\Languageforge\Lexicon\Config\LexConfigOptionList;
+use Api\Model\Languageforge\Lexicon\Config\LexConfigMultiOptionList;
+use Api\Model\Languageforge\Lexicon\Config\LexConfigMultiParagraph;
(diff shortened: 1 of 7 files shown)


test/e2e/utils/TestControl.php
```diff
@@ -99,7 +111,15 @@ public function change_password($username, $password)
         return '';
     }
 
-    public function init_test_project($projectCode = null, $projectName = null, $ownerUsername = null)
+    public function reset_projects()
+    {
+        $db = MongoStore::connect(DATABASE);
+        $db->dropCollection('projects');
+        $db->createCollection('projects');
```
megahirt, line 118: AFAIK `$db->createCollection('projects');` is optional here.  The collection is created on first write.

test/e2e/utils/TestControl.php
```diff
@@ -117,21 +136,119 @@ public function init_test_project($projectCode = null, $projectName = null, $own
...
+
+    public function get_project_json(string $projectCode) {
+        $db = MongoStore::connect(DATABASE);
+        $project = $db->projects->findOne(['projectCode' => $projectCode]);
+        return $project;
+    }
+
+    public function new_method() {
+        return 'hello';
+    }
```
megahirt, line 253: Looks like a method we don't need in this PR :)

megahirt (changes requested):
@rmunn lots of great improvements you're building into the testSetup.  Can you remove the new_method() from this PR?  Thanks!

test/e2e/utils/TestControl.php
rmunn: Sure. Will remove.

test/e2e/utils/TestControl.php
rmunn: I think you're right. I'll remove it too.

rmunn:
> @rmunn lots of great improvements you're building into the testSetup. Can you remove the new_method() from this PR? Thanks!

@megahirt Done.

megahirt (approved):
Great!  Thanks!

PR #1370, opened 2022-03-31
Run E2E tests from correct working directory

### LINKED ISSUE #1369: bug: Integration E2E tests running from wrong directory
**Describe the bug**
The `.github/workflows/integrate-and-deploy.yml` script has a default `working_directory` of "docker" since most of its commands use `make` or `docker-compose` and the Makefile and docker-compose.yml file live in `docker`.

However, the E2E tests need to be run from the root of the repo, since they search for tests from their current location and won't go up to a parent directory while searching. That means that if they're run from `docker` they won't find any tests. And therefore the E2E test script includes `docker-compose -f docker/docker-compose.yml up -d app-for-playwright`, which will fail if run from `docker` since there's no `docker/docker/docker-compose.yml` file.

The solution is to simply add `working_directory: .` to the Playwright test step.

rmunn:
## Description

E2E tests assume they are being run from root of repo, not from `docker` directory, so they need `working_directory: .` in the GHA workflow file.

Fixes #1369.

### Type of Change

### DIFF
--- a/.github/workflows/integrate-and-deploy.yml
+++ b/.github/workflows/integrate-and-deploy.yml
@@ -59,6 +59,7 @@ jobs:
         run: make unit-tests-ci
       -
         name: Playwright E2E Tests
+        working-directory: .
         run: |
           docker-compose -f docker/docker-compose.yml up -d app-for-playwright
           npx playwright install


megahirt (approved):
Great thanks, @rmunn  This should fix the issue that @longrunningprocess identified with E2E tests not running on integrate-and-deploy

longrunningprocess:
Thanks for taking a look at this and I'll pull this into my branch on my next rebase.  I'd like to note however, had this followed convention and put those commands behind a make target like `make e2e` or something, we could've avoided this.  Most likely in the make target there would've been a `cd ..` and then all the necessary commands needed to make it work.

longrunningprocess:
actually looking at the `Makefile`, I see there are still targets in there... maybe we should've tweaked those?

megahirt:
Yes, I still want to keep the old ones in `make e2e-tests` as I can still run those on my machine locally.  Yes, we should have a make target, probably make `playwright-tests`.  Or we could rename the old ones to `make protractor-tests` and the new ones be `make e2e-tests`.

Yes, it would be better to use Make in the build script.  There's a number of places where we aren't doing that since there are a few small differences - but I think those could be ironed out.

My goal after the lfnext branch is merged, is to move the Makefile to the repo root.  No more `cd docker` for Make commands.

longrunningprocess:
ok, that sounds good

rmunn:
> Thanks for taking a look at this and I'll pull this into my branch on my next rebase. I'd like to note however, had this followed convention and put those commands behind a make target like `make e2e` or something, we could've avoided this. Most likely in the make target there would've been a `cd ..` and then all the necessary commands needed to make it work.

Unfortunately, the way Playwright's `webServer` config works is that it registers the process ID of the command that runs, and kills that process ID when it's done — which means that I couldn't use `make e2e-tests` (I tried that first). I'm not always good at concisely explaining all the technical details behind my decisions, which means that sometimes I decide not to write an essay and to just do things; sorry about that.

longrunningprocess:
gotcha, thanks for at least trying!

PR #1376, opened 2022-04-04
Fix saving custom fields on senses and examples

### LINKED ISSUE #1375: Custom fields not being saved on senses or examples
**Describe the bug**
Sense-level and example-level custom fields do not appear to be being saved; only entry-level custom fields are being saved.

**To Reproduce**
1. Create a new non-S/R project
2. Load the `TestLangProj.7z` data from `test/common`
3. Edit the CustMultiListRef field in Meaning 1 of any entry
4. Click away from the entry and click back to it
5. Note the custom field data did not get saved

**Expected behavior**
The custom field data is saved.

rmunn:
## Description

We were treating every field as an object, but the senses and example fields are actually arrays of objects. This was resulting in custom fields not being sent to the server in the right format, so the server was simply not seeing any custom field changes in senses or examples: only custom fields at the entry level were being saved.

Fixes #1375.

### Type of Change

## Testing on your branch

See #1375's repro steps.

## Checklist

### DIFF
--- a/src/angular-app/languageforge/lexicon/editor/editor.component.ts
+++ b/src/angular-app/languageforge/lexicon/editor/editor.component.ts
@@ -1157,6 +1157,9 @@ export class LexiconEditorController implements angular.IController {
   }
 
   private prepCustomFieldsForUpdate(data: any): any {
+    if (Array.isArray(data)) {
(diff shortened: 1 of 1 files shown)


megahirt (approved):
I have followed @rmunn 's repo steps on his branch locally and can verify that this fixes issue
#1375

PR #1378, opened 2022-04-08
Run lfmerge in separate container

rmunn:
This replaces #1315 with a deployment setup that uses a .json file, named `other-dependencies.json`, at the root of the repository to keep track of the lfmerge version we want to use. This can be overridden by using environment variables, e.g. for manual deployments. But most of the time when a new lfmerge version is desired, we'll bump the version number recorded in other-dependencies.json.

### Type of Change

## Testing on your branch

- Run `make` as usual. This should pull down the version of the lfmerge Docker image defined in the `other-dependencies.json` file.
- Do a Send/Receive of a test project from Language Depot, if you have one, to verify that lfmerge still works correctly.

## Checklist


docker/Makefile
```diff
@@ -35,7 +35,8 @@ unit-tests-ci:
 .PHONY: build
 build:
 	npm install
-	docker-compose build mail app ld-api next-proxy next-app
+    # TODO: When this Makefile is moved to repo root, turn "../" below into "./"
+	LFMERGE_VERSION="$(shell jq -r .lfmerge ../other-dependencies.json)" docker-compose build mail app ld-api next-proxy next-app
```
longrunningprocess, line 39: `shell` nor `jq` can be assumed to be present on the dev's machine unless we add instructions to the README about additional local installs which I'd prefer not to do.  Let's discuss this approach further.

docs/RELEASE.md
```diff
@@ -25,7 +25,7 @@ Current workflow:
 
 [languageforge.org](https://languageforge.org)
 
-Production deployments can be manually run with `VERSION=<some-docker-tag-or-semver> make deploy-prod`.
+Production deployments can be manually run with `VERSION=<some-docker-tag-or-semver> VERSION_LFMERGE=<some-docker-tag-or-semver> make deploy-prod`.
```
longrunningprocess, line 28:
```suggestion
Production deployments can be manually run with `VERSION_APP=<some-docker-tag-or-semver> VERSION_PROXY=<some-docker-tag-or-semver> VERSION_NEXT_APP=<some-docker-tag-or-semver> VERSION_LFMERGE=<some-docker-tag-or-semver> make deploy-prod`.
```

longrunningprocess: apparently I missed that in my previous work, sorry about that.

docker/Makefile
longrunningprocess, line 39: did you happen to check whether doing this type of thing in the `docker-compose.yml` is possible?  I'm thinking something like `ghcr.io/sillsdev/lfmerge:${jq ...}` or 
```yml
environment:
    - LFMERGE_VERSION=`jq ...`
```

I'm not certain off the top of my head but would be better if possible.

.github/workflows/integrate-and-deploy.yml
```diff
@@ -46,11 +49,19 @@ jobs:
...
+          TAG_LFMERGE:
         run: |
           echo ::set-output name=NAMESPACE::sillsdev/web-languageforge
           echo ::set-output name=TAG_APP::${{ inputs.image-tag-app }}
           echo ::set-output name=TAG_PROXY::${{ inputs.image-tag-proxy }}
           echo ::set-output name=TAG_NEXT_APP::${{ inputs.image-tag-next-app }}
+          echo ::set-output name=LFMERGE_NAMESPACE::ghcr.io/sillsdev/lfmerge
+          # LfMerge tag from parameter if provided, use .json file if exists, fallback to "latest" as default
+          TAG_LFMERGE=${{ inputs.image-tag-lfmerge }}
+          TAG_LFMERGE=${TAG_LFMERGE:-$(jq -r .lfmerge other-dependencies.json)}
```
longrunningprocess, line 62: it probably is but we need to verify `jq` is present in GHA's runners

docker/docker-compose.yml
```diff
@@ -76,6 +76,37 @@ services:
...
+      - LFMERGE_BASE_DIR=/var/lib/languageforge/lexicon/sendreceive
+      - LFMERGE_WEBWORK_DIR=webwork
+      - LFMERGE_TEMPLATES_DIR=Templates
+      - LFMERGE_MONGO_HOSTNAME=db
+      - LFMERGE_MONGO_PORT=27017
+      - LFMERGE_MONGO_MAIN_DB_NAME=scriptureforge
+      - LFMERGE_MONGO_DB_NAME_PREFIX=sf_
+      - LFMERGE_VERBOSE_PROGRESS=true
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
```
longrunningprocess, line 102: remind me why this is required?

.github/workflows/integrate-and-deploy.yml
```diff
@@ -14,6 +14,9 @@ on:
       image-tag-next-app:
         type: string
         required: true
+      image-tag-lfmerge:
```
longrunningprocess, line 17: the lfmerge version is really driven from the `.json`, I'm not sure we actually need this at all.  Was there a use-case you were imagining where we'd tweak the `staging.yml` or `production.yml` to feed this workflow a version?

longrunningprocess (changes requested):
yep, this is pretty close to what I would've done, nice work.

.github/workflows/integrate-and-deploy.yml
rmunn: I think I was picturing a use case where we'd add a workflow_dispatch trigger to `integrate-and-deploy.yml`, in which case it would be useful to have the ability to override the .json values. Say, because we've discovered that I introduced a bug in the current version of LfMerge and want to quickly roll back to the previous version.

You're right that that's not needed most of the time, but I thought having an override available might be a good idea.

.github/workflows/integrate-and-deploy.yml
rmunn: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md lists `jq` version 1.6, and the 18.04 runner lists `jq` version 1.5. So we're good.

docker/Makefile
rmunn: I checked, and you can't use shell commands inside the docker-compose.yml file. I agree that it would have been simpler — it's the first thing that I tried — but no, if you want things that will change you have to pass them in as environment variables. (Or pass the docker-compose file through sed or something, but the documentation recommends env vars).

rmunn: `shell` is a Makefile keyword, but you're right about `jq` needing to be added to the README.

docker/docker-compose.yml
rmunn: I don't remember right now. It had something to do with being able to access `localhost` from inside the container and have it redirected to the Mongo database mounted on a port of the host machine, but I'll have to look up why I needed that when I get back to the office.

docker/Makefile
megahirt: I was thinking about this too.  We have several common *nix tools that we are using that have snuck in.  sed is also on the list and it's been in our scripts for a while.  

I think while it would have been nice to have a pure Windows dev on the team, both Hanna and Joseph were keen to use WSL Ubuntu and so we actually only support Linux and MacOS dev environments at this point.

Our small list of dev dependencies is now:
Docker
make
sed
bash
jq

Anything else?

docker/docker-compose.yml
megahirt: Correct on the localhost thing.  I don't think it's for mongo though.  We have it commented out in the LF app and only use it for xdebug (which needs to access localhost)

megahirt: Maybe you can remove it and it will still work?

docker/Makefile
longrunningprocess: I had another thought that would eliminate the need for `jq`, I just don't want to keep adding install requirements...it just creates more maintenance issues.

What if we did something like this.

`docker/lfmerge/Dockerfile`:
```
FROM ghcr.io/sillsdev/lfmerge:2.0.98
```

That will eliminate the `.json` and the need for additional installs.  Grabbing the version now can use `head` and `cut`.  The `docker-compose.yml` would just need a reference to the `Dockerfile` like many of our other containers.

longrunningprocess: I proposed a different idea that would eliminate some of these work-arounds.

.github/workflows/integrate-and-deploy.yml
longrunningprocess: see my other comments on this topic.

.github/workflows/integrate-and-deploy.yml
longrunningprocess: let's wait until that need comes up.

docker/docker-compose.yml
longrunningprocess: Let's try to remember why and if it is still needed, let's capture that "why" in a comment.

docker/Makefile
longrunningprocess: FYI, `head` and `cut` are both available in our GHA runners: https://github.com/longrunningprocess/gha-sandbox/runs/6064688705?check_suite_focus=true

docker/docker-compose.yml
rmunn: It's used in xdebug (see docker/app/docker-php-ext-xdebug.ini), so it's needed in the app and app-for-playwright containers. But it's not needed in the lfmerge container. Will remove.

.github/workflows/integrate-and-deploy.yml
rmunn: Done in commit d7b3e50.

docker/docker-compose.yml
rmunn: Done in commit 6e74b41.

docker/Makefile
rmunn: Done in commits 02b2456 and e508b93.

rmunn:
@longrunningprocess Finished addressing all your review comments. There's one merge conflict remaining, which shouldn't be a merge conflict (check it out in the GitHub UI and you'll see that it should be a trivial resolution), but apparently my edits have managed to confuse Git. So this might need resolving in the GitHub merge-conflict resolution UI, which will create a merge commit merging develop into this branch, the very thing I was trying to avoid. Oh well, I think it'll go away when we squash so it won't be a big deal.

longrunningprocess (approved):

.github/workflows/integrate-and-deploy.yml
```diff
@@ -46,11 +46,18 @@ jobs:
...
+        env:
+          TAG_LFMERGE:
         run: |
           echo ::set-output name=NAMESPACE::sillsdev/web-languageforge
           echo ::set-output name=TAG_APP::${{ inputs.image-tag-app }}
           echo ::set-output name=TAG_PROXY::${{ inputs.image-tag-proxy }}
           echo ::set-output name=TAG_NEXT_APP::${{ inputs.image-tag-next-app }}
+          echo ::set-output name=LFMERGE_NAMESPACE::ghcr.io/sillsdev/lfmerge
+          # LfMerge tag from parameter if provided, use .json file if exists, fallback to "latest" as default
+          TAG_LFMERGE=$(head -n 1 docker/lfmerge/Dockerfile | cut -d: -f2)
```
longrunningprocess, line 58: `head -1` also works that same if you like the shorter syntax

.github/workflows/integrate-and-deploy.yml
```diff
@@ -46,11 +46,18 @@ jobs:
       -
         name: Establish image name
         id: image
+        env:
```
longrunningprocess, line 49: artifact from a previous age, can be removed

.github/workflows/integrate-and-deploy.yml
rmunn: I do know that, but I've always preferred the `-n num` syntax. For one thing, if you want to specify a negative number then you must use the `-n` syntax (`head -n -1` displays all but the last line of the input, while `head --1` is a syntax error). And finally, I just checked and the `-#` syntax (where # is a number, of course) is actually *not documented* in the `head` manpage! It's a GNU tool, and they like their non-standard `info` tool, so I had to check the `info head` documentation, where I found the following:

> For compatibility ‘head’ also supports an obsolete option syntax ‘-[NUM][bkm][cqv]’, which is recognized only if it is specified first. NUM is a decimal number optionally followed by a size letter (‘b’, ‘k’, ‘m’) ...

"An obsolete option syntax" doesn't give me confidence that it will continue to be supported going forward. So I've gotten used to using the `-n num` syntax instead.

rmunn: BTW, commit 8bd4169 fixes the comment, which still had a reference to the input parameter we removed earlier.

.github/workflows/integrate-and-deploy.yml
rmunn: Done in commit 8bd4169.

.github/workflows/integrate-and-deploy.yml
longrunningprocess: :-D you're too funny!  You had me at the compat reasoning which I can appreciate but a `-1` to get the last line over `tail`?!  That's sheer madness 🤪

PR #1385, opened 2022-04-18
Add .editorconfig entry for Makefiles

rmunn:
This will force Tab characters in Makefiles, reducing the amount of time we have to spend fussing with whitespace when we edit them.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,6 +8,11 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
+; applies only to Makefiles
+[Makefile]
+indent_style = tab
+tab_width = 4
+
 ; applies only to JS, TS, and YML files
 [*.{js,ts,yml}]
 tab_width = 2


megahirt (approved):
Looks great to me!

longrunningprocess:
yes, good idea!  I forget some people have tweaked their local environments to replace tabs with spaces...thanks!

rmunn:
> yes, good idea! I forget some people have tweaked their local environments to replace tabs with spaces...thanks!

Heh. I have a rant about how tab characters *almost* work for their intended purpose, but our editors aren't smart enough to deal with them, so space characters are better in almost all situations. I'll bring it up sometime in standup meetings if there's a couple minutes to kill. Basically, it goes like this:

```text
class Foo {
>---function bar(someParam,
>---.............otherParam) {
>--->---return false;
}
```

That's how parameter alignment *should* be done (if you're going to do it at all). Tabs for indentation, spaces for alignment.

```text
class Foo {
>---function bar(someParam,
>--->--->--->---.otherParam) {
>--->---return false;
}
```

That's how parameter alignment usually ends up after you save your file, because your editor "helpfully" replaced those spaces with tabs for you.

```text
class Foo {
>-function bar(someParam,
>-.............otherParam) {
>->-return false;
}
```

That's what the function *should* look like when someone else, who sets his tab width to 2 spaces instead of 4, loads it. Because you did it correctly, the function parameters are still aligned.

```text
class Foo {
>-function bar(someParam,
>->->->-.otherParam) {
>->-return false;
}
```

That's what the function will *actually* look like, because your editor was "helpful". Now the function parameters aren't aligned any more.

```text
class Foo {
>-function bar(someParam,
>->->->->->->-.otherParam) {
>->-return false;
}
```

That's what your colleague will do when he tries to line up the parameters that were clearly supposed to be aligned. (Or he'll do it right with tabs for indentation and spaces for alignment, but his editor will be "helpful" and produce that result).

```text
class Foo {
>---function bar(someParam,
>--->--->--->--->--->--->---.otherParam) {
>--->---return false;
}
```

That's what you will see when your colleague passes the file back to you.

PR #1400, opened 2022-05-05
Chore/user management tests

JeanneSon:
## Description
**These tests do not need to be converted since the user management app is not
used any more and got replaced by the share button.**

The user management tests were converted from Protractor to Playwright.
Some problems were encountered: The manager of the problem does not see
the other users in the user management, nor does he/she see him-/herself.

### Type of Change
E2E tests

## Checklist


rmunn:
As discussed with @JeanneSon last week, we'll be closing this PR without merging it since the usermanagement app has been replaced by the share button.

PR #1389, opened 2022-04-23
Remove lfmerge from LF app container

rmunn:
## Description

Now that lfmerge is running in a separate container, we can remove the lfmerge package (and all its dependencies) from the LF app container, reducing the size of the app container.

This is a continuation of #1315.

### Type of Change

Only keep lines below that describe this change, then delete the rest.

- New feature (non-breaking change which adds functionality)

## Testing on your branch

Please describe how to test and/or verify your changes. Provide instructions so we can reproduce.  Please also provide relevant test data as necessary.  These instructions will be used for QA testing below.

- [ ] Load Language Forge and create a new project from Language Depot
- [ ] Edit the project in Language Forge (add an entry, modify a couple of entries) and do another Send/Receive
- [ ] Send/Receive that same project in FieldWorks to ensure that the changes really did come through

## Checklist

- [X] I have performed a self-review of my own code
- [X] I have reviewed the title/description of this PR which will be used as the squashed PR commit message
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added tests that prove my fix is effective or that my feature works

## qa.languageforge.org testing

Reviewers: add/replace your name below and check the box to sign-off/attest the feature works as expected on qa.languageforge.org

- [ ] Reviewer1 (YYYY-MM-DD HH:MM)
- [ ] Reviewer2 (YYYY-MM-DD HH:MM)

PR #1407, opened 2022-05-23
Move lfmerge container to separate pod using pod affinity

### LINKED ISSUE #1406: feat: Move LfMerge container to separate pod
In #1378, we moved the lfmerge files to a separate container within the main LF app pod. Now we want to move them to a completely separate pod, so that each pod is running only one container. At the moment both pods still need to be assigned to the same node so that we can use ReadWriteOnce volumes instead of needing ReadWriteMany volumes, but at some point Longhorn (the storage backend we're using) [should be able to manage ReadWriteMany volumes](https://longhorn.io/blog/longhorn-v1.1.0/) so we'll be able to remove that pod affinity requirement at some point.

An lfmerge-deployment.yaml file that uses pod affinity to ensure the lfmerge pod and app pods always run on the same node.

Leaving it as-is, with lfmerge being a second container in the app pod.

rmunn:
## Description

As a follow-up to #1389, this PR moves the lfmerge container into a separate pod, using pod affinity to ensure that it will be assigned to the same node as the main app. (Since they share a persistent volume, this is necessary so that the volume can be mounted in ReadWriteOnce mode instead of ReadWriteMany — ReadWriteMany greatly limits the number of storage backends available, while ReadWriteOnce is a lot easier for storage backends to use. (I see https://longhorn.io/blog/longhorn-v1.1.0/ from January 2021 that announces ReadWriteMany availability in Longhorn, so perhaps the pod affinity can be gotten rid of at some point).

Fixes #1406.

### Type of Change

## Screenshots

## Testing on your branch

- [ ] Ensure LF app still builds with `make`
- [ ] Clone a project with Send/Receive and make sure LfMerge still runs

This PR makes basically no changes to the docker-compose build (one very minor change), so I don't expect it to have any effect. Most of the testing will need to be on the QA server.

## Checklist

### DIFF
--- a/docker/deployment/Makefile
+++ b/docker/deployment/Makefile
@@ -10,9 +10,11 @@ list-contexts:
 list-config:
 	cat ~/.kube/config
 
-logs: logs-app logs-db logs-mail logs-next-app logs-next-proxy
+logs: logs-app logs-lfmerge logs-db logs-mail logs-next-app logs-next-proxy
 logs-app:
 	kubectl logs deploy/app
+logs-lfmerge:
+	kubectl logs deploy/lfmerge
 logs-next-app:
 	kubectl logs deploy/next-app
 logs-next-proxy:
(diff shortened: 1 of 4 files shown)


longrunningprocess (approved):
well, I like the way this looks a lot more, let's hope it works!  🤞🏻

PR #1409, opened 2022-05-25
Add lfmerge deployment to GHA workflow

rmunn:
## Description

Now that the lfmerge container is in a separate pod, we also need to tweak the GHA workflow to update the proper Kubernetes deployment when the workflow runs. Instead of setting the `lfmerge` image in the `app` deployment, the `lfmerge` deployment is now the one that will contain the `lfmerge` image.

This really should have been included in #1407, but I forgot this one step.

### Type of Change

## Testing on your branch

- [ ] Clone a project from Language Depot
- [ ] Verify that the clone completed and the data is there

Cloning just one project should be enough: if there are any problems with this deployment, they will cause lfmerge not to run at all, rather than having it run but fail half-way through.

## Checklist

### DIFF
--- a/.github/workflows/integrate-and-deploy.yml
+++ b/.github/workflows/integrate-and-deploy.yml
@@ -115,6 +115,7 @@ jobs:
         uses: sillsdev/common-github-actions/install-kubectl@v1
       -
         run: |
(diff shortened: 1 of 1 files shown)


longrunningprocess (approved):
yeah, wish I had thought of it in the other as well, sorry about that.

rmunn:
Merging this before https://github.com/sillsdev/web-languageforge/pull/1413 goes in, so that we can watch it happening on QA.

PR #1419, opened 2022-06-06
Match emails case-insensitively

### LINKED ISSUE #1418: bug: login email address field is case sensitive
**Describe the bug**
When users try to sign in at https://languageforge.org/auth/login, if they use any capital letters in their email address, they will be presented with the message "Invalid username or password".

This is a very vexing, high-priority issue that I have run into in the past, and which has kept a key stakeholder from signing into LF for over six months. 

Users generally won't think email case is an issue, and will assume there, is a problem with the password, resulting in many fruitless password reset attempts. This would also be problematic for those who do pay attention to email case. If I'm not mistaken, if they use capital letters during account registration, LF silently converts them and stores the email address with all lowercase letters. (This converting is expected behaviour, but may cause this bug to express itself more often.)

**To Reproduce**
1. Pick a Language Forge account for which you sign in using an email address and password (not sign in with Facebook or Google).
2. Type in your email address, but using at least one capital letter.
3. Type in the correct password
4.  Click "Login"
5. Be presented with the "Invalid username or password." message

**Expected behavior**
Language forge ignores case in the email address field. 

**Screenshots**
(none)

 - OS: Windows 10
 - Browser: Chrome

**Ideas**
* Quick fix: put "Write email address with all lower-case letters." above the email address field (and remove this line after the proper fix)
* Give keyboard hint to not initial-cap text input in the email address field.
* (And of course do the proper fix whereby LF ignotes case 😀)

rmunn:
## Description

Fixes #1418.

We try a case-sensitive match first in case it matters (there could be some email domains that treat "xyz@example.com" and "XYZ@example.com" as different addresses, and we want to be able to retrieve the correct one in the rare case where that could matter), but we then fall back to case-insensitive matching on email addresses for the vast majority of cases where that's the right thing.

This is a first pass at solving the issue. I haven't tested the code yet, so this is a draft PR instead of a full PR. Unit tests would be needed to make sure that the `readByPropertyInsensitive` function I just created will do the right thing.

### Type of Change

## Screenshots

## Testing on your branch

TODO: Write tests, then write testing instructions.

- [ ] Test A
- [ ] Test B

## Checklist

### DIFF
--- a/src/Api/Model/Shared/Mapper/MapperModel.php
+++ b/src/Api/Model/Shared/Mapper/MapperModel.php
@@ -69,6 +69,16 @@ class MapperModel extends ObjectForEncoding
     }
 
     /**
+     * @param string $property
+     * @param string $value
+     * @return boolean
+     */
+    public function readByPropertyInsensitive($property, $value)
+    {
+        return $this->_mapper->readByPropertyInsensitive($this, $property, $value);
+    }
+
+    /**
      * @param array $properties
      * @return boolean
      */
--- a/src/Api/Model/Shared/Mapper/MongoMapper.php
+++ b/src/Api/Model/Shared/Mapper/MongoMapper.php
@@ -261,6 +261,47 @@ class MongoMapper
 
     /**
      * @param mixed $model
+     * @param string $property
+     * @param string $value
+     * @return bool true on document found, false otherwise
+     * Note that unlike the read() method, readByProperty() does NOT throw an exception if no document is found
+     *
+     */
+    public function readByPropertyInsensitive($model, $property, $value)
+    {
+        CodeGuard::checkTypeAndThrow($value, 'string');
+        $escaped = \preg_quote($value);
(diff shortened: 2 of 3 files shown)


megahirt:
I tested locally and I can confirm that this PR does address the issue.  I suggest that an appropriate fix can be done simpler, e.g. using `strtolower()` instead of adding an additional case insensitive property match.

On our production mongo instance I confirmed that no emails contain uppercase characters.  This is because the email address is [converted to lowercase on registration](https://github.com/sillsdev/web-languageforge/blob/develop/src/Api/Model/Shared/Command/UserCommands.php#L362).

Here is the output of my research on production Mongo:

```
> db.users.find().count();
2439
> db.users.find({'email': /[A-Z]/}).count();
0
> db.users.find({'email': /[a-z]/}).count();
2319
> db.users.find({'email': null}).count();
120
>
```

I propose that we close this PR in favor of the simpler strtolower() fix in a forthcoming PR.

rmunn:
Closing in favor of #1420.

PR #1420, opened 2022-06-06
Convert to lowercase before email lookup

### LINKED ISSUE #1418: bug: login email address field is case sensitive

megahirt:
## Description

This fixes an issue where a user could not login because they are accustomed to case-sensitive email addresses.  This is a simpler implementation of #1419 , given that we know that all our email addresses are already lowercase.

Fixes #1418 

### Type of Change

## Testing on your branch

- [ ] Locally on develop, sign-up for an account and then login using your email address (lowercase).  Verify that it works.  Now, logout and login with part of the email typed with capital letters e.g. `Me@example.com` You will get an invalid username/password error.

- [ ] Now on this branch, repeat the testing and verify that capital letters in the email address login field do not make a difference.  You should be able to login using e.g. `ME@example.com` just fine. 

## Checklist

### DIFF
--- a/src/Api/Model/Shared/UserModel.php
+++ b/src/Api/Model/Shared/UserModel.php
@@ -313,7 +313,7 @@ class UserModel extends MapperModel
      */
     public function readByEmail($email)
     {
-        return $this->readByProperty('email', $email);
+        return $this->readByProperty('email', strtolower($email));
     }
 
     /**


megahirt:
My thanks to @rmunn for his PR #1419 which inspired this fix.  I had previously attempted a fix using `strtolower()` in [Auth.php](https://github.com/sillsdev/web-languageforge/blob/develop/src/Site/Controller/Auth.php#L177) but hadn't gotten a working solution yet.

longrunningprocess (approved):
Doesn't get much simpler, nice one!

rmunn (approved):
Agreed, this is better than doing a case-insensitive search with Mongo, because in Mongo that requires using the `$regex` matcher and emails can have all kinds of characters that have special meanings in regex, such as the widely-used `robin.munn+sometag@gmail.com` feature. So my solution would have required extensive testing of the regex quoting, whereas if we know that we are already storing all email addresses in lowercased form, then this is better.

PR #1416, opened 2022-06-02
Migrate project activity to LFNext

longrunningprocess:
## Description

This PR seeks to migrate the project activity capability to LFNext.  While at it, discussion arose about a "Project landing page".  That is also included in this work.  Refer to #1394 for further discussion

Fixes #1393 

### Type of Change

Only keep lines below that describe this change, then delete the rest.

- New feature (non-breaking change which adds functionality)
- UI change

## Screenshots

This video runs through the look of it, mobile views and all of the links back into this page.

https://user-images.githubusercontent.com/4412848/171720727-00b36ca5-b32a-4598-9b54-363a0bd9b540.mov

forgot to include some of the h-scroll behavior of the activity table on mobile, that demo is here:

https://user-images.githubusercontent.com/4412848/171729717-8faec45a-7064-40fc-8b1d-15122efb5c15.mov

## Testing on your branch

- [ ] Start a new project without any activity and ensure the activity table reflects that.
- [ ] With your new project, add a word and some activity to ensure the most recent data shows up on the activity table.
- [ ] Pick a project with known activity and ensure only the last 30 days is initially shown
- [ ] When viewing a project with activity, ensure the data is sorted by date, then by user within the same date
- [ ] When viewing a project with activity, ensure clicking "show all" retrieves all activity
- [ ] Ensure the project link under the "My Projects" dropdown in the header takes you to the correct project landing page
- [ ] Ensure while on the project landing page, when you click "work on this project" you are taken to the editor view for the correct project
- [ ] Ensure the project links from under the "My Projects" page (`/app/projects`) takes you to the correct project landing page
- [ ] Ensure the project link in the breadcrumb takes you to the correct project landing page
- [ ] Ensure you cannot see the activity for a project when you are not logged in
- [ ] Ensure you cannot see the activity for a project you are not allowed to view
- [ ] Create a link for sharing the project and ensure when the link is used it redirects to the correct project's landing page
- [ ] Create a project via upload and check the activity pages for it

## Checklist

- [x] I have performed a self-review of my own code
- [x] I have reviewed the title/description of this PR which will be used as the squashed PR commit message
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added tests that prove my fix is effective or that my feature works

## qa.languageforge.org testing

Reviewers: add/replace your name below and check the box to sign-off/attest the feature works as expected on qa.languageforge.org

- [ ] Reviewer1 (YYYY-MM-DD HH:MM)
[...]

PR #1430, opened 2022-07-06
Fix TypeScript errors in next-app

rmunn:
This PR fixes a few TypeScript errors I found while reviewing #1416. No logic changes, just a few type annotations where TypeScript couldn't figure them out by itself.


next-app/src/lib/progress/index.ts
```diff
@@ -2,7 +2,7 @@ import { writable } from 'svelte/store'
 
 export const loading = writable(false)
 
-const pending = []
+const pending: any[] = []
```
longrunningprocess, line 5: I don't see the point of `any`... I'd more inclined to force a `string` if you feel something is needed here.

next-app/src/routes/projects/[project_code]/activities.ts
```diff
@@ -3,7 +3,7 @@ import { sf } from '$lib/fetch/server'
 export async function get({ params: { project_code }, request: { headers } }) {
 	const cookie = headers.get('cookie')
 
-	const activities = await get_activities({ project_code, cookie })
+	const activities = await get_activities({ project_code, cookie, start_date: undefined, end_date: undefined })
```
Comment (longrunningprocess) [inline next-app/src/routes/projects/[project_code]/activities.ts:6]: I'd like to understand this one better, seems unnecessary to me but I'd like to hear your thoughts.

longrunningprocess (changes requested):

next-app/src/routes/password/index.ts
```diff
@@ -30,7 +30,7 @@ export async function put({ request }) {
 		})
 	} catch (error) {
 		return {
-			status: error.code,
+			status: (error as Error).code,
```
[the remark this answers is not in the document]
rmunn: I tried that first, but Typescript says that a catch expression can only be typed as `any` or `unknown`, so I had to typecast it later on for Typescript to accept it.

next-app/src/routes/projects/[project_code]/activities.ts
Comment (rmunn) [reply next-app/src/routes/projects/[project_code]/activities.ts:6]: Typescript says that the type `{ project_code, cookie }` isn't compatible with the type `{ project_code, cookie, start_date, end_date }` and flags this as a type error. I could have resolved this another way, such as making a type declaration that had `start_date` and `end_date` as optional, but that seemed overkill for a single usage.

next-app/src/lib/progress/index.ts
rmunn: Sure, it could be `string[]`. I used `any[]` because I didn't know what types would be used here and this was the quickest way to get Typescript to stop complaining. But `string[]` probably does make more sense.

next-app/src/lib/fetch/client.ts
```diff
@@ -39,7 +39,7 @@ async function customFetch(method, url, body) {
 		//	the host is refusing connections
 		//	client is offline, i.e., airplane mode or something
 		//	CORS preflight failures
-		throwError(e)
+		throwError((e as Response).statusText, (e as Response).status)
```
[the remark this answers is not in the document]
rmunn: Ah right, when `fetch` throws it's not going to be throwing a Response object. Thing is, I don't think it throws strings either, so unless I'm mistaken about that it's not the right thin to pass the thrown object  into `throwError`. We'll discuss it.

megahirt:
Thanks @rmunn for this PR - I appreciate it.  Out of curiosity, I'm wondering how @longrunningprocess hasn't encountered these errors before on his branch, at least that's my assumption.  I'm wondering if perhaps Typescript is not configured the same between the two of you, and if that's the case, how can we ensure that you're both running the same TS config so that you see the same errors?

TODO: The PR needs to be rebased to resolve the conflicts.

megahirt:
@longrunningprocess other than rebasing (wish I could do that on my current machine but I cannot ATM), is there any reason why you cannot accept these suggestions from @rmunn and then get this into your branch?

longrunningprocess:
> @longrunningprocess other than rebasing (wish I could do that on my current machine but I cannot ATM), is there any reason why you cannot accept these suggestions from @rmunn and then get this into your branch?

just the unresolved conversations... would you like to talk through these instead of waiting on Robin?

megahirt:
Robin is pretty busy so we can resolve for him and rebase for him.

longrunningprocess:
I'm going to close this one since it's so far behind, still unresolved and Robin is gone.  We can go through all of LFNext code with TS in mind as part of another effort.

megahirt:
That sounds fair to me - thanks.

PR #1433, opened 2022-08-08
k8s config updated to ensure isolated node for lfmerge

longrunningprocess:
## Description

lfmerge utilizes a bunch of RAM at times therefore this PR will ensure lfmerge and the app are isolated when deployed to a node.  This will make it less likely for lfmerge to run out of memory on a node and it will also protect other containers from a node being brought down by memory pressure from lfmerge.

### Type of Change

- k8s infrastructure change

## TODO for rmunn

- [x] Remove PHP dependency in line sf.php 657 (see slack message)
- [x] Remove startLfMergeIfNeeded() method

## Screenshots

N/A

## Testing on your branch

- [ ] A general smoke test with a S/R will suffice

## Checklist


rmunn (approved):
LGTM. Using node "taints" is definitely the right approach here, as it will require no changes to The Combine (or any future projects) to make them actively avoid the node dedicated to lfmerge. Nice work.

rmunn:
I've edited `Sf.php` on staging to remove the calls to `startLfMergeIfNeeded` in Sf.php, and lfmerge started up just fine when I did a Send/Receive (and it ran to completion). I did *not* do all the code deletion from this PR on staging, the only code I deleted from staging was the two lines in Sf.php that call the `startLfMergeIfNeeded` function.

However, I did run `make` locally after building this PR with the commit that I removes a bunch of code. And again, lfmerge started up just fine. So all the code I deleted was all safe to delete, as far as not breaking Send/Receive.

I have not *yet* built this PR into an image to deploy to staging, so it's remotely possible that I missed something that would make it work locally but fail when the image is deployed to k8s. But that's highly unlikely. I believe this is safe to merge.

megahirt (approved):

megahirt:
> I have not _yet_ built this PR into an image to deploy to staging, so it's remotely possible that I missed something that would make it work locally but fail when the image is deployed to k8s. But that's highly unlikely. I believe this is safe to merge.

This is an appropriate usage of staging, to test and entire working system that cannot be easily tested on a local machine.

PR #1434, opened 2022-08-10
had to ensure app and lfmerge run on same node

longrunningprocess:
## Description

We learned that lfmerge's inotify will only receive messages from the kernel when the kernel is the same kernel used to write the files into place, i.e., lf-app  in this case.  Therefore, app and lfmerge must run on the same node until another mechanism can be considered that will work in a remote nfs environment.

### Type of Change

- Bug fix (non-breaking change which fixes an issue) (introduced in #1433 )

## Screenshots

N/A

## Testing on your branch

- [ ] A simple S/R smoke test will suffice

## Checklist

### DIFF
--- a/docker/deployment/lfmerge-deployment.yaml
+++ b/docker/deployment/lfmerge-deployment.yaml
@@ -76,7 +76,7 @@ spec:
                 values:
                 - lfmerge
             weight: 1
-        # need to keep this on the same pod as the app since inotify only gets notification wwhen updates occur on the same kernal
+        # need to keep this on the same pod as the app since inotify only gets notification when updates occur on the same kernel
         podAffinity:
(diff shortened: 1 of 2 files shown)


rmunn (approved):
LGTM.

docker/deployment/lfmerge-deployment.yaml
```diff
@@ -76,6 +76,16 @@ spec:
                 values:
                 - lfmerge
             weight: 1
+        # need to keep this on the same pod as the app since inotify only gets notification wwhen updates occur on the same kernal
```
megahirt, line 79:
```suggestion
        # need to keep this on the same pod as the app since inotify only gets notification when updates occur on the same kernel
```

megahirt (approved):
LGTM

PR #1449, opened 2022-09-06
Establish resource limits of containers in k8s

longrunningprocess:
## Description

There were times when lfmerge would utilize so much memory it would cause issues on the node where it was running.  This PR will help establish some limits to resource utilization so any overuse of resources will only affect lfmerge.

### Type of Change

- k8s infrastructure optmization

## Screenshots

N/A

## Testing on your branch

- [ ] Some simple smoke tests will ensure the environment is functioning

## Checklist

### DIFF
--- a/docker/deployment/app-deployment.yaml
+++ b/docker/deployment/app-deployment.yaml
@@ -65,23 +65,6 @@ spec:
       labels:
         app: app
     spec:
-      affinity:
-        # required to ensure this container makes it to lfmerge's dedicated node
-        nodeAffinity:
-          preferredDuringSchedulingIgnoredDuringExecution:
-          - preference:
-              matchExpressions:
-              - key: dedicated
-                operator: In
-                values:
-                - lfmerge
-            weight: 1
-      # required to ensure this container makes it to lfmerge's dedicated node
-      tolerations:
-      - effect: NoSchedule
-        key: dedicated
-        operator: Equal
-        value: lfmerge
       volumes:
       - name: assets
         persistentVolumeClaim:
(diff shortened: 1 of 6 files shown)


megahirt:
@longrunningprocess thanks for this!  I have a few questions that the PR doesn't specifically speak to:

- do we no longer need nodeAffinity for lfmerge?  These sections were deleted
- the approach here seems to be to assign initial memory allocations and hard memory limits for each container, not just lfmerge.  I had expected to see just the lfmerge container have memory limits in place since that's the problem container we are dealing with.  Thoughts?
- If we now have hard limits in place for every container, there's a chance that we could see increased failures in some hard-to-repro edge cases where memory exceeds the limits.  I don't know why we would subject ourselves to such limits if we're not currently having an issue with it.  Thanks for your explanation.

longrunningprocess:
> @longrunningprocess thanks for this! I have a few questions that the PR doesn't specifically speak to:
> 
> * do we no longer need nodeAffinity for lfmerge?  These sections were deleted

No, those were there with `tolerations` to keep lfmerge on a specific node that was beefed up.  that will no longer be necessary with our limits.

> * the approach here seems to be to assign initial memory allocations and hard memory limits for each container, not just lfmerge.  I had expected to see just the lfmerge container have memory limits in place since that's the problem container we are dealing with.  Thoughts?

Gotta have them for all otherwise any container could potentially consume everything on the node... this is a decision that is broader than just LF.

> * If we now have hard limits in place for every container, there's a chance that we could see increased failures in some hard-to-repro edge cases where memory exceeds the limits.  I don't know why we would subject ourselves to such limits if we're not currently having an issue with it.  Thanks for your explanation.

very unlikely based on testing we've done in the new cluster as well as reviewing metrics from our existing prod containers, these limits are actually quite generous.  Worst-case scenario however will be that I guessed wrong and we'll bump them, it's a very quick fix.

megahirt (approved):
Great work @longrunningprocess.  Thank you very much for the explanation and further conversation on WhatsApp.

PR #1461, opened 2022-09-13
Fixed bug where num of entries was being limited to 1,000 on project dashboard

### LINKED ISSUE #1460: Project dashboard limited to max of 1,000 entries
While testing #1416 , I noticed entries are limited to 1,000.

longrunningprocess:
## Description

The number of entries on the project dashboard were being limited to 1,000.

Fixes #1460 

### Type of Change

## Screenshots

image

## Testing on your branch

- [ ] Visit the project landing page for a project with more than 1,000 entries, e.g., test-rwr-flex, qxh-flex.

## Checklist

### DIFF
--- a/next-app/src/routes/projects/[project_code]/meta/+server.ts
+++ b/next-app/src/routes/projects/[project_code]/meta/+server.ts
@@ -6,7 +6,7 @@ export async function get({ project_code, cookie }) {
 		{ entries, comments }
 	] = await Promise.all([
 		sf({ name: 'project_read_by_code', args: [ project_code ], cookie }),
-		sf({ name: 'lex_stats', args: [ project_code ], cookie }),
+		sf({ name: 'lex_stats_all', args: [ project_code ], cookie }),
 	])
 
 	const entries_with_picture = entries.filter(has_picture)
--- a/src/Api/Service/Sf.php
+++ b/src/Api/Service/Sf.php
@@ -558,6 +558,18 @@ class Sf
         throw new UserUnauthorizedException("User $this->userId is not a member of project $projectCode");
 	}
 
+	public function lex_stats_all($projectCode)
+	{
+		$projectModel = ProjectModel::getByProjectCode($projectCode);
+        $user = new UserModel($this->userId);
+
+        if ($user->isMemberOfProject($projectModel->id->asString())) {
+            return LexDbeDto::encode($projectModel->id->asString(), $this->userId, 1);
+        }
+
+        throw new UserUnauthorizedException("User $this->userId is not a member of project $projectCode");
+	}
+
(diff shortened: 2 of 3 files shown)


longrunningprocess:
We should make sure we're ok with the perf hit of this page since we will gather all entries, not just 1,000 at a time.  If the perf is unacceptable, there are plenty of solutions we can consider.

next-app/src/routes/projects/[project_code]/meta/+server.ts
```diff
@@ -6,7 +6,7 @@ export async function get({ project_code, cookie }) {
 		{ entries, comments }
 	] = await Promise.all([
 		sf({ name: 'project_read_by_code', args: [ project_code ], cookie }),
-		sf({ name: 'lex_stats', args: [ project_code ], cookie }),
```
Comment (megahirt) [inline next-app/src/routes/projects/[project_code]/meta/+server.ts:9]: just curious, are you still using lex_stats in addition to this?  Or are you only using lex_stats_all now?  If it's no longer used, how about we remove it from PHP.

megahirt (approved):
LGTM.  This is fixing up stuff related to the activity feed.  My sense is that future data accesses should go directly into Mongo from Svelte.  Perhaps the next thing to do, is to refactor the activity feed in a new paradigm that accesses Mongo directly.  we need to spec this out as a separate task.

I still think there is value in developing an external API with proper JWT and OAuth that could be used by both Language Forge and other apps.

You can resolve the comment yourself so this will merge.

next-app/src/routes/projects/[project_code]/meta/+server.ts
Comment (longrunningprocess) [reply next-app/src/routes/projects/[project_code]/meta/+server.ts:9]: I do not use lex_stats but it might be used in the PHP code... if you'd like I'll go digging around to make sure

src/Api/Service/Sf.php
```diff
@@ -558,6 +558,18 @@ public function lex_stats($projectCode)
         throw new UserUnauthorizedException("User $this->userId is not a member of project $projectCode");
 	}
 
+	public function lex_stats_all($projectCode)
+	{
+		$projectModel = ProjectModel::getByProjectCode($projectCode);
+        $user = new UserModel($this->userId);
+
+        if ($user->isMemberOfProject($projectModel->id->asString())) {
+            return LexDbeDto::encode($projectModel->id->asString(), $this->userId, 1);
```
megahirt, line 567: So, on later inspection of this code, I'm not really clear how passing 1 into the lastFetchTime achieves more than 1000 entries at a time... maybe it does for you but I actually don't understand how it works.

megahirt: ok, yeah that's definitely a hack.  I want to make a proper stats endpoint to support what you are doing with the dashboard.  That will be a separate card.

PR #1466, opened 2022-09-15
Add symbolic link steps for Windows. Fix URLs. Docs clarification.

myieye:
## Description

Made some improvements to the docs:
1. Symbolic links don't just work out of the box on Windows, so I added the necessary steps (and removed an unsed/outdated image on the same topic)
2. Fixed image URLs, added some clarification etc.

### Type of Change

- Doc update

## Checklist


docs/DEVELOPER.md
```diff
@@ -4,12 +4,15 @@ Welcome!  We're glad that you are interested in helping develop Language Forge.
...
-5. Clone the repo:  `git clone https://github.com/sillsdev/web-languageforge`
+1. Install [Docker](https://www.docker.com/get-started).
+    1. Linux users will need some additional steps: Please visit https://docs.docker.com/compose/install for info on installing the engine and compose.
+    2. Windows users, use Ubuntu (WSL) and follow the instructions at https://docs.docker.com/engine/install/ubuntu/. Then permit your user (and not just "sudo") to contact the Docker daemon by running `sudo usermod -aG docker $yourUsername` and `sudo chmod 666 /var/run/docker.sock`.
+3. Install [Make](https://www.gnu.org/software/make/): `sudo apt install make`.  This is actually optional but simplifies things a bit.
+4. Install [Node 16.14.0](https://nodejs.org/en/download/).  We recommend using a Node version manager e.g. [nvm](https://github.com/nvm-sh/nvm#installation-and-update).
+5. Clone the repo: `git clone -c core.symlinks=true https://github.com/sillsdev/web-languageforge`. Windows "Developer Mode" must be enabled before cloning or the symbolic links won't be created.
```
megahirt, line 12: Is the clone step specific to cloning on Windows?  I haven't run into this.  And by Windows I'm not talking about WSL but straight up Windows (which is what I'm assuming you are referring to).

megahirt (approved):
These are some great improvements to the docs - thanks!

docs/DEVELOPER.md
myieye: Both `-c core.symlinks=true` and enabling "Developer Mode" are specific to cloning into a Windows file system regardless of whether you're running the command in Windows or WSL.  (`-c core.symlinks=true` is not necessary if `core.symlinks` is explicitly enabled globally)

I didn't think it was worth putting all that info in the docs, because there are no negative side effects to adding the argument on other systems. But, this would probably be an improvement: `Windows "Developer Mode" must be enabled before cloning to a Windows file system or the symbolic links won't be created.`

I realize now that it's suboptimal to clone to Windows, but it still seems reasonable to try to support it.

PR #1471, opened 2022-09-20
Make `make clean-powerwash` delete base-php image

megahirt:
## Description

Creating a PR on @rmunn 's behalf

The `docker compose down --rmi all` command doesn't appear to delete the base-php image, for reasons unknown. Since it occasionally will change, we'll just delete it explicitly in `make clean-powerwash`.

### Type of Change

- chore


longrunningprocess (approved):
It's not removed because that image isn't built in the docker-compose.yml.

Issue #1402, opened 2022-05-07
bug: Race condition in Playwright webServer config


rmunn:
**Describe the bug**
There's a race condition in the [Playwright `webServer` config](https://github.com/sillsdev/web-languageforge/blob/8164ccf592aa0f6b0054882c3f1b0bccd4a3811f/playwright.config.ts#L101): Playwright watches for the specified port to open, and as soon as the port is open, it assumes the configured webserver is up and running. Whereas Docker-Compose will open the port on localhost as soon as the container is up, even if nothing is listening to the port inside the container yet. That means that once we run `docker-compose up`, there's a race condition between the Apache process starting inside the container, and Playwright trying to talk to the webserver. If Playwright wins the race, it exits immediately with an ECONNREFUSED error, because nothing was actually listening on the port yet.

**To Reproduce**
1. Stop the webserver (via `docker-compose down`)
2. Run `npx playwright test`
3. Sometimes it will work, sometimes it will exit immediately with an error
4. If it exits, re-running it will work.

**Expected behavior**
We need to have a way to tell Playwright to wait for the port to actually be *ready* rather than merely open. That's not possible right now, but [in a future version of Playwright](https://github.com/microsoft/playwright/pull/13779#discussion_r861344756) there should be a `readinessCheck` parameter in the Playwright config. Once that is implemented, we'll be able to tell Playwright "Wait until this URL returns 200", which will solve this race condition and let us close the issue.

myieye:
Just for the record, this is probably the relevant log. I forced this by just removing most of the contents of `docker/app/Dockerfile` so that the application doesn't run:

```
Running 1 test using 1 worker

[WebServer] lf-db is up-to-date
[WebServer] ld-db is up-to-date
[WebServer] lf-mail is up-to-date
[WebServer] ld-api is up-to-date
Starting app-for-playwright ... 
[WebServer] Starting app-for-playwright ... done
Error: apiRequestContext.post: connect ECONNREFUSED 127.0.0.1:3238
=========================== logs ===========================
→ POST http://localhost:3238/api/testControl
  user-agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/[53](https://github.com/myieye/web-languageforge/actions/runs/3088369860/jobs/4994793273#step:7:54)7.36 (KHTML, like Gecko) HeadlessChrome/105.0.5195.19 Safari/537.36
  accept: */*
  accept-encoding: gzip,deflate,br
  content-type: application/json
[...]

PR #1472, opened 2022-09-20
Local and GitHub playwright testing improvements

myieye:
## Description

Various minor playwright improvements:
- Cache playwright browsers in GHA (per playwright version as documented). Not a huge performance boost, but not a huge change either
- Upload the playwright results as a GitHub artifact, which is then available for download (see screenshot)
- Allow passing playwright params to make file. E.g. `make playwright-tests params="--headed"`
- Allow running make files with npm (I find it very handy for the commands to be directory independent) E.g. `npm run make playwright-tests params="--headed"`
- Wait for forgotten promise in `globalSetup.ts` and add ts-lint rule to watch for cases like this
- Close browserContexts in `globalSetup.ts`, because: (1) it's documented as best practice, (2) it doesn't break anything and (3) it eliminates a potential culprit of "socket hang up" mentioned in #1402

### Type of Change

- New feature

## Screenshots

image

## Checklist

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -55,6 +55,12 @@ jobs:
           docker-compose -f docker/docker-compose.yml up -d app-for-playwright
           npx playwright install chromium
           npx playwright test
+      - name: Upload test results
+        if: always()
+        uses: actions/upload-artifact@v2
+        with:
+          name: test-results
+          path: test-results
 
   # protractor-tests:
   #   runs-on: ubuntu-latest
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -47,6 +47,22 @@ jobs:
         run: |
           npm ci
           docker-compose build app
+      - name: Get installed Playwright version
+        id: playwright-version
+        # https://github.com/microsoft/playwright/issues/7249#issuecomment-1201476980
+        run: echo -n "::set-output name=version::$(npm ls @playwright/test --json | jq --raw-output '.dependencies["@playwright/test"].version')"
+
+        # https://github.com/microsoft/playwright/issues/7249#issuecomment-1154603556
+      - name: Cache playwright browsers
+        uses: actions/cache@v3
+        id: playwright-cache
+        with:
+          path: '~/.cache/ms-playwright'
+          # Make each playwright version use its own cache in case it uses different browser versions
+          # Cache entries are deleted if not accessed for 7 days
+          # https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy
+          key: 'playwright-${{ steps.playwright-version.outputs.version }}'
+
       -
         name: Playwright E2E Tests
         working-directory: .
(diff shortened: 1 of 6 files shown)


.github/workflows/pull-request.yml
```diff
@@ -47,6 +47,22 @@ jobs:
         run: |
           npm ci
           docker-compose build app
+      - name: Get installed Playwright version
+        id: playwright-version
+        # https://github.com/microsoft/playwright/issues/7249#issuecomment-1201476980
+        run: echo -n "::set-output name=version::$(npm ls @playwright/test --json | jq --raw-output '.dependencies["@playwright/test"].version')"
+
+        # https://github.com/microsoft/playwright/issues/7249#issuecomment-1154603556
+      - name: Cache playwright browsers
```
JeanneSon, line 56: Unfortunately, this does not seem to work for me (forked this repo and pushed twice; the browsers were also downloaded the second time)
image

.github/workflows/pull-request.yml
```diff
@@ -55,6 +71,12 @@ jobs:
           docker-compose -f docker/docker-compose.yml up -d app-for-playwright
           npx playwright install chromium
           npx playwright test
+      - name: Upload test results
+        if: always()
+        uses: actions/upload-artifact@v2
```
JeanneSon, line 76: @myieye Could you add a description how these uploads can be downloaded?

myieye: @JeanneSon can you link me to the run? The cache only gets written to if the whole job was successful. Perhaps it failed?

.github/workflows/pull-request.yml
myieye: @JeanneSon can you link me to the run? The cache only gets written to if the whole job was successful. Perhaps it failed?

myieye: Nevermind, it's obviously pretty easy to find the run myself 😆.
Yeah, [the first execution failed](https://github.com/JeanneSon/fork-web-languageforge/actions/runs/3096685403/jobs/5012515627), so you can see that the "Post Cache playwright browsers" step didn't happen. That's when things get cached. The "Cache playwright browsers" step is when the cache gets loaded if it exists.
image

The more recent run is green, so if you trigger the action again it **will** use the cache. Pinky promise 😰.

.github/workflows/pull-request.yml
myieye: @JeanneSon do you mean at a description/comment in the code?

You can see an artifact near the bottom of this page: https://github.com/JeanneSon/fork-web-languageforge/actions/runs/3096685403

I just realized that, beacuse playwright doesn't save test-results for successful executions (and there's not artifact to upload/download) we'd never be able to compare a failure to a success. Perhaps that's not so great. 🤔

docker/Makefile
```diff
@@ -23,7 +23,7 @@ playwright-tests:
     # wait until the app-for-playwright service is serving up HTTP before continuing
 	until curl localhost:3238 > /dev/null 2>&1; do sleep 1; done
 
-	cd .. && npx playwright install chromium && npx playwright test
+	cd .. && npx playwright install chromium && npx playwright test $(params)
```
megahirt, line 26: nice!

megahirt (approved):
These are great additions, thanks @myieye !

PR #1474, opened 2022-09-21
Bugfix/use media recorder

### LINKED ISSUE #1381: bug: audio jumpiness in LF audio recordings
**Describe the bug**
Audio recorded through the Language Forge website (whether mobile device or PC) results in degraded audio recording, identified by  "audio jumpiness", clicks or pops, or even repeated audio segments of ~100ms length

**To Reproduce**
1. Go to a project that has an audio writing system e.g. chris-test-sr1
2. Click on the microphone button to the right of the field (assuming no previous recording)
3. Click the large microphone button to start recording.
4. Record a 10 second clip with constant talking.
5. Save the recording by clicking the large microphone button again and then clicking the checkmark.
6. Play back the audio you just recorded and observe that it contains "jumpiness".  In my experience, the jumpiness occurs most often within the first second of the audio, but can also appear later on in the recording.

**Expected behavior**
Audio is recorded and saved as expected without additional degraded audio artifacts

**Screenshots**
image

Reproduced on:
- Windows 10 - MS Edge (Chris)
- Android 12 Chrome (Chris)
- iPhone iOS 14 (Jay)

Additional context:
As one of the developers, I (Chris) know that this used to work 5 years ago when it was introduced.  However, frequent OS, browser and now LF dependency upgrades have changed the environment in which this code runs and there is obviously a regression identified by this bug report.

Proposed solution:
@rmunn and I did some initial investigation into the code responsible for this feature, and found that while it uses the standard [AudioContext API](https://developer.mozilla.org/en-US/docs/Web/API/AudioContext) (what was available when this feature was developed) there is a new [MediaStream Recording API ](https://developer.mozilla.org/en-US/docs/Web/API/MediaStream_Recording_API )that we can use that should fix this bug and simplify/update our code.  See #1382 for more details.

### LINKED ISSUE #1382: feat: Switch audio recording to use MediaRecorder API
While looking into #1381, I discovered that our audio-recording component could be updated. It currently uses the [AudioContext API](https://developer.mozilla.org/en-US/docs/Web/API/AudioContext), and then has to do quite a lot of work to massage the raw PCM data into a form that's acceptable to pass into lamejs to convert to MP3.

But there's a newer Web API, supported by all browsers except IE: [MediaRecorder](https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder). The last browser to support this was Safari, which added support in Safari 14.1, as of April 2021. The MediaRecorder API allows you to write code like this:

```js
navigator.mediaDevices.getUserMedia({audio: true}).then(stream => {
  const recorder = new MediaRecorder(stream, { mimeType: "audio/mp3" });
  elem.addEventListener('dataavailable', event => saveDataToFile(event));
  recorder.start();
  // Sometime later, probably in the click handler for your stop button:
  recorder.stop();  // This triggers a `dataavailable` event
});
```

No need for external libraries like lamejs; the browser handles the encoding for you and presents you with data in the format you specify (MP3, MP4, WAV...). The media type can include a `codecs` parameter, such as `audio/mpeg;codecs="opus"`, and if the user's browser supports it, you'll get an MP4 container with Opus-format audio inside. (You can query which formats the browser supports using `MediaRecorder.isTypeSupported` beforehand).

Change https://github.com/sillsdev/web-languageforge/blob/9f6519e93d9617f10cc9f689fa32f6bb0dddf603/src/angular-app/bellows/shared/audio-recorder/audio-recorder.component.ts to use MediaRecorder to encode the MP3 files. (Or whatever other format; we might want to move up to M4A, for example). This will also reduce our external Javascript dependencies and remove some tech debt.

Doing nothing and leaving our code as-is, since it does work.

### LINKED ISSUE #1399: feat: Allow more audio media types, including FLAC
As a follow-up to https://github.com/sillsdev/web-languageforge/issues/1383, it seems to me that we should allow other file types to be uploaded. For example, the Free Lossless Audio Codec is a format that usually is 50-60% smaller than the .wav file of the same audio, but since it's a lossless codec, it can decompress into a .wav file that's byte-for-byte identical to the original. Also, if we're going to support the FLAC format (MIME type `audio/flac`, with `audio/x-flac` sometimes encountered even though it's non-standard), it would make sense to support the Ogg container format that FLAC files are often embedded in (MIME type `audio/ogg` and/or `video/ogg`).

Add the `.flac` extension, and the `audio/flac` and `audio/x-flac` MIME types, to our file type whitelist in the `uploadAudioFile` method of `LexUploadCommands`. And have a discussion about whether to support other file types.

We don't necessarily need to do this urgently, as I'm not aware of anyone actually trying to upload FLAC or Ogg format files to Language Forge. Ogg is rarely used, for example. But since FLAC is lossless, and linguists often want to save their audio files in a lossless format, I wouldn't be surprised if at some point, someone were to decide to upload .flac files to save space. When (and if) that happens, it would be nice if we already have support in place for that format.

For other formats, it's more of a "well, why not?" case, even though I don't expect most of them to be used. E.g., https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Containers#ogg says "While Ogg has been around for a long time, it has never gained the wide support needed to make it a good first choice for a media container. You are typically better off using WebM ..." in its advice on which media type to offer from your website. So I don't expect people to be beating down our door with .ogg files they want to upload. But if it's a commonly-used format, or a rarely-used format that <i>would</i> be useful for linguists' use cases (e.g. FLAC), there's probably not a good reason <i>not</i> to support it.

laineyhm:
## Description

MediaRecorder is the most current API for audio recording in browser. With MediaRecorder, Language Forge will capture audio in WEBM containers with Opus codec. 

ffmpeg is added as a dependency to perform audio conversions. The audio files are converted and stored as either .wav or .mp3 files to accommodate send/receive with Fieldworks software. If the recording is short enough (<6 s), based on settings in conversion to WAV, it will be within the 1MB file size limit. Recordings longer than 6 seconds are converted to MP3. 

webm-fix-duration is added as a small (25 KB) dependency. Currently, navigator.mediaDevices.getUserMedia with MediaRecorder returns WEBM file blobs that have no duration metadata. This affects the display of the sound player as well as decisions for encoding. The webm-fix-duration package takes care of that by appending the duration data to the blob so it may be accessed by the app.

Fixes #1382
Fixes #1381
Fixes #1399 

### Type of Change

## Screenshots

The UI works as before. In the console here, see saved files converted to .wav and .mp3.

image

## Testing on your branch

- [ ] Test the following in Chrome, Firefox, Edge, and Safari, as well as on iOS and Android mobile devices.
To test on mobile, install ngrok and run:
`./ngrok http http://localhost`
in a bash terminal. The same command with https://localhost may not work, so be careful to try http://localhost in particular. You will get a URL that you can use for testing from a mobile device.

1. Open the HTML docker container in lf-app. Open assets/lexicon/sf_my_dictionary/audio. This is where all saved recordings and uploaded audio files go.
2. Record a short (<6 second) clip online, play it back, and save it. See if it is found in the container twice: once in the .webm format  and once as a .wav file.
3. Record a longer (>6 second) clip online, play it back, and save it. See if it is found in the container twice: once in the .webm format  and once as an .mp3 file.
4. Upload a long and a short clip of each of these file types: .ogg, .flac, .m4a. See if they are found in the container twice, with conversion to .wav for the short files and .mp3 for the long files.
5. Upload an .mp3 and a .wav file to see if these files are stored in the container "as-is"; no conversion, no second copy. Here is a folder of test files to use: [Audio Test Upload Files.zip](https://github.com/sillsdev/web-languageforge/files/9642997/Audio.Test.Upload.Files.zip)
- [ ] See if files added online come into FLEx via Send/Receive.

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works -- have not done this, maybe I can learn.

### DIFF
--- a/src/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php
+++ b/src/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php
@@ -18,6 +18,7 @@ class LexUploadCommands
         ".lift"
     );
 
+
     /**
      * Upload an audio file
      *
@@ -53,9 +54,11 @@ class LexUploadCommands
         $allowedTypes = array(
             "application/octet-stream",
 
-            // allow m4a audio uploads, which curiously has a mime type of video/mp4
+            // allow m4a audio uploads, which curiously has a mime type of video/mp4, audio/m4a, audio/mp4, or audio/x-m4a
             "video/mp4",
-
+            "audio/m4a",
+            "audio/mp4",
+            "audio/x-m4a",
             "audio/mpeg",
             "audio/x-mpeg",
             "audio/mp3",
@@ -66,27 +69,70 @@ class LexUploadCommands
             "audio/x-mpg",
             "audio/x-mpegaudio",
             "audio/x-wav",
-            "audio/wav"
+            "audio/wav",
+            "audio/flac",
+            "audio/x-flac",
+            "audio/ogg",
+            "audio/webm",
+            // allow Google Chrome to handle MediaRecorder recordings as video/webm MimeType
+            "video/webm"
         );
         $allowedExtensions = array(
             ".mp3",
             ".mpa",
             ".mpg",
             ".m4a",
-            ".wav"
+            ".wav",
+            ".ogg",
+            ".flac",
+            ".webm",
         );
 
         $response = new UploadResponse();
+
         if (in_array(strtolower($fileType), $allowedTypes) && in_array(strtolower($fileExt), $allowedExtensions)) {
 
             // make the folders if they don't exist
             $project->createAssetsFolders();
             $folderPath = $project->getAudioFolderPath();
 
-            // move uploaded file from tmp location to assets
+            // move uploaded/recorded file from tmp location to assets
             $filePath = self::mediaFilePath($folderPath, $fileNamePrefix, $fileName);
             $moveOk = copy($tmpFilePath, $filePath);
-            @unlink($tmpFilePath);
+
+            // convert audio file to mp3 or wav format if necessary
+            // FLEx only supports mp3 or wav format as of 2022-09
+
+            if (strcmp(strtolower($fileExt), ".mp3") !== 0 && strcmp(strtolower($fileExt), ".wav") !== 0) {
+                //First, find the duration of the file
+                $sanitizedTmpFilePath = escapeshellarg($tmpFilePath);
+                $ffprobeCommand = `ffprobe -i $sanitizedTmpFilePath -show_entries format=duration -v quiet -of csv="p=0" 2> /dev/null`;
+                $audioDuration = floatval($ffprobeCommand);
+
+                // Convert to .wav if the result will be less than 1 MB at 165Kb/s (recording is shorter than 6 seconds)
+                // and .mp3 otherwise (recording is longer than 6 seconds)
+                $extensionlessTmpFilePath = substr($sanitizedTmpFilePath, 0, strrpos($sanitizedTmpFilePath, strtolower($fileExt)));
+                $extensionlessFileName = substr($fileName, 0, strrpos($fileName, strtolower($fileExt)));
+
+                if($audioDuration < 6){
+                    `ffmpeg -i $sanitizedTmpFilePath -b:a 165K -maxrate 165K -bufsize 80K $extensionlessTmpFilePath.wav 2> /dev/null')`;
+                    `mv $sanitizedTmpFilePath $extensionlessTmpFilePath.wav 2> /dev/null`;
+                    $fileName = $extensionlessFileName . '.wav';
+                    $tmpFilePath = $extensionlessTmpFilePath. '.wav';
+                }
+                else{
+                    `ffmpeg -i $sanitizedTmpFilePath -b:a 165K -maxrate 165K -bufsize 80K $extensionlessTmpFilePath.mp3 2> /dev/null')`;
+                    `mv $sanitizedTmpFilePath $extensionlessTmpFilePath.mp3 2> /dev/null`;
+                    $fileName = $extensionlessFileName . '.mp3';
+                    $tmpFilePath = $extensionlessTmpFilePath . '.mp3';
+                }
+
+
+                // move uploaded file from tmp location to assets
+                $filePath = self::mediaFilePath($folderPath, $fileNamePrefix, $fileName);
+                $moveOk = copy($tmpFilePath, $filePath);
+                @unlink($tmpFilePath);
+            }
 
             // construct server response
             if ($moveOk && $tmpFilePath) {
@@ -95,10 +141,11 @@ class LexUploadCommands
                 $data->fileName = $fileNamePrefix . '_' . $fileName;
                 $response->result = true;
 
-                if (array_key_exists('previousFilename', $_POST)) {
-                    $previousFilename = $_POST['previousFilename'];
-                    self::deleteMediaFile($projectId, $mediaType, $previousFilename);
-                }
+                //Uncomment to ensure that only one format for each audio file is stored in the assets. We want to keep up to two formats right now (09-2022): the original and if needed, a FLEx-compatible one
+                // if (array_key_exists('previousFilename', $_POST)) {
+                //     $previousFilename = $_POST['previousFilename'];
+                //     self::deleteMediaFile($projectId, $mediaType, $previousFilename);
+                // }
             } else {
                 $data = new ErrorResult();
                 $data->errorType = 'UserMessage';
--- a/src/angular-app/bellows/shared/audio-recorder/audio-recorder.component.ts
+++ b/src/angular-app/bellows/shared/audio-recorder/audio-recorder.component.ts
@@ -1,24 +1,98 @@
-import * as angular from 'angular';
-import * as lamejs from 'lamejs';
-
-declare var webkitAudioContext: {
-  new(): AudioContext;
-};
+import { webmFixDuration } from "webm-fix-duration";
+import * as angular from "angular";
 
 export class AudioRecorderController implements angular.IController {
+  static $inject = ["$interval", "$scope"];
 
-  static $inject = ['$interval', '$scope'];
-  constructor(private $interval: angular.IIntervalService, private $scope: angular.IScope) {}
-
+  mediaRecorder: MediaRecorder;
+  chunks: string[] = [];
   isRecording = false;
   hasRecorded = false;
+  recordingStartTime: Date;
   audioSrc: string;
   blob: Blob;
   recordingTime: string;
   errorMessage: string;
-  stopMediaStream: () => void;
   callback: (blob: Blob) => void;
+  durationInMilliseconds: number;
   interval: angular.IPromise<void>;
+
+  constructor(
+    private $interval: angular.IIntervalService,
+    private $scope: angular.IScope
+  ) {}
+
+  private startRecording() {
+    this.recordingTime = "0:00";
+
+    navigator.mediaDevices.getUserMedia({ audio: true }).then(
+      (stream) => {
+        this.mediaRecorder = new MediaRecorder(stream);
+
+        this.$scope.$apply(() => {
+          this.hasRecorded = true;
+          this.errorMessage = null;
+          this.isRecording = true;
+        });
+
+        this.mediaRecorder.addEventListener(
+          "dataavailable",
+          async (e: { data: any }) => {
+            this.chunks.push(e.data);
+            var roughBlob = new Blob(this.chunks, {
+              type: "audio/webm; codecs=opus",
+            });
+            //In some browsers (Chrome, Edge, ...) navigator.mediaDevices.getUserMedia with MediaRecorder creates WEBM files without duration metadata  //2022-09
+            //webmFixDuration appends missing duration metadata to a WEBM file blob.
+            this.blob = await webmFixDuration(
+              roughBlob,
+              this.durationInMilliseconds,
+              "audio/webm; codecs=opus"
+            );
+            this.chunks = [];
+            this.audioSrc = window.URL.createObjectURL(this.blob);
+            this.$scope.$digest();
+          }
+        );
+
+        this.recordingStartTime = new Date();
+
+        this.interval = this.$interval(() => {
+          const seconds = Math.floor(
+            (new Date().getTime() - this.recordingStartTime.getTime()) / 1000
+          );
+          this.recordingTime =
+            Math.floor(seconds / 60) +
+            ":" +
+            (seconds % 60 < 10 ? "0" : "") +
+            (seconds % 60);
+        }, 1000);
+
+        this.mediaRecorder.start();
+      },
+      (err) => {
+        this.$scope.$apply(() => {
+          this.errorMessage = "Unable to record audio from your microphone.";
+          this.isRecording = false;
+          this.hasRecorded = false;
+        });
+
+        console.error(err);
+      }
+    );
+  }
+
+  private stopRecording() {
+    this.durationInMilliseconds = Math.floor(
+      new Date().getTime() - this.recordingStartTime.getTime()
+    );
+
+    this.mediaRecorder.stop();
+
+    if (this.interval) {
+      this.$interval.cancel(this.interval);
+    }
+  }
 
   toggleRecording() {
     if (this.isRecording) this.stopRecording();
@@ -27,7 +101,9 @@ export class AudioRecorderController implements angular.IController {
   }
 
   close() {
-    this.stopRecording();
+    if (this.isRecording) {
+      this.stopRecording();
+    }
     this.callback(null);
   }
 
@@ -36,132 +112,26 @@ export class AudioRecorderController implements angular.IController {
   }
 
   recordingSupported() {
-    return navigator.mediaDevices && navigator.mediaDevices.enumerateDevices && navigator.mediaDevices.getUserMedia &&
-      ((window as any).AudioContext || (window as any).webkitAudioContext);
+    return (
+      navigator.mediaDevices &&
+      navigator.mediaDevices.enumerateDevices &&
+      navigator.mediaDevices.getUserMedia &&
+      ((window as any).AudioContext || (window as any).webkitAudioContext)
+    );
   }
 
   $onDestroy() {
-    this.stopRecording();
-  }
-
-  private startRecording() {
-    this.recordingTime = '0:00';
-
-    navigator.mediaDevices.getUserMedia({audio: true}).then(stream => {
-
-      this.$scope.$apply(() => {
-        this.hasRecorded = true;
-        this.errorMessage = null;
-        this.isRecording = true;
-      });
-
-      const recordingStartTime = new Date();
-      // webkit prefix required for Safari
-      const context = (window as any).AudioContext ? new AudioContext() : new webkitAudioContext();
-      const bufferSize = 0;
-      const channels = 1;
-      const processor = context.createScriptProcessor(bufferSize, channels, channels);
-      context.createMediaStreamSource(stream).connect(processor);
-      processor.connect(context.destination);
-      const sampleRate = context.sampleRate;
-      const bitrate = 128;
-      const mp3Encoder = new MP3Encoder(channels, sampleRate, bitrate);
-
-      function handleAudioData(event: AudioProcessingEvent) {
-        mp3Encoder.appendData(event.inputBuffer.getChannelData(0));
-      }
-      processor.addEventListener('audioprocess', handleAudioData);
-
-      mp3Encoder.onMp3Blob(blob => {
-        this.blob = blob;
-        this.audioSrc = URL.createObjectURL(blob);
-      });
-
-      this.interval = this.$interval(() => {
-        const seconds = Math.floor((new Date().getTime() - recordingStartTime.getTime()) / 1000);
-        this.recordingTime = Math.floor(seconds / 60) + ':' + (seconds % 60 < 10 ? '0' : '') + seconds % 60;
-      }, 1000);
-
-      this.stopMediaStream = () => {
-        processor.removeEventListener('audioprocess', handleAudioData);
-        mp3Encoder.end();
-        stream.getAudioTracks()[0].stop();
-      };
-
-    }, err => {
-      this.$scope.$apply(() => {
-        this.errorMessage = 'Unable to record audio from your microphone.';
-        this.isRecording = false;
-        this.hasRecorded = false;
-      });
-      console.error(err);
-    });
-  }
-
-  private stopRecording() {
-    if (this.interval) this.$interval.cancel(this.interval);
-    if (this.stopMediaStream) this.stopMediaStream();
-  }
-
-}
-
-export class MP3Encoder {
-
-  constructor(private channels: number, private sampleRate: number, private bitrate: number) {}
-
-  buffer: Float32Array[] = [];
-  mp3BlobListeners: Array<(data: Blob) => void> = [];
-
-  lame = new lamejs.Mp3Encoder(this.channels, this.sampleRate, this.bitrate);
-
-  appendData(data: Float32Array) {
-    // copy the data, because Chromium 66.0.3359.139 overwrites it (FF 60.0.1 does not)
-    this.buffer.push(data.slice(0, data.length));
-  }
-
-  end() {
-    // Flatten the buffer array while converting it from Float32 to Int16
-    const pcmData = new Int16Array(this.buffer.length * this.buffer[0].length);
-
-    // max and min 16-bit values
-    const max = 0x7FFF;
-    const min = 0x8000;
-    let index = 0;
-    for (let i = 0, len1 = this.buffer.length; i < len1; ++i) {
-      const chunk = this.buffer[i];
-      for (let j = 0, len2 = chunk.length; j < len2; ++j) {
-        pcmData[index] = chunk[j] < 0 ? chunk[j] * min : chunk[j] * max;
-        ++index;
-      }
+    if (this.isRecording) {
+      this.stopRecording();
     }
-
-    const blockSize = 1152;
-    const mp3Data: Int8Array[] = [];
-
-    let encoded: Int8Array;
-    for (let i = 0; i < pcmData.length; i += blockSize) {
-      const chunk = pcmData.subarray(i, i + blockSize);
-      // This line takes the most time of this function, about 90%
-      encoded = this.lame.encodeBuffer(chunk);
-      if (chunk.length > 0) mp3Data.push(new Int8Array(encoded));
-    }
-
-    encoded = this.lame.flush();
-    if (encoded.length > 0) mp3Data.push(new Int8Array(encoded));
-    const blob = new Blob(mp3Data, {type: 'audio/mp3'});
-
-    this.mp3BlobListeners.forEach(cb => cb(blob));
-  }
-
-  onMp3Blob(cb: (data: Blob) => void) {
-    this.mp3BlobListeners.push(cb);
   }
 }
 
 export const AudioRecorderComponent: angular.IComponentOptions = {
   bindings: {
-    callback: '<' // TODO probably change to > or <, not sure which
+    callback: "<",
   },
   controller: AudioRecorderController,
-  templateUrl: '/angular-app/bellows/shared/audio-recorder/audio-recorder.component.html'
+  templateUrl:
+    "/angular-app/bellows/shared/audio-recorder/audio-recorder.component.html",
 };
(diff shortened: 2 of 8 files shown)


megahirt:
Also fixes #1381

Great job @laineyhm !

The unit tests are failing because we need to build and push a base-php image that contains ffmpeg.  We can do that together tomorrow.

src/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php
```diff
@@ -66,23 +67,55 @@ public static function uploadAudioFile($projectId, $mediaType, $tmpFilePath)
...
             // make the folders if they don't exist
             $project->createAssetsFolders();
             $folderPath = $project->getAudioFolderPath();
 
+            // convert audio file to mp3 or wav format if necessary //LHM
+            // FLEx only supports mp3 or wav format as of 2022-09
+
+            // First, find duration of audio file
+            // @return string
+            function getDuration($path) {
```
longrunningprocess, line 98: I want to be sure all of these variables are either generated and trusted by our code or if there is user input being used, it's properly sanitized since this stuff is being released to the shell...

src/angular-app/bellows/shared/audio-recorder/audio-recorder.component.ts
```diff
@@ -1,163 +1,126 @@
...
-  new(): AudioContext;
-};
+declare var MediaRecorder: any;
 
 export class AudioRecorderController implements angular.IController {
 
   static $inject = ['$interval', '$scope'];
-  constructor(private $interval: angular.IIntervalService, private $scope: angular.IScope) {}
 
+  mediaRecorder:any;
```
longrunningprocess, line 10:
```suggestion
  mediaRecorder: any;
```

src/angular-app/bellows/shared/sound-player.component.html
```diff
@@ -2,9 +2,9 @@
...
 </a>
 <span class="controls">
-    <span data-ng-if="$ctrl.audioElement.duration" class="audio-progress text-muted">
-        {{$ctrl.currentTime()}} / {{$ctrl.duration()}}
+    <span id="audio-progress" data-ng-if="$ctrl.audioElement.duration" class="audio-progress text-muted">
+        {{$ctrl.currentTime()}} {{$ctrl.duration()}}
     </span>
     <!--suppress HtmlFormInputWithoutLabel -->
-    <input class="seek-slider" type="range" min="0" max="{{$ctrl.durationInSeconds()}}" value="0">
+    <input class="seek-slider" type="range" min="0" max="{{$ctrl.durationInSeconds()}}" value="0" step="0.0000000000000001">
```
longrunningprocess, line 9: :-D I'm very curious as to the why of that level of precision!

src/angular-app/bellows/shared/sound-player.component.ts
```diff
@@ -51,6 +54,10 @@ export class SoundController implements angular.IController {
       this.isUserMovingSlider = true;
     });
 
+    // this.audioElement.addEventListener('ended', () => {
```
longrunningprocess, line 57: can these be removed?

src/angular-app/bellows/shared/sound-player.component.ts
```diff
@@ -60,38 +67,67 @@ export class SoundController implements angular.IController {
...
+
   }
 
   currentTimeInSeconds(): number {
     return this.audioElement.currentTime;
   }
 
   durationInSeconds(): number {
-    return this.audioElement.duration;
+    if(this.audioElement.duration !== NaN){
```
longrunningprocess, line 114: a stylistic consideration that makes code easier to read is:

1. eliminate the else in a case like this
2. make you conditionals "positive", e.g., "if it is a number..."

longrunningprocess (changes requested):
there's a lot of complexity in [src/angular-app/bellows/shared/audio-recorder/audio-recorder.component.ts](https://github.com/sillsdev/web-languageforge/pull/1474/files#diff-139bb621d84c71ffa9b5e7fa654e2831bb1e666fdfffff8268a72664219ef880), please be sure we've covered testing well in here.

longrunningprocess:
> Also fixes #1381
> 
> Great job @laineyhm !
> 
> The unit tests are failing because we need to build and push a base-php image that contains ffmpeg. We can do that together tomorrow.

I took care of that (sorry if it was meant to be a teaching moment) https://github.com/sillsdev/web-languageforge/actions/runs/3099569423/jobs/5018860502

longrunningprocess:
unit test failure looks like an inheritance issue that needs to be worked out:  

```
PHP Fatal error:  Cannot redeclare Api\Model\Languageforge\Lexicon\Command\getDuration() (previously declared in /var/www/html/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php:98) in /var/www/html/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php on line 98
```

megahirt:
> unit test failure looks like an inheritance issue that needs to be worked out:  
> 
> ```
> PHP Fatal error:  Cannot redeclare Api\Model\Languageforge\Lexicon\Command\getDuration() (previously declared in /var/www/html/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php:98) in /var/www/html/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php on line 98
> ```

Yeah, I saw that too.  Something is not right with the PHP code as well.  Lainey hasn't been running the unit tests locally so this is probably # the first time she's seeing it.

megahirt:
> I took care of that (sorry if it was meant to be a teaching moment) https://github.com/sillsdev/web-languageforge/actions/runs/3099569423/jobs/5018860502

Totally fine!  Thanks for taking care of that.

src/angular-app/bellows/shared/audio-recorder/audio-recorder.component.ts
laineyhm: Accepted

src/angular-app/bellows/shared/sound-player.component.html
laineyhm: This is the best way I've found to address a puzzle. The puzzle is that the slider is stopping when it has completed its last full step. Then there is a gap between the slider and the end of the slider, and visually, it appears to have not finished playing. The default "step" size is 1 second, and many smaller sizes are also too big. When the step size approaches 0, the gap becomes invisible. So, a very small step size fixes the appearance of the slider.

src/angular-app/bellows/shared/sound-player.component.ts
laineyhm: Yes, thanks for seeing these and commenting.

src/angular-app/bellows/shared/sound-player.component.ts
laineyhm: Thanks for pointing this out, a good consideration to keep in mind. In this case, the conditional is actually no longer needed, as other code now takes care of the duration being returned properly.

src/angular-app/bellows/shared/sound-player.component.ts
```diff
@@ -60,38 +67,67 @@ export class SoundController implements angular.IController {
...
+    }
+    else {
+      return 0;
+    }
+
   }
 
   duration(): string {
-    return SoundController.formatTimestamp(this.audioElement.duration * 1000);
+    if(this.audioElement.duration !== Infinity){
```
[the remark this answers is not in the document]
laineyhm: Thank you! Same comment from me too, the conditional is no longer needed so it has been taken out.

src/angular-app/bellows/shared/sound-player.component.html
laineyhm: I recognize that this is kind of a hack, but I have tried many other logical solutions to no avail. If you find something else that clears up this behavior, please send my way

src/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php
laineyhm: Thanks. I took out the helper function and put those lines of code in the uploadAudioFile method itself. The only new variables introduced store the shell command and the resulting duration from an audio recording. User input is not used. Duration is calculated off of a variable (the filepath) that was already in the code.

src/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php
```diff
@@ -66,23 +67,52 @@ public static function uploadAudioFile($projectId, $mediaType, $tmpFilePath)
...
 
             // make the folders if they don't exist
             $project->createAssetsFolders();
             $folderPath = $project->getAudioFolderPath();
 
+            // convert audio file to mp3 or wav format if necessary //LHM
+            // FLEx only supports mp3 or wav format as of 2022-09
+
+            //First, find the duration of the file
+            $ffprobeCommand = 'ffprobe -i ' . $tmpFilePath . ' -show_entries format=duration -v quiet -of csv="p=0" 2> /dev/null';
```
megahirt, line 97: @longrunningprocess is right that we should employ some method of sanitizing the variable before passing to the shell.  The variable, while not user input, is coming in from outside the function and so we have less control on its contents.

See https://www.php.net/manual/en/function.escapeshellarg.php

```suggestion
            $ffprobeCommand = 'ffprobe -i ' . escapeshellarg($tmpFilePath) . ' -show_entries format=duration -v quiet -of csv="p=0" 2> /dev/null';
```

megahirt: On second thought, since $tmpFilePath is used often enough below, it would make sense to have a new variable $sanitizedTmpFilePath that you first escape, and then reference in each place below.

src/angular-app/bellows/shared/audio-recorder/audio-recorder.component.ts
```diff
@@ -1,163 +1,126 @@
+import { webmFixDuration } from 'webm-fix-duration';
 import * as angular from 'angular';
-import * as lamejs from 'lamejs';
 
-declare var webkitAudioContext: {
-  new(): AudioContext;
-};
+declare var MediaRecorder: any;
```
megahirt, line 4: I'm not sure about this - is this a way to declare the type MediaRecorder?  Maybe there is a type for MediaRecorder that we can download on npm?

src/angular-app/bellows/shared/audio-recorder/audio-recorder.component.ts
```diff
@@ -1,163 +1,126 @@
...
-      processor.addEventListener('audioprocess', handleAudioData);
-
-      mp3Encoder.onMp3Blob(blob => {
-        this.blob = blob;
-        this.audioSrc = URL.createObjectURL(blob);
+
+      this.mediaRecorder.addEventListener('dataavailable', async (e: { data: any; }) => {
+        this.chunks.push(e.data);
+        var roughBlob = new Blob(this.chunks, {type: 'audio/webm; codecs=opus'});
+        //2022-09
```
megahirt, line 47: nit: this date comment can come on the same line as the next comment

src/angular-app/bellows/shared/audio-recorder/audio-recorder.component.ts
```diff
@@ -1,163 +1,126 @@
...
-  }
 
-  private stopRecording() {
-    if (this.interval) this.$interval.cancel(this.interval);
-    if (this.stopMediaStream) this.stopMediaStream();
-  }
+      console.error(err);
 
-}
+      });//END getUserMedia call
```
megahirt, line 76: nit: space between semi colon.  I think we need Prettier installed to help with this.  Can you install the Prettier VSCode extension?

src/angular-app/bellows/shared/sound-player.component.html
```diff
@@ -6,5 +6,5 @@
         {{$ctrl.currentTime()}} / {{$ctrl.duration()}}
     </span>
     <!--suppress HtmlFormInputWithoutLabel -->
-    <input class="seek-slider" type="range" min="0" max="{{$ctrl.durationInSeconds()}}" value="0">
+    <input class="seek-slider" type="range" min="0" max="{{$ctrl.durationInSeconds()}}" step="0.0000000000000001" value="0">
```
megahirt, line 9: An HTML comment about the "step" hack would be good here.

src/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php
```diff
@@ -66,23 +67,52 @@ public static function uploadAudioFile($projectId, $mediaType, $tmpFilePath)
...
+            $audioDuration = floatval(shell_exec($ffprobeCommand));
+
+            // Convert to .wav if the result will be less than 1 MB at 165Kb/s (recording is shorter than 6 seconds)
+            // and .mp3 otherwise (recording is longer than 6 seconds)
+            if($audioDuration < 6){
+                shell_exec('ffmpeg -i ' . $tmpFilePath . ' -b:a 165K -maxrate 165K -bufsize 80K ' . trim($tmpFilePath, '.webm') . '.wav 2> /dev/null');
+                shell_exec('mv ' . $tmpFilePath . ' ' . trim($tmpFilePath, '.webm') . '.wav 2> /dev/null');
```
megahirt, line 104: I realize this wasn't in the spec, but I can see real value in keeping the .webm file around alongside the wav and mp3 files instead of getting rid of it, even if we don't reference it.  So I would like to see lines like:
- convert webm to flex file (wav/mp3)
- move webm and flex file to assets folder

megahirt (changes requested):
Great job!!!  I am very pleased to see that you were able to overcome the challenges in the angular component that weren't working, and ended up getting it all working.

I'll approve this PR when you:
- sanitize the shell arg
- keep around the webm file as well, even though we don't point to it directly

Future TODO items not in this PR
- write playwright tests that exercise the audio component.  We don't have previous tests anyway, so you're breaking new ground here
- handle conversion of uploaded files to wav/mp3 for FLEx compatibility

src/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php
longrunningprocess: ok, thanks!

src/angular-app/bellows/shared/sound-player.component.html
longrunningprocess: do me a favor and drop a comment next to it then explaining the "why" at least for posterity, thanks for being so ingenious! 🧠

src/angular-app/bellows/shared/sound-player.component.ts
longrunningprocess: I'm a big fan of no conditionals so 🎉 !!!

src/angular-app/bellows/shared/sound-player.component.html
longrunningprocess: :-D great minds!

src/angular-app/bellows/shared/audio-recorder/audio-recorder.component.ts
laineyhm: Thanks. Installed Prettier and made that change. I'm also taking out a couple unneeded comments that I see I left in this file.

src/angular-app/bellows/shared/audio-recorder/audio-recorder.component.ts
laineyhm: Yes, Besides the fact that I two mediaRecorder vars (just removed this one, which was not used), I just installed the @types/dom-mediacapture-record devDependency to add the MediaRecorder type.

src/angular-app/bellows/shared/audio-recorder/audio-recorder.component.ts
laineyhm: Thanks, now edited

src/angular-app/bellows/shared/sound-player.component.html
laineyhm: Yes, just added

src/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php
laineyhm: Got it. Working on this. Currently, our uploadAudioFile() function in LexUploadCommands.php uploads one file at a time. I am working on having it upload both.

laineyhm: Code has been updated to keep both the original, uploaded/recorded format, as well as the FLEx-compatible format, in case originals might be useful to have at a later date.
image

laineyhm: Support for FLAC and OGG uploads has also been added, addressing Issue #1399

src/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php
laineyhm: Got it. This is in the latest commits.

src/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php
```diff
@@ -54,8 +55,8 @@ public static function uploadAudioFile($projectId, $mediaType, $tmpFilePath)
             "application/octet-stream",
 
             // allow m4a audio uploads, which curiously has a mime type of video/mp4
+            // allow Google Chrome to handle MediaRecorder recordings as video/webm MimeType //LHM
```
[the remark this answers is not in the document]
laineyhm: You got it

src/angular-app/bellows/shared/audio-recorder/audio-recorder.component.ts
```diff
@@ -1,163 +1,120 @@
...
-  new(): AudioContext;
-};
 
 export class AudioRecorderController implements angular.IController {
 
   static $inject = ['$interval', '$scope'];
-  constructor(private $interval: angular.IIntervalService, private $scope: angular.IScope) {}
 
+  mediaRecorder: MediaRecorder;
+  chunks: any = [];
```
megahirt, line 9: I am uncertain if this is the right type, but it might work

```suggestion
  chunks: string[] = [];
```

src/angular-app/bellows/shared/audio-recorder/audio-recorder.component.ts
```diff
@@ -1,163 +1,120 @@
...
   isRecording = false;
   hasRecorded = false;
+  recordingStartTime: Date;
   audioSrc: string;
-  blob: Blob;
+  blob: any;
```
megahirt, line 14: If possible, I'd like to have the right type here as well.

src/angular-app/languageforge/lexicon/editor/field/dc-audio.component.ts
```diff
@@ -112,6 +113,7 @@ export class FieldAudioController implements angular.IController {
           const isUploadSuccess = response.data.result;
           if (isUploadSuccess) {
             this.dcFilename = response.data.data.fileName;
+            console.log("uploadAudio:filename " + this.dcFilename);
```
megahirt, line 116:
```suggestion
```

src/angular-app/languageforge/lexicon/editor/field/dc-audio.component.ts
```diff
@@ -145,7 +147,8 @@ export class FieldAudioController implements angular.IController {
   audioRecorderCallback = (blob: Blob) => {
     if (blob) {
       const date = new Date();
-      const fileName = 'recording_' + format(addMinutes(date, date.getTimezoneOffset()), 'yyyy_MM_dd_HH_mm_ss') + '.mp3';
+      const fileName = 'recording_' + format(addMinutes(date, date.getTimezoneOffset()), 'yyyy_MM_dd_HH_mm_ss') + '.webm';
+      console.log("audioRecorderCallback:filename " + fileName);
```
megahirt, line 151:
```suggestion
```

megahirt (changes requested):
a few more very minor changes - looking good!

src/angular-app/bellows/shared/audio-recorder/audio-recorder.component.ts
laineyhm: That works!

src/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php
```diff
@@ -66,27 +69,70 @@ public static function uploadAudioFile($projectId, $mediaType, $tmpFilePath)
...
+                $ffprobeCommand = `ffprobe -i $sanitizedTmpFilePath -show_entries format=duration -v quiet -of csv="p=0" 2> /dev/null`;
+                $audioDuration = floatval($ffprobeCommand);
+
+                // Convert to .wav if the result will be less than 1 MB at 165Kb/s (recording is shorter than 6 seconds)
+                // and .mp3 otherwise (recording is longer than 6 seconds)
+                $extensionlessTmpFilePath = substr($sanitizedTmpFilePath, 0, strrpos($sanitizedTmpFilePath, strtolower($fileExt)));
+                $extensionlessFileName = substr($fileName, 0, strrpos($fileName, strtolower($fileExt)));
+
+                if($audioDuration < 6){
+                    `ffmpeg -i $tmpFilePath -b:a 165K -maxrate 165K -bufsize 80K $extensionlessTmpFilePath.wav 2> /dev/null')`;
```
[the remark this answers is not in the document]
laineyhm: This and the next few suggestions are changed in the latest commit

src/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php
megahirt: awesome!!!

megahirt:
@laineyhm we are going to need some "how to test this PR" instructions in the PR description so others of us can test this functionality appropriately.  A few scenarios include uploading a variety of formats (webm, m4a, flac, ogg, etc) and see that they get converted.  Of course also in-browser recording, especially on mobile.

megahirt (approved):
Woohoo approved!  This PR has it all: Typescript, PHP, MediaRecorder, FFMPEG :)

laineyhm:
> 

Description updated with testing instructions

LHayashi:
Thanks for getting this working so that recording can take place in Language Forge! That said, it does pain me to store recordings as compressed mp3s which doesn't meet the standard expectation for language documentation. It'd be nice to see that 1MB limit increased : )

longrunningprocess:
# staging tests

Used `chris-test-sr1`

## Chrome on macOS

### < 6s with playback and save ✅ 

Playback was great and files were there.
image

### > 6s with playback and save ✅ 

Playback was great and files were there.
image

megahirt:
> Thanks for getting this working so that recording can take place in Language Forge! That said, it does pain me to store recordings as compressed mp3s which doesn't meet the standard expectation for language documentation. It'd be nice to see that 1MB limit increased : )

I agreed 100% @LHayashi .  As it stands, FLEx only supports wav and mp3.  Up until now, LF only recorded in mp3 but you could upload wav or m4a.  With this change, we support a bunch more formats AND for the first time encode as wav or mp3 (for FLEx compatibility) based upon recording length.

The main problem as you know is the 1MB limit, which is a chorus limit, not a LF limit.  But this is something we need to push for a resolution.

megahirt:
c.f. https://github.com/sillsdev/chorus/issues/277

megahirt:
@laineyhm see the comments by @rmunn in https://github.com/sillsdev/chorus/issues/277 that talk about the exact length of the wav file to fit within the size limit. Robin comes up with 5.6 seconds whereas you came up with 6 seconds.  Do you want to double-check whether a 6 second wav file fits within 1MB?  I acknowledge the "line" for 1MB is not clear here.  I would suggest that we go with Robin's assessment as it it's lower.  One thing would be to check by uploading a 5.9 second wav file to see whether it can successfully be send/received.

laineyhm:
image
When converting to .wav I restricted the output bitrate to a max 165K bits/s. 165,000 bits/s * 6 s gives 990,000 bits, which I thought was just under a MB. But, alas, a byte is 8 bits, so a MB is actually a lot bigger. The other piece is that my max output bitrate was too low; I must have mixed something up. It should be closer to 1400KB/s. 
8 million bits divided by 1400 KB/s gives ~5.6 s. Sorry for the confusion! Should I open a PR with this fix ?

laineyhm:
So far in S/R with FLEx, the audio files are coming through into FLEx, but the playback is not audible. Troubleshooting that.

megahirt:
@laineyhm and I paired and discovered what we think to be the issue regarding wav files not coming through on FLEx.  Also an adjustment to the bitrate per the above comment will be in a PR now in progress.

PR #1479, opened 2022-09-26
Instructions for mobile device testing on a branch

### LINKED ISSUE #1478: chore: developer instructions for mobile device testing on a local branch
We need clear developer instructions for how to do mobile device testing for local development code on a branch using `ngrok`.  There are a number of tunnel-to-your-machine solutions out there, ngrok being just one of them.

Instructions should probably go as a separate section at the bottom of https://github.com/sillsdev/web-languageforge/blob/develop/docs/DEVELOPER.md

laineyhm:
## Description

We can use ngrok to create a URL that accesses what's in localhost from another device. This allows us to test local branches' functionality on mobile devices. This PR adds instructions about how to do this to Developer.md.

Also, we disable HTTPS redirects in Caddy to facilitate easy ngrok usage

Fixes #1478

### Type of Change

-Expanding developer instructions

## Testing on your branch

N/A

## Checklist

- [N/A ] I have commented my code, particularly in hard-to-understand areas
- [N/A] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/docker/ssl/Caddyfile
+++ b/docker/ssl/Caddyfile
@@ -1,7 +1,7 @@
 # https://caddyserver.com/docs/caddyfile
 {
 	#debug
-	#auto_https disable_redirects
+	auto_https disable_redirects
 }
 
 localhost {
--- a/docs/DEVELOPER.md
+++ b/docs/DEVELOPER.md
@@ -6,6 +6,7 @@ Welcome!  We're glad that you are interested in helping develop Language Forge.
     - [Supported Development Environments](#supported-development-environments)
     - [Project Setup](#project-setup)
     - [Running the App Locally](#running-the-app-locally)
+    - [Mobile Device Testing on a Branch](#mobile-device-testing-on-a-branch)
   - [Tests](#tests)
     - [Running Playwright E2E Tests](#running-playwright-e2e-tests)
     - [Running Protractor E2E Tests](#running-protractor-e2e-tests)
@@ -54,6 +55,35 @@ On Windows, the project should be opened with the [Remote - WSL](https://marketp
 1. Use `admin` and `password` to login
 
 Note: The application is accessible via HTTP or HTTPS.  HTTPS is required for service-worker functionality.
+
+### Mobile device testing on a branch ###
+
+Sometimes there may be a need to hit the locally running app from a device other than the machine the app is running on.  In order to do that, you'll need to do the following:
+### If your machine's firewall is already configured for external access e.g. you use Docker Desktop ###
+
+1. Figure out your local ip address
+1. Access the app via http at that address
+
+On a Mac for example:
+```
+ifconfig | grep broadcast
+	inet 192.168.161.99 netmask 0xfffffc00 broadcast 192.168.163.255
+```
+
+then hit `http://192.168.161.99` from your phone or other device on the same network.
+
+NOTE: disabling cache on your device may not be trivial, you'll either need to wipe the site settings on your device's browser or you'll need to do it via USB debugging.
+
+### If your machine's firewall is not configured for external port 80/443 access, you can use ngrok ###
+
+[Here](https://gist.github.com/SalahHamza/799cac56b8c2cd20e6bfeb8886f18455) are instructions for installing ngrok on WSL (Linux Subsystem for Windows).
+[Here](https://ngrok.com/download) are instructions for installing ngrok on Windows, Mac OS, Linux, or Docker.
+
+Once ngrok is installed, run:
+`./ngrok http http://localhost`
+in a bash terminal. The same command with https://localhost may not work, so be careful to try http://localhost in particular.
+
+ngrok will return two URLs, one http and one https, that contain what is being served in localhost. Test on another device using one or both of these URLs.
 
 ## Tests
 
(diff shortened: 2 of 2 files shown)


megahirt:
I would like to get @longrunningprocess review before this is merged.

docker/ssl/Caddyfile
```diff
@@ -1,7 +1,7 @@
 # https://caddyserver.com/docs/caddyfile
 {
 	#debug
-	#auto_https disable_redirects
+	auto_https disable_redirects
```
longrunningprocess, line 4: this will disabled the service worker locally... I'd like to think through this a little more together before we move forward.

docs/DEVELOPER.md
```diff
@@ -186,3 +186,17 @@ A [tutorial on YouTube is available showing how to use XDebug and VSCode](https:
 Additional considerations:
 
 If you encounter errors such as VSCode cannot find a file in the path "vendor", these source files are not available to VSCode as they are running inside Docker.  If you want to debug vendor libraries (not required), you will have to use Composer to download dependencies and put them in your source tree.
+
```
longrunningprocess, line 189: I'm just curious of the instructions on line 23 stopped working?  I don't know if ngrok is going to be a good way to go about this or not.

longrunningprocess (changes requested):
I'm not quite on board with this yet, I need to better understand.  Maybe we can get together and discuss "in-person" first.

docker/ssl/Caddyfile
megahirt: does it really disable the service worker?

I would expect the service worker to fail if the application is accessed through HTTP, which is what this change allows.  If the application is accessed via HTTPS e.g. https://localhost I would expect the service worker to work normally.

This change simply disables auto_redirect allowing the dev to choose whether they want http or https.  That's my read of what this setting does in any case.

docs/DEVELOPER.md
megahirt: So, testing here in the office demonstrates that this doesn't work out of the box.  My understanding is that a web server in docker running on localhost should not automatically be exposed outside of the dev machine unless the firewall had a port 80/443 hole explicitly poked in it.  On Windows I would imagine you have to create a Windows firewall rule to allow port 80/443 in from the outside.  In addition, I'm not sure if our office network is blocking port 80/443 between machines (unlikely) but I wasn't able to ask Tim E about it yet.

docker/ssl/Caddyfile
longrunningprocess: yes, you're right, I'm getting things mixed up, this is probably ok.  We should just review the docs and make sure we're calling out the idea of being able to use both http and https and when to use each.

docs/DEVELOPER.md
longrunningprocess: I just followed the original instructions for accessing the app from a device an external device on the same network and it worked fine.  Let's meet some time and discuss the goal here and make sure I'm following everything that's being tried.

megahirt: So when you say it worked for you on your original device, are you saying you tested from your MacBook or do you have another machine you can spin up a dev environment on?  I am wondering if your machine is already "special" and has an open firewall whereas by default other machines do not.

I'll test at home as you asked for

longrunningprocess: Yes, I spun things up on my MacBook and then accessed the so from my phone via my IP address

megahirt: I can confirm that this works on my Macbook Pro using Billy's instructions.  The reason this works, as far as I can tell, is because I am using Docker Desktop for Mac which has communicated with my Mac's firewall to poke the appropriate hole in the machine's firewall.  On both my and Lainey's machine, we are using WSL without Docker Desktop, which means that docker engine in ubuntu cannot automatically communicate with Windows Firewall to open the appropriate port automatically.

Some conclusions:
- we should give Docker Desktop for Windows and Linux another try to see if we can get away from using docker engine in Ubuntu.  This brings us closer to the same instructions for devs across platforms.
- we should put the ngrok instructions next to Billy's device testing section, as they are really two different ways to achieve the same thing.  Ngrok is really more for "worldwide access", which in theory would be nice to have Alex test out a prototype on my machine across the world but in practice we haven't needed it.

megahirt: @longrunningprocess I have combined the two device testing instructions into a single place at the bottom of the DEVELOPER readme.  IMHO both methods are valid and useful, and I'd like to keep them.  Now that I think we have a handle on why it wasn't working for @laineyhm or I on Windows, I'm satisfied with merging this.

docker/ssl/Caddyfile
megahirt: I added a [note that clarifies that both HTTP and HTTPS are available](https://github.com/sillsdev/web-languageforge/pull/1479/commits/cc01515e06e4b19c96236f6978de27ae079ecdef), but HTTPS is required for service-workers.

longrunningprocess: thank you!

docs/DEVELOPER.md
longrunningprocess: seems like a good idea for those local situations where using Desktop isn't feasible or for an extended use case like others accessing your local machine.

docker/docker-compose.yml
```diff
@@ -337,6 +337,7 @@ services:
       dockerfile: docker/test-php/Dockerfile
     image: test-php
     container_name: test-php
+    platform: linux/amd64
```
longrunningprocess, line 340: so, is this necessary again now?

longrunningprocess (commented):
I really like those updates to the docs, thanks @megahirt , just one final question about the necessity of that `platform` line

docker/docker-compose.yml
megahirt: thanks for catching this!  There are several unintended commits in this branch that I will rebase and delete, and then this should go away

megahirt (approved):
This is ready to merge

PR #1482, opened 2022-09-26
Fix flaky test by waiting for specific URL

### LINKED ISSUE #1481: bug: Flaky E2E test: "Redirects to projects page if already logged in"
**Describe the bug**
"Redirects to projects page if already logged in" appears to be flaky as identified here: https://github.com/sillsdev/web-languageforge/actions/runs/3126152835

Preumably `waitForNavigation({waitUntil: "networkidle"})` sometimes resolves too quickly, while still on the initially requested URL and before the redirect has happened.

**Expected behavior**
The test should always succeed.

myieye:
## Description

By using playwright's built tools for waiting for the URL we want, we can wait for exactly what we want to verify rather than wait for the network to be idle.

Fixes #1481 

### Type of Change

### DIFF
--- a/test/e2e/signup.spec.ts
+++ b/test/e2e/signup.spec.ts
@@ -117,16 +117,16 @@ test.describe('E2E Signup app', () => {
     await signupPage.signupButton.click();
 
     // Verify new user logged in and redirected to projects page
-    await signupPage.page.waitForNavigation({waitUntil: "networkidle"});
+    await signupPage.page.waitForURL("**/app/projects");
     expect(signupPage.page.url()).toContain('/app/projects');
   });
 
   test('Redirects to projects page if already logged in', async ({ memberTab }) => {
-    const signupPage = new SignupPage(memberTab);
+    const signupPageMember = new SignupPage(memberTab);
     await Promise.all([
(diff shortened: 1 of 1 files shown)


megahirt (approved):
YES!!!  This looks like just what we need.

@myieye How is this method different from [waitForNavigation({ url: ProjectsPage.url })](https://github.com/sillsdev/web-languageforge/blob/develop/test/e2e/project- ?settings.spec.ts#L85)

megahirt:
No idea why php unit tests are failing with
```
service.platform should be part of the service.build.platforms: "linux/amd64"
```

megahirt:
We are currently experiencing an issue with unit-tests failing due to a recent change in the GHA agent environment. This PR does not affect any PHP code, so I am overriding the tests and merging.

myieye:
> @myieye How is this method different from [waitForNavigation({ url: ProjectsPage.url })](https://github.com/sillsdev/web-languageforge/blob/develop/test/e2e/project- ?settings.spec.ts#L85)

@megahirt I believe those are equivalent. Theoretically, `waitForNavigation` could take additional options.

megahirt:
> @megahirt I believe those are equivalent. Theoretically, `waitForNavigation` could take additional options.

OK great, thanks for that!

PR #1497, opened 2022-09-30
Correct the naming and storage of FLEx-compatible audio files

### LINKED ISSUE #1496: bug: FLEx-compatible audio files are being renamed and stored incorrectly
**Describe the bug**
The code in LexUploadCommands for uploadFile(), which converts audio files into .wav or .mp3 format for send/receive with FLEx as well, is inadvertently rewriting the FLEx-compatible files so that they actually contain media of the original recording type. Thus, they are not playable in FLEx.

**To Reproduce**

1. Record/upload and save something on qa.languageforge.org in a project that is linked to a FLEx project.
2. Do a Send/Receive with FLEx
3. Find that recording in FLEx and press play.
4. Notice that while there is a file there, it won't play back. This file has the right extension in the filename, but is actually of a different type.

**Expected behavior**
Audio should come into FLEx as MP3 or WAV and be playable.

 - OS: Windows 10
 - Browser: Chrome

laineyhm:
Fixes #1496

## Description

This PR corrects code in LexUploadCommands.php in uploadFile(), which converts audio files from Language Forge into WAV or MP3 format for send/receive with FLEx if needed. Previously, part of this function was inadvertently rewriting the FLEx-compatible files so that they actually contained media of the original (not compatible) recording type. Thus, they were not playable in FLEx. 

The code has been corrected and we will confirm that the files are now playable when brought into FLEx.

### Type of Change

## Checklist

- [  ] I have added tests that prove my fix is effective or that my feature works

## How to test

- [ ] On qa.languageforge.org, open a project that is linked to FLEx and upload or record some audio of non-MP3, non-WAV type. Save it, and do a Send/Receive. Open FLEx and go to the audio field where your recording should be. Verify that the audio is there, and play it back. Make sure it's audible and sounds like the recording that's in Language Forge.
- [ ] Repeat the above step with a short recording (<5.6 s), a long recording (>5.6 s), and uploaded audio files of M4a, FLAC, and OGG types. Here is a folder of test audio files: 
[Audio Test Upload Files.zip](https://github.com/sillsdev/web-languageforge/files/9681369/Audio.Test.Upload.Files.zip)

### DIFF
--- a/src/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php
+++ b/src/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php
@@ -105,8 +105,7 @@ class LexUploadCommands
 
             if (strcmp(strtolower($fileExt), ".mp3") !== 0 && strcmp(strtolower($fileExt), ".wav") !== 0) {
                 //First, find the duration of the file
-                $sanitizedTmpFilePath = escapeshellarg($tmpFilePath);
-                $ffprobeCommand = `ffprobe -i $sanitizedTmpFilePath -show_entries format=duration -v quiet -of csv="p=0" 2> /dev/null`;
+                $ffprobeCommand = `ffprobe -i $tmpFilePath -show_entries format=duration -v quiet -of csv="p=0" 2> /dev/null`;
                 $audioDuration = floatval($ffprobeCommand);
 
                 // Convert to .wav if the result will be less than 1 MB (recording is shorter than 5.6 seconds)
@@ -123,7 +122,7 @@ class LexUploadCommands
 
                 //unlink the original file as well, now that we've both stored it and made the converted copy
                 @unlink($tmpFilePath);
-                
+
             }
 
             // construct server response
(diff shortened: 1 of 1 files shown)


src/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php
```diff
@@ -109,29 +109,21 @@ public static function uploadAudioFile($projectId, $mediaType, $tmpFilePath)
                 $ffprobeCommand = `ffprobe -i $sanitizedTmpFilePath -show_entries format=duration -v quiet -of csv="p=0" 2> /dev/null`;
```
megahirt, line 109: Per our discussion in person, we decided upon inspection of the variable that $tmpFilePath does not contain user input and is given to us by the calling method in the controller with a unique id.  So there is no need to sanitize the variable.  Thought I would note this for @longrunningprocess since he and I had flagged it earlier.

Let's remove the sanitize parts all together to make this simpler.

megahirt (changes requested):
almost there!  Just want to remove the sanitize parts per my discussion.

src/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php
laineyhm: The sanitized variable has been removed

megahirt (approved):
Looks great!  Let's merge and test in FLEx now :)

PR #1504, opened 2022-10-04
On delete, remove all stored versions of the target media file

### LINKED ISSUE #1491: bug: Deleting audio files only removes the converted copy
**Describe the bug**
When we record or upload audio, we save the original file and also save it as a FLEx-compatible file, of MP3 or WAV type. When we delete, we need to delete both files, not just one. Currently, only the converted file is being deleted.

**To Reproduce**
1. Open the lf-app/var/www/assets/lexicon/[project-name]/audio Docker container.
2. In Language Forge, (versions after pull request 1474), upload or record audio for an entry (don't upload an MP3 or WAV file for this). Notice two files for it saved in the container.
3. In Language Forge, delete this audio from the entry.
4. Notice that only the converted file was deleted from the container, and the original file is still there.

**Expected behavior**
Deleting audio from an entry/sense/example will delete all copies of that audio from the audio assets folder.

**Screenshots**
image
The top file got left behind when its audio was 'deleted' from Language Forge. The converted .wav file was deleted, but the recorded .webm file was not.

 - OS: Windows 10, Mac OS 12.6
 - Browser Chrome 105

**Additional context**
See [PR 1474](https://github.com/sillsdev/web-languageforge/pull/1474): staging tests

laineyhm:
Fixes #1491 

## Description

At times we are storing multiple versions of an audio file -- the original and the FLEx-compatible version. When a user deletes the audio for an entry, all the versions of the target delete file will now be removed, not just the specific version the entry points to. The code changed handles pictures as well as audio files, although pictures are not currently being converted or stored in different formats.

### Type of Change

## Checklist

- [x ] I have added tests that prove my fix is effective or that my feature works

## How to test

- [ ] Record something on Language Forge. Save it. Look in the Docker container; see two files, one webm and one converted. Then delete this audio, from the Language Forge site. Look in the Docker container; both files should be gone.
- [ ] Upload some audio of MP3 or WAV type and check to see that even when the audio is not stored in multiple versions and there is only one file, delete works as expected.
- [ ] Upload an image and delete it. Verify that this works as expected.

### DIFF
--- a/src/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php
+++ b/src/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php
@@ -279,7 +279,7 @@ class LexUploadCommands
                 $data->errorMessage = $errorMsg;
                 return $response;
         }
-        $filePath = $folderPath . '/' . $fileName;
+        $filePath = $folderPath . '/' . $fileName; //Need to get both the original and potential converted file here 
         if (file_exists($filePath) and ! is_dir($filePath)) {
             if (unlink($filePath)) {
                 $data = new MediaResult();
--- a/src/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php
+++ b/src/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php
@@ -279,20 +279,37 @@ class LexUploadCommands
                 $data->errorMessage = $errorMsg;
                 return $response;
         }
-        $filePath = $folderPath . '/' . $fileName; //Need to get both the original and potential converted file here 
+
+        //Path to the specific file the entry points to
+        $filePath = "$folderPath/$fileName";
+        //Put any other stored versions of the file (e.g. the same file saved in other formats) in an array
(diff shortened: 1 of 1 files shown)


src/Api/Model/Languageforge/Lexicon/Command/LexUploadCommands.php
```diff
@@ -279,20 +279,37 @@ public static function deleteMediaFile($projectId, $mediaType, $fileName) {
...
+        $versionsOfTheSameFile = glob("$folderPath/$fileNameWithoutExt.*");
+
         if (file_exists($filePath) and ! is_dir($filePath)) {
+
+            //Delete the file the entry points to and create the server response
             if (unlink($filePath)) {
                 $data = new MediaResult();
                 $data->path = $folderPath;
-                $data->fileName = $fileName;
+                $data->fileName = $filePath;
```
megahirt, line 295: I am trying to understand why this is changed to $filePath, but I don't see it.  Can you clarify?

laineyhm: That was my mistake. The latest commit restores the original use of $fileName.

megahirt (approved):
LGTM You've got great momentum on this audio recording story!

PR #1514, opened 2022-10-05
Apply audio durations to scope immediately upon their availability

### LINKED ISSUE #1512: bug: audio length in playback does not reliably display on page load or when switching entries
**Describe the bug**
On MS Edge on MacOS.  I suspect this bug occurs on other browsers/platforms, but need someone else to repro on Chrome on Windows, etc.

The playback widget displays the audio duration "most of the time", but sometimes (under what conditions I'm not sure) it does not display.

Here is a video demo of the bug:

https://user-images.githubusercontent.com/3444521/193802599-95f225ad-c506-43ce-8c02-620cd7802f71.mp4

**To Reproduce**
1. Open a project and entry in Language Forge that contains an audio recording.  It helps if you have several entries to switch between that each contain an audio recording in the entry
2. Make a project like above if you don't have one
3. With an entry selected, refresh the page.  Try refreshing the page multiple times.  On occasion, the audio elapsed/duration does not display
4. Alternatively, click between entries.  In many cases, the audio time elapsed/duration will display after loading for a second.  Sometimes the audio elapsed/duration will not display.

**Expected behavior**
For each field that contains audio, the elapsed time / duration should always display in the audio control.

**Screenshots**
Proper display of audio duration:
image

Audio duration missing:
image

 - OS: MacOS
 - Browser: Edge

laineyhm:
Fixes #1512 

## Description

With our AngularJS bindings, the duration of an audio file is not displayed unless it has been loaded. Previously, the DOM was not being updated immediately upon the duration becoming available. This PR tells the $scope to apply its changes to the DOM once the "durationchange" event has occurred on the HTML audio element.

Because it takes a small amount of time for the duration to become available, there is still visible movement; there is a split second from when the entry is clicked on and the duration is not there, to when the duration appears.

### Type of Change

- UI change

## Screenshots

Before, I was able to reproduce Chris's bug video in issue #1512, for me on Windows 10 with Google Chrome. Now I am seeing this:

https://user-images.githubusercontent.com/56163492/194026070-6acebb2e-ce3f-45e3-968c-6f6699559a8d.mp4

A potential concern of mine is that the quick flash when the duration appears might be considered unpleasant to look at. But, I am not sure I can make it appear with zero delay, since our code already says the duration should appear if it's available, the instant you had clicked it.

## Checklist

## How to test

- [ ] Load the page. See if an entry's audio duration appears right after you load the page.
- [ ] Click on some entries with audio. See if the audio duration appears right after you click on it.
- [ ] Upload an audio file. See if the audio duration displays right after the upload completes.
- [ ] Record an audio segment. See if the audio duration displays right after the upload completes.

### DIFF
--- a/src/angular-app/bellows/shared/sound-player.component.ts
+++ b/src/angular-app/bellows/shared/sound-player.component.ts
@@ -16,8 +16,12 @@ export class SoundController implements angular.IController {
 
   $onInit(): void {
 
-
     this.slider = this.$element.find('.seek-slider').get(0) as HTMLInputElement;
+
+    //So that duration appears immediately once it is available
+    this.audioElement.addEventListener('durationchange', () => {
+      this.$scope.$apply();
+    });
 
     this.audioElement.addEventListener('ended', () => {
       this.$scope.$apply(() => {


megahirt (approved):
Wow!  Nailed it!  Well done.

PR #1519, opened 2022-10-07
upgrade mongo to 4.2

### LINKED ISSUE #1173: Upgrade Mongo to 4.2 or later
Mongo 4.0 introduced aggregation pipelines, and Mongo 4.2 introduced the ability to pass an aggregation pipeline to a Mongo `update()` call. This led to the Mongo team declining several feature requests like https://jira.mongodb.org/browse/SERVER-4798 (pop multiple items from an array at once), because in Mongo 4.2 you can do that with an aggregation pipeline.

Some of the Language Forge code (such as the code that makes entry updates into deltas, instead of replacing the entire entry at once) will occasionally need to remove multiple items from an array (for example, if someone removes two Meanings from an entry before saving the entry), and that's not possible before Mongo 4.2.

Upgrade the Mongo container to at least Mongo 4.2, and probably the latest version of Mongo 4.x so we don't have to pay the cost of a database upgrade as often. (For thoughts on Mongo 5.0, see below). Then we'll be able to use aggregation pipelines in Mongo updates, which would allow some efficiency wins, and in some cases allow updates that wouldn't have been possible to write before. For example, at one point we were considering having a `dirtySR` flag on projects, and the Send/Receive code would decrement the flag if and only if it was greater than 0. That's not possible before Mongo 4.2, but aggregation pipelines make it possible to write that logic into a single atomic Mongo update.

We could stay on Mongo 4.0, and add logic that forces entry saves before adding or removing a meaning or example. That way any Mongo update will contain at most a single array push or pop.

**Additional context**
Mongo 5.0 is worth considering, but it's a bit riskier as it's a major version number and a recent change. However, looking through the Mongo 5.0 release notes, the only thing I see that could cause us issues is that the `mongo` shell command is being deprecated and replaced with `mongosh`, which is only an issue of habits (when connecting to the Mongo DB directly, got to get used to typing `mongosh` instead of `mongo`).

longrunningprocess:
Fixes #1173 

## Description

There is a desire to utilize aggregation pipelines and take advantage of possible perf improvements.

### Type of Change

- infrastructure upgrade

## Screenshots

N/A

## Checklist

## How to test

- [ ] A general smoke test in addition to successful automated tests should be enough to move forward with this one.

### DIFF
--- a/docker/deployment/db-deployment.yaml
+++ b/docker/deployment/db-deployment.yaml
@@ -49,7 +49,7 @@ spec:
     spec:
       containers:
       - name: db
-        image: mongo:4.0
+        image: mongo:4.2
         # https://kubernetes.io/docs/concepts/configuration/manage-resources-containers
         resources:
           requests:
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -193,7 +193,7 @@ services:
     command: sh -c "postconf -e 'default_transport = retry:no outbound email allowed' && /run.sh"
 
   db:
-    image: mongo:4.0
+    image: mongo:4.2
     container_name: lf-db
     ports:
       # exposed this to host for admin tools


longrunningprocess:
https://www.mongodb.com/docs/manual/release-notes/4.2-upgrade-standalone

# smoke test ✅ 
Automated tests ran and some basic local testing did not uncover any issues.

# driver compat ✅ 
We're currently running `driver: { name: "mongoc / ext-mongodb:PHP ", version: "1.22.1 / 1.14.1 " }` and according to https://www.mongodb.com/docs/drivers/php/#compatibility-table-legend, we should be good.  With that info as well as good automated tests and good local smoke testing, this one looks good.

# storage engine compat
- [ ] need to ensure our storage engine is WiredTiger

# api compat
https://www.mongodb.com/docs/manual/release-notes/4.2-compatibility
- [ ] need to ensure we don't use any of these changed API's in our code, e.g., `$group` in particular.

# config
https://www.mongodb.com/docs/manual/release-notes/4.2-upgrade-standalone/#feature-compatibility-version
- [ ] need to ensure this config is correct

megahirt:
> need to ensure our storage engine is WiredTiger

```
chris@MacBook docker % docker exec lf-db mongo --eval "db.serverStatus().storageEngine;"
MongoDB shell version v4.0.28
connecting to: mongodb://127.0.0.1:27017/?gssapiServiceName=mongodb
Implicit session: session { "id" : UUID("ccabba5d-de99-4e88-912f-25f062eb293e") }
MongoDB server version: 4.0.28
{
	"name" : "wiredTiger",
	"supportsCommittedReads" : true,
	"supportsSnapshotReadConcern" : true,
	"readOnly" : false,
	"persistent" : true
}
```

yup

megahirt:
> need to ensure we don't use any of these changed APIs in our code, e.g., $group in particular.

I searched through our code base and could not find any usage of the mongo `$group`

megahirt:
```
chris@MacBook docker % docker exec lf-db mongo --eval "db.adminCommand( { getParameter: 1, featureCompatibilityVersion: 1 } )"
MongoDB shell version v4.0.28
connecting to: mongodb://127.0.0.1:27017/?gssapiServiceName=mongodb
Implicit session: session { "id" : UUID("61b32da1-e5cb-48e2-9508-82d2e64aa1e8") }
MongoDB server version: 4.0.28
{ "featureCompatibilityVersion" : { "version" : "4.0" }, "ok" : 1 }
```
The config is set to 4.0, so we are good there

megahirt (approved):
LGTM

I think the general idea here is to ship 4.2, let it go for a few weeks, and then upgrade to 4.4 a few weeks from now.

laineyhm:
Tested this today. I clicked around on the site, added an entry, and did a Send/Receive. I didn't notice any problems

PR #1532, opened 2022-10-18
Chore: make prettier pre commit executable

myieye:
## Description

The pre-commit script file meant for prettier is not marked as executable. Presumably this is because I was shuffling files between Windows and Linux 🤣 while working on those changes.

image


megahirt (approved):
LGTM

PR #1547, opened 2022-10-20
Rework Makefile to remove npm-ci target

myieye:
## Description

This seems cleaner than having a target for a single command.
`$(MAKE)` is the recommended way rather than simply `make`

This change will result in `make playwright-tests-ci` running `npm ci` and then also `npm install`, but I think that's harmless and is simpler than introducing more make targets or make variables with shell `if`s.


longrunningprocess (approved):
ahead, that's cleaner

Issue #1536, opened 2022-10-19
bug: 500 server error on activity dashboard when accessing Tampuan Flex project


megahirt:
**Describe the bug**
In some cases (I'm still trying to figure out exactly in which cases), switching between projects via the "My Projects" dropdown or via the Projects listing, results in a 500 error with a message similar to "don't have privileges accessing api method"

**To Reproduce**
Steps to reproduce the behavior: Add yourself to the Tampuan Flex project and click the link.
Alternatively visit https://languageforge.org/projects/tpu-flex if you are already a member.

https://languageforge.org/app/lexicon/5a2f5fa88b2147063928aaf3 is the lexicon application link which still works.

longrunningprocess:
# Use cases currently failing

## new entry gets added to two different projects simultaneously

## `500` after deleting a project

## no work-around for Tampuan Flex project

longrunningprocess:
the issue is with the `checkPermissions` call and its assumption that the `projectId` is set... we may consider either finding a different UI treatment that would allow a user to get to the dashboard or we will need to create some awareness of an incoming project code within the legacy authz model, i.e., checkPermissions and/or Sf.php.

PR #1563, opened 2022-10-26
need to ensure the only way to the dashboard is through the editor

longrunningprocess:
Fixes portions of #1536 

## Description

Authorization checks expect the "current" project to be loaded into session state on the server so when users were trying to get to the dashboard without first going to the editor, the "current" project state was out of sync or missing, therefore authz checks were failing.

This PR simply forces navigation to the dashboard to be done through the editor, ensuring the correct project data is loaded into server session.

### Type of Change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Screenshots

N/A

## Checklist

## How to test

Please describe how to test and/or verify your changes. Provide instructions so we can reproduce. Please also provide relevant test data as necessary. These instructions will be used for QA testing below.

- [ ] Ensure project links in the header dropdown, "My Projects" take you into the editor
- [ ] Ensure project links on the "My projects" page take you into the editor
- [ ] Ensure the project link in the breadcrumbs will take you into the dashboard

### DIFF
--- a/src/angular-app/bellows/apps/projects/projects-app.component.html
+++ b/src/angular-app/bellows/apps/projects/projects-app.component.html
@@ -50,8 +50,6 @@
                         <div class="col-md-4">
                             <a data-ng-show="$ctrl.isInProject(project)" data-ng-href="/app/lexicon/{{project.id}}">
                                 <span class="larger-text">{{project.projectName}}</span></a>
-                            <a data-ng-show="$ctrl.isInProject(project)" data-ng-href="/projects/{{project.projectCode}}">
-                                <span>(Dashboard)</span></a>
                             <span data-ng-show="!$ctrl.isInProject(project)" class="larger-text">{{project.projectName}}</span>
                             <small class="text-muted"> {{project.projectCode}}</small>
                         </div>


megahirt:
This doesn't really address the core issue described in #1536.  We will likely run into this issue (wrong project context) as we build out more views in Svelte, so we need to take a step back and engineer something that allows the Svelte code to set the project Id directly.  The AngularJS application doesn't have to do this because the project ID is embedded in the URL of the PHP code that serves up the application.  The svelte application works differently, so we need to add functionality to our API to facilitate this new way.

How about adding a PHP API method called `setProject` whose purpose is to set the project context before calling subsequent API methods.

I think ultimately we want an api where you pass the project id as a parameter (stateless), but that would require re-architecting all the api methods and calling (a big job).  We would design our new rest API in this way, with all the things.

longrunningprocess:
> This doesn't really address the core issue described in #1536. We will likely run into this issue (wrong project context) as we build out more views in Svelte, so we need to take a step back and engineer something that allows the Svelte code to set the project Id directly. The AngularJS application doesn't have to do this because the project ID is embedded in the URL of the PHP code that serves up the application. The svelte application works differently, so we need to add functionality to our API to facilitate this new way.
> 
> How about adding a PHP API method called `setProject` whose purpose is to set the project context before calling subsequent API methods.
> 
> I think ultimately we want an api where you pass the project id as a parameter (stateless), but that would require re-architecting all the api methods and calling (a big job). We would design our new rest API in this way, with all the things.

I can take a look at converting the project code to an id in the Svelte backend before any legacy calls.  I'd suggest we get this fix in ASAP since that's a bug right now and we can circle back with our desired, long-term solution.  What are your thoughts on that approach?

megahirt (approved):
I agree with your suggestion to remove the dashboard link and then circle back around in a separate issue.

megahirt:
@longrunningprocess can you create a new issue to fix dashboard access for a project regardless of the origin?

megahirt:
TESTED:

- The My projects page links go to the editor
- The My projects drop down links go to the editor
- The breadcrumbs link goes to the dashboard

PR #1581, opened 2022-11-07
docker services diagram

hahn-kev:
As part of my on-boarding I decided to make a mermaid diagram, this is a diagram in markdown, it's not an image that I copied in so it'll be easy to keep up to date as things change.

Edit: make sure you view the markdown preview of the file

### DIFF
--- a/docs/DIAGRAM.md
+++ b/docs/DIAGRAM.md
@@ -17,7 +17,7 @@ flowchart TB
     lfmerge -- port 27017 --> db
     lfmerge --> ld-api
 
-    subgraph prodcution
+    subgraph production
         ssl[ssl Caddy \n ports: 80, 443]
         ssl -- port 80 --> next-proxy
 
--- a/docs/DIAGRAM.md
+++ b/docs/DIAGRAM.md
@@ -1,0 +1,83 @@
+Docker services diagram (dev)
+
+dashed line means it connects but the service does not depend on the next container
+
+```mermaid
+%%{init: {'flowchart': {'curve': 'basis'}}}%%
+flowchart TB
+    ui-builder
+    app[app php]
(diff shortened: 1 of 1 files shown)


docs/DIAGRAM.md
```diff
@@ -0,0 +1,83 @@
...
+    app -- port 25 --> mail
+    app -- port 3000 --> ld-api
+    app -- fs --> lfmerge
+    app -- /data --> ui-builder
+
+    lfmerge[lfmerge C#]
+    lfmerge -- port 27017 --> db
+    lfmerge --> ld-api
+
+    subgraph prodcution
```
longrunningprocess, line 20: typo

longrunningprocess (changes requested):
I LOVE that you did this and you are the perfect person to do something like this as a newcomer, thank you for the initiative.

These diagrams only appear to represent the local dev environment and I'm wondering if we'd like to better represent the differences between our local dev and our operational environments.  I believe it will be more useful to have the operational environments diagrammed and then, in addition, supply the local dev envs.  Hopefully we can chat more in person.

hahn-kev:
> I LOVE that you did this and you are the perfect person to do something like this as a newcomer, thank you for the initiative.
> 
> These diagrams only appear to represent the local dev environment and I'm wondering if we'd like to better represent the differences between our local dev and our operational environments. I believe it will be more useful to have the operational environments diagrammed and then, in addition, supply the local dev envs. Hopefully we can chat more in person.

Thanks, it was a great way to figure out how all the different things were tied together which can be tricky when just looking at the compose file.

yes that's correct, it is just the docker compose file. I'll take a look into the k8s config and build another diagram (probably in another file) to represent that environment.

longrunningprocess (approved):

PR #1584, opened 2022-11-08
improved delete and move UX on entry, meaning and examples

### LINKED ISSUE #1537: feat: Better UI for deleting entries
How to delete an entry from Language Forge is not obvious. A user had to ask how to do that (answer: use the three dots at the top right of the editing area). 

Trash can symbol or red X symbol, visible (not in the three dots) for Entry, Meaning, and Example. These should display on mobile.

What we already have

longrunningprocess:
Fixes #1537 

## Description

This PR changes the UI in accordance with some developer observations of users struggling with how to perform some activities on their dictionary entries.

### Type of Change

- UI change

### Entry 

image

### Meaning

image

image

image

### Example

image

### Tooltips

https://user-images.githubusercontent.com/4412848/200657385-8f6fcc8b-527f-41e7-9bea-0a8c76dec26c.mov

### Mobile

looks good but I do demonstrate the widely-known issue of hover behavior on mobile devices... we can either live with it, consider bringing popper or some other 3rd-party lib in, remove the tooltips entirely or utilize more javascript to isolate hovers to appropriate devices.

https://user-images.githubusercontent.com/4412848/200659475-dff5cb87-8371-4e8a-9d11-3fd6ea6a634c.mov

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works

## How to test

Please describe how to test and/or verify your changes. Provide instructions so we can reproduce. Please also provide relevant test data as necessary. These instructions will be used for QA testing below.

- Add an entry, hover over it ensuring the tooltip content is correct and ensure the entry is actually deleted
- Add a few meanings, hover over the movement and delete icons ensuring the tooltip content is correct and ensure the meaning is actually moved and/or deleted
- Add a few examples, hover over the movement and delete icons ensuring the tooltip content is correct and ensure the example is actually moved and/or deleted

### DIFF
--- a/src/angular-app/languageforge/lexicon/editor/_editor.scss
+++ b/src/angular-app/languageforge/lexicon/editor/_editor.scss
@@ -860,3 +860,37 @@ form.submitted .ng-invalid {
   outline: none;
   border: 3px solid #f00;
 }
+
+// approach derived from some MDN articles and https://dev.to/link2twenty/native-html-tooltips-3od1
+[data-tooltip] {
+  position: relative;
+  cursor: help;
+}
+
+[data-tooltip]::after {
+  opacity: 0;
+  pointer-events: none;
+
+  position: absolute;
+  right: 1rem;
+  top: -0.5rem;
+  z-index: 99;
+
+  content: attr(data-tooltip);
+  padding: 1rem;
+  white-space: nowrap;
+
+  border-radius: 3px;
+  box-shadow: 0 0 5px 2px rgba(100, 100, 100, 0.6);
+  background-color: whitesmoke;
+
+  transform: translateX(0);
+  transition: all 500ms cubic-bezier(0.25, 0.8, 0.25, 1);
+}
+
+[data-tooltip]:hover::after {
+  opacity: 1;
+
+  transform: translateX(-1rem);
+  transition-duration: 750ms;
+}
--- a/src/angular-app/languageforge/lexicon/editor/_editor.scss
+++ b/src/angular-app/languageforge/lexicon/editor/_editor.scss
@@ -860,37 +860,3 @@ form.submitted .ng-invalid {
   outline: none;
   border: 3px solid #f00;
 }
-
-// approach derived from some MDN articles and https://dev.to/link2twenty/native-html-tooltips-3od1
-[data-tooltip] {
-  position: relative;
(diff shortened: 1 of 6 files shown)


longrunningprocess:
I'm moving this back to draft... I'm not happy with the current solution and did not hear back about whether we could simply eliminate it or not.

longrunningprocess:
after discussing at stand-up today, we've decide to eliminate the tooltips altogether.  We will also go with just the icons, no extra text to turn on and off based on the viewport since we believe the icons are pretty universally clear.

megahirt (approved):
Great - thank you for this!

I did look up the title= attribute as a cheap tooltip and MDN says it still works:

https://developer.mozilla.org/en-US/docs/web/html/global_attributes/title

but has "accessibility concerns"

I personally think that adding a title attribute / tooltip on the up down arrows is a cheap win and only adds value for those not using a touch interface.

longrunningprocess:
> Great - thank you for this!
> 
> I did look up the title= attribute as a cheap tooltip and MDN says it still works:
> 
> https://developer.mozilla.org/en-US/docs/web/html/global_attributes/title
> 
> but has "accessibility concerns"
> 
> I personally think that adding a title attribute / tooltip on the up down arrows is a cheap win and only adds value for those not using a touch interface.

Have you tried it?  I don't think it works anymore...

longrunningprocess:
yeah, doesn't work for me in Chrome, works improperly in Safari and works as expected in FireFox and Edge

megahirt:
> yeah, doesn't work for me in Chrome, works improperly in Safari and works as expected in FireFox and Edge

So I tried it just now and works as expected for me in Chrome, Safari and Edge on Mac.  https://www.w3schools.com/TAGS/tryit.asp?filename=tryhtml5_global_title
(apologies for the W3schools link - yuck)

When I suggested `title=` as a cheap win, I was referring to the fact that we have nothing right now, so even if it only works for some users (not mobile/touch), that is better than what we have right now, which is no tooltip, with no immediate plans to change that.

laineyhm:
image

Delete and move icons are readily visible and working as expected to delete entries/meanings/examples and move meanings/examples. 

* In the (delete) chicken page, I recommend that we add a question mark to the end of the question "Are you sure you want to delete the entry ______"

PR #1591, opened 2022-11-14
refactor kubeSyncMongo executable documentation

megahirt:
This is an untested script to facilitate copying prod data to qa - which is basically executable documentation.  We may never use this, but I recently updated this file before I realized that @longrunningprocess would work with LTOps to copy prod data to qa, and that seems to have worked fine.

### DIFF
--- a/docker/deployment/scripts/kubeSyncMongo.sh
+++ b/docker/deployment/scripts/kubeSyncMongo.sh
@@ -1,41 +1,39 @@
 #!/bin/bash
 
 set -e
+set -x
 
-echo Mongo Dump
-ssh mongo-xf mongodump
+PRODCONTEXT=--context='languageforge'
+QACONTEXT=--context='qa-languageforge'
+PRODSERVER=$(kubectl $PRODCONTEXT get pods --selector='app=db' -o name | sed -e s'/pod\///')
+QASERVER=$(kubectl $QACONTEXT get pods --selector='app=db' -o name | sed -e s'/pod\///')
 
-echo Remove Scripture Forge data from dump
-ssh mongo-xf rm -r dump/xforge
+echo "Mongo Dump on production"
+kubectl $PRODCONTEXT exec -c db $PRODSERVER -- bash -c "cd /data/db && mongodump"
 
-echo Create tar file of dump
-ssh mongo-xf tar -czvf mongodump.tgz dump
+echo "Create tar file of dump"
+kubectl $PRODCONTEXT exec -c db $PRODSERVER -- bash -c "cd /data/db && tar -czvf mongodump.tgz dump"
 
-echo Copy tar file to local filesystem
-scp mongo-xf:mongodump.tgz .
+echo "Copy tgz file to local filesystem"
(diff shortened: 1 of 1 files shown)


docker/deployment/scripts/kubeSyncMongo.sh
```diff
@@ -1,41 +1,39 @@
 #!/bin/bash
 
 set -e
+set -x
 
-echo Mongo Dump
-ssh mongo-xf mongodump
+PRODCONTEXT=--context='languageforge'
```
longrunningprocess, line 6: these will require anyone wanting to use this script will have to ensure their context names are exactly the same FYI.

docker/deployment/scripts/kubeSyncMongo.sh
```diff
@@ -1,41 +1,39 @@
...
+QACONTEXT=--context='qa-languageforge'
+PRODSERVER=$(kubectl $PRODCONTEXT get pods --selector='app=db' -o name | sed -e s'/pod\///')
```
longrunningprocess, line 8: it would simpler to simply use `deployment/db` in these.

docker/deployment/scripts/kubeSyncMongo.sh
```diff
@@ -1,41 +1,39 @@
...
+QASERVER=$(kubectl $QACONTEXT get pods --selector='app=db' -o name | sed -e s'/pod\///')
 
-echo Remove Scripture Forge data from dump
-ssh mongo-xf rm -r dump/xforge
+echo "Mongo Dump on production"
+kubectl $PRODCONTEXT exec -c db $PRODSERVER -- bash -c "cd /data/db && mongodump"
```
longrunningprocess, line 12: would you anticipate these dumps ever needing more space than might be available?

docker/deployment/scripts/kubeSyncMongo.sh
```diff
@@ -1,41 +1,39 @@
...
+echo "Copy tgz file to local filesystem"
+kubectl $PRODCONTEXT cp -c db $PRODSERVER:/data/db/mongodump.tgz .
 
-MONGOPOD=$(kubectl get pods --selector='app=db' -o name | sed -e s'/pod\///')
+echo "See how big mongo dump file is"
+ls -lh mongodump.tgz
 
-echo Copying mongodump.tgz into $MONGOPOD/data/db
-kubectl cp -c db mongodump.tgz $MONGOPOD:/data/db
+echo "Copy tgz file to QA"
```
longrunningprocess, line 23: we might need a flag in here to "force" an overwrite in case a previous file by that name already exists

longrunningprocess (commented):
I'd like to see us test this at some point before folding this into our code.  Additionally, should we think about the underlying filesystems as well?

megahirt:
During a conversation with @longrunningprocess we resolved the conversation and concluded this is simply a documentation update that is not tested and we have no intention on running without further testing and customization.  The code is not connected to any part of our application or automated process.  We decided I would just merge this documentation updated myself.

PR #1593, opened 2022-11-14
refactor activity call into consistent convention

### LINKED ISSUE #1576: stateless LFNext calls need to accommodate stateful legacy PHP calls
> This doesn't really address the core issue described in #1536. We will likely run into this issue (wrong project context) as we build out more views in Svelte, so we need to take a step back and engineer something that allows the Svelte code to set the project Id directly. The AngularJS application doesn't have to do this because the project ID is embedded in the URL of the PHP code that serves up the application. The svelte application works differently, so we need to add functionality to our API to facilitate this new way.
> 
> How about adding a PHP API method called `setProject` whose purpose is to set the project context before calling subsequent API methods.
> 
> I think ultimately we want an api where you pass the project id as a parameter (stateless), but that would require re-architecting all the api methods and calling (a big job). We would design our new rest API in this way, with all the things.

I can take a look at converting the project code to an id in the Svelte backend before any legacy calls.  I'd suggest we get this fix in ASAP since that's a bug right now and we can circle back with our desired, long-term solution.  What are your thoughts on that approach?

_Originally posted by @longrunningprocess in https://github.com/sillsdev/web-languageforge/issues/1563#issuecomment-1298709695_

longrunningprocess:
Fixes #1576 

## Description

This PR extends the convention of keeping the stateless calls into the LFNext portions of the app in sync with the stateful requirements of the legacy app.

### Type of Change

## Screenshots

N/A

## Checklist

## How to test

- Simply hitting the project dashboard on a project with activity and ensuring there are no failures is a good test.
- While on the dashboard for a project, ensure the "Show all" works as well
- Also ensure clicking on "work on this project" and coming back to the dashboard retains correct project

### DIFF
--- a/next-app/src/routes/projects/[project_code]/activities/+server.ts
+++ b/next-app/src/routes/projects/[project_code]/activities/+server.ts
@@ -4,23 +4,24 @@ import { sf } from '$lib/fetch/server'
 export async function GET({ params: { project_code }, request: { headers } }) {
 	const cookie = headers.get('cookie')
 
-	const activities = await get_activities({ project_code, cookie })
+	await sf({ name: 'set_project', args: [ project_code ], cookie })
+
+	const activities = await get_activities({ cookie })
 
 	return json(activities)
 }
 
 // src/Api/Model/Shared/Dto/ActivityListDto.php
 // src/Api/Model/Shared/Dto/ActivityListDto.php->ActivityListModel.__construct
-export async function get_activities({ project_code, cookie, start_date, end_date }) {
+export async function get_activities({ cookie, start_date, end_date }) {
 	const args = {
-		name: 'activity_list_dto_for_project',
+		name: 'activity_list_dto_for_current_project',
 		args: [
-			project_code,
 			{
 				startDate: start_date,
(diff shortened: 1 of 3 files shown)


megahirt (approved):
I approve of this change if this is the way that you want to go.  I do think that moving toward a stateless API surface where you pass the project ID in is a good direction to go in (one we've discussed before), so it's a bit of a shame to undo that in this PR.  But you would have to be willing to refactor the activity_list_dto_for_project() to also do the work that set_project() already does in order to get the same effect.

We have multiple goals in this code, and one of them is keeping the code decoupled, which you are doing.  The other would be to move toward a stateless API.  Obviously we have a ways to go but I leave it up to you in determining when the right time is to refactor some of this - if at all.

longrunningprocess:
in my earlier research, the bigger issue is the `checkPermissions` call, that would have to be refactored to get the context right before checking authz.  It's definitely possible but a bit more intrusive to existing calls if we were to change something there.

josephmyers:
Ran through the bullet points on QA. Looks good

PR #1599, opened 2022-11-17
add .dockerignore to prevent vendor files from getting included

hahn-kev:
per the slack conversation files in `src/vendor` could be included from a developers machine into the docker image, this has the potential to break things unexpectedly. We should exclude all files in the vendor folder as those should only come from compose.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,13 @@
+# The .dockerignore file excludes files from the container build process.
+#
+# https://docs.docker.com/engine/reference/builder/#dockerignore-file
+
+# Exclude locally vendored dependencies.
+src/vendor/
+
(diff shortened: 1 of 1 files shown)


longrunningprocess:
devs should not have a local vendor folder at all... I'm not sure how that could be present unless the dev is doing things outside of docker and/or make.  Have you seen environments with a `src/vendor` folder present in their local filesystem?

hahn-kev:
Some IDEs require you to install packages on your local machine in order to provide type/error checking so we should at least allow the vendor folder to exist without breaking the build.

longrunningprocess:
Ok, that makes sense. Thanks for explaining the use case.

longrunningprocess (approved):

PR #1602, opened 2022-11-17
Clarify use of title & description in merge commits and add auto-merge reminder

myieye:
## Description

As far as I can tell, only the title of a PR is used in the PR's commit message.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -22,9 +22,10 @@ Please provide screenshots / animations for any change that involves the UI. Ple
 ## Checklist
 
 - [ ] I have performed a self-review of my own code
-- [ ] I have reviewed the title/description of this PR which will be used as the squashed PR commit message
+- [ ] I have reviewed the title & description of this PR which I will use as the squashed PR commit message
 - [ ] I have commented my code, particularly in hard-to-understand areas
 - [ ] I have added tests that prove my fix is effective or that my feature works
+- [ ] I have enabled auto-merge (optional)
 
 ## How to test
 


longrunningprocess (approved):

megahirt:
Technically you are correct.  For myself, I tend to copy the PR description into the squash commit message, so it's not automatic, but it is generally what I do use in the final commit message.  I would generally encourage others to do the same practice.

myieye:
I see. For reference here's your latest commit:
image

Compared to mine:
image

And here's your snazy commit in VS-Code.
image

.github/pull_request_template.md
```diff
@@ -22,9 +22,10 @@ Please provide screenshots / animations for any change that involves the UI. Ple
 ## Checklist
 
 - [ ] I have performed a self-review of my own code
-- [ ] I have reviewed the title/description of this PR which will be used as the squashed PR commit message
+- [ ] I have reviewed the title & description of this PR which I will use as the squashed PR commit message
 - [ ] I have commented my code, particularly in hard-to-understand areas
 - [ ] I have added tests that prove my fix is effective or that my feature works
+- [ ] I have enabled auto-merge (optional)
```
myieye, line 28: @megahirt this is something we briefly discussed at our team meeting today and it seemed we somewhat agreed that a simple reminder here could be valuable (at least for me 😆).

longrunningprocess (approved):

megahirt:
@myieye some team members prefer prolific committing on their branch, which is totally fine by me.  When it gets squashed into a PR commit, however, it could end up like this [fine piece of work](https://github.com/sillsdev/web-languageforge/commit/825bd1975b43c8a75c93f4b3e5dbc7b85e88b9c4).  At the end of the day, I would like each commit message to be a summary of the work done.  This is usually best represented in the PR description.  If you only have a few commits, and your commits are meaningfully documented, then the default squash message is probably pretty good and no action is required.

PR #1604, opened 2022-11-17
adjust audio input system regex to be more inclusive

### LINKED ISSUE #1601: bug: audio input system detection not working for non-standard language tag
**Describe the bug**
In working today with a team member interested in using Language Forge, I pulled in their project from Language Depot and showed them their dictionary in LF.  What was unexpected was that the audio input system with the audio files were represented as plain text field instead of the audio playback widget.  See screenshot:

image

I was initially confused about this, but then I looked at the language tag for the audio input system.  I note that it does not follow the standard format for an audio input system that LF expects (expecting `Zxxx-lang-x-audio`).  In this project, the audio input system look like` Zxxx-lang-x-other-audio` where `other `is an arbitrary string.

LF should make its pattern matching detection of audio writing systems more flexible to be able to handle this kind of uncommon input system tag, one I've never seen before.

[Elawa200810](https://languageforge.org/app/lexicon/6375e5395be4fc4cbb03d862)

I also want to note that the LFNext dashboard has audio detection built in that appears to work fine, since it correct shows the number of entries which contain audio:
image

**To Reproduce**
Open the project above and note that the audio input system does not have a play/record button, but instead has a plain text field.

**Expected behavior**
The audio play/record button is visible for fields which contain an audio file.

megahirt:
Fixes #1601 

## Description

Per a user's report, their audio input system was not properly detected in LF due to a too-strict regex.  This loosens the regex a bit to allow LF to properly identify the field as an audio input system.

With this change, a tag must have "-Zxxx" and end with "audio"

Before this change, a non-standard input system tag like `lwl-Zxxx-x-minority-audio` is not detected as an audio input system (although FLEx handles it fine).  After this change, this input system tag should work.

Please note that extra careful testing will be necessary for this PR since I cannot verify that my code change actually works.  I only have my phone with me and do not have a dev environment to test locally.

I used regex101 to test my regex and verify it against two strings to make sure they both work:
`lwl-Zxxx-x-audio`
`lwl-Zxxx-x-minority-audio`

### Type of Change

## Screenshots

Refer to the reported bug for screenshots.

## Checklist

## How to test

An input system having the tag lwl-Zxxx-x-minority-audio ought to show up fine as an audio input system with this change.  I have imported the problem project into QA and so we can verify that the change makes a difference there in QA before shipping this fix.

### DIFF
--- a/src/angular-app/bellows/core/utility.service.ts
+++ b/src/angular-app/bellows/core/utility.service.ts
@@ -24,7 +24,7 @@ export class UtilityService {
   }
 
   static isAudio(tag: string): boolean {
-    const tagAudioPattern = /^\w{2,3}-Zxxx-x(-\w{2,3})*-[aA][uU][dD][iI][oO]$/;
+    const tagAudioPattern = /^\w{2,3}-Zxxx-x-.*[aA][uU][dD][iI][oO]$/;
     return tagAudioPattern.test(tag);
   }
 


myieye:
Well this is interesting (at least for me). It parses "minority" out of the name and displays it in the UI. Do we want that?:
image

And it correctly shows up as audio:
image

myieye (approved):

longrunningprocess:
# Testing in staging

✅ 
image

megahirt:
> Well this is interesting (at least for me). It parses "minority" out of the name and displays it in the UI. Do we want that?

I think it is a bit weird, but it's also needed as the "private use area" of the language tag is often used to distinguish it from other otherwise identical tags.

PR #1619, opened 2022-11-29
Add installing chromium dependencies for playwright to developer readme

myieye:
## Description

While our GitHub CI images seem to include all chromium dependencies by default, dev environments likely don't. So, they'll likely need to install them manually once before they can run playwright tests.

We could install them explicitly in `make playwright-tests`, but that would be a bit of a nuisance, because it requires root privileges.

### Type of Change

- Documentation fix


megahirt (approved):
LGTM

PR #1626, opened 2022-11-29
updated error message about FLEx support

### LINKED ISSUE #1625: bug: outdated message about FLEx support
**Describe the bug**
An error message shows reporting the latest supported version of FLEx is 8.2 but it's actually supposed to be 9 now.

**To Reproduce**

Try to import a project from an unsupported version of FLEx

**Expected behavior**
The message should simply let the user know the project version is not currently supported

**Additional context**

https://github.com/sillsdev/web-languageforge/blob/47bb3acbc07baeb2db47d62aa126749d6c8974fc/src/angular-app/languageforge/lexicon/core/lexicon-send-receive.service.ts#L137

longrunningprocess:
Fixes #1625 

## Description

This PR simply updates an outdated compatibility message

### Type of Change

## Screenshots

N/A

## Checklist

## How to test

- somehow or another, find a project on LD that is older than 9.0 and try to import it.

### DIFF
--- a/src/angular-app/languageforge/lexicon/core/lexicon-send-receive.service.ts
+++ b/src/angular-app/languageforge/lexicon/core/lexicon-send-receive.service.ts
@@ -135,7 +135,7 @@ export class LexiconSendReceiveService {
             break;
           case SendReceiveErrorCodes.ProjectTooOld:
             errorMessage = 'The project \'' + this.status.ProjectCode + '\' is from an ' +
-              'unsupported version of FLEx. The oldest supported FLEx version is 8.2';
+              'unsupported version of FLEx.  Please update to the latest version and try again.';
             break;
           case SendReceiveErrorCodes.ProjectTooNew:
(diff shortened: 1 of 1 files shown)


megahirt (approved):
LGTM

myieye:
My impression is we should publish this without testing it in QA 😬.

longrunningprocess:
it occurred to me that we have our prod data in qa now and this issue was originally related to the `mgb-ib-flex` so maybe trying to S/R that one would do the job.  I think any project whose model version is '7000070' would also work.  I suspect a way to determine that model version might be to do a little searching across plans in language depot for QA.

megahirt:
When I went to test this on QA, the mgb-ib-flex project was already created and already synced, so perhaps this was not the right project?  Or maybe it was already upgraded, I'm not sure.  In any case, I'm happy to mark this as approved/tested as we are just looking at a content change.

PR #1627, opened 2022-11-30
Remove 'semantic domains only' interface languages

### LINKED ISSUE #1616: UI Language picker should not show semantic domain localizations
The UI language picker shows a bunch of "semantic domain only" languages in the list, however this was never a requested feature and IMO only serves to clutter up the actual list of localized UI languages.  We should remove them from the list.  If we find out later that people are using the semantic domain localizations in LF, we can reinstate them, perhaps with better UI.

The PR to fix this should surgically remove the code that adds these semantic domain items to the list, and related plumbing.  If we decide to put the feature back, we can revert the commit.

**Screenshots**
The UI language picker occurs in two places:

The project properties tab under settings:
image

The menu bar UI language picker:
image

laineyhm:
Fixes #1616

## Description

We are removing the languages that were listed as "semantic domains only" in the interface language picker in Project Settings.

### Type of Change

- UI change

## Screenshots
Before:
Screenshot (3)

After:
Screenshot (2)

## Checklist

- [N/A] I have commented my code, particularly in hard-to-understand areas

## How to test

Open Project Settings and click on the dropdown for 'Default Interface Language'. Verify that the 'semantic domains only' language options are no longer in the list.


megahirt:
@laineyhm I cannot see immediately why the e2e tests are failing, but I would suggest that perhaps we should just delete these tests all together - what value do they bring?  They feel like cheap tests that aren't testing much.  Thoughts?

laineyhm:
> @laineyhm I cannot see immediately why the e2e tests are failing, but I would suggest that perhaps we should just delete these tests all together - what value do they bring? They feel like cheap tests that aren't testing much. Thoughts?

Sure, sounds good. I removed the tests. Let me know if you want me to work on adding tests back in in the future.

longrunningprocess:
# testing

✅  they're gone

image

PR #1630, opened 2022-12-02
hide comments and activity from project "observers"

### LINKED ISSUE #1579: Unresolved comments on dashboard should be hidden for users with observer role
Need to review stars for authz, e.g., 

        Here's something I just now realized: the "unresolved comments" should not be visible or clickable for user's whose role is Observer.  They shouldn't be able to see comments in the first place.  So this is a future todo.

_Originally posted by @megahirt in https://github.com/sillsdev/web-languageforge/issues/1525#issuecomment-1303000296_

longrunningprocess:
Fixes #1579 

## Description

This PR establishes some authorization around the visibility of comments to the appropriate project members.

### Type of Change

## Screenshots

view of project dashboard when logged in as a project member with the "observer" role:

image

view of the dashboard for all other roles:

image

## Checklist

## How to test

- View the project dashboard as an observer and ensure "unresolved comments" and the "Activity" section are not shown
- View the project dashboard as an observer (with comments) and ensure "unresolved comments" is shown and the "Activity" section is not shown
- View the project dashboard as any other role and ensure "unresolved comments" and "Activity" section is shown

### DIFF
--- a/next-app/src/routes/password/+server.ts
+++ b/next-app/src/routes/password/+server.ts
@@ -17,10 +17,6 @@ export async function PUT({ request }) {
 
 	const { id } = await fetch_current_user(cookie)
 
-	if (! id) {
-		throwError('User unknown', 404)
-	}
-
 	await sf({
 		name: 'change_password',
 		args: [id, password],
--- a/next-app/src/routes/password/+server.ts
+++ b/next-app/src/routes/password/+server.ts
@@ -1,4 +1,5 @@
 import { json } from '@sveltejs/kit'
+import { current_user } from '$lib/data/user'
 import { throwError } from '$lib/error'
 import { sf } from '$lib/fetch/server'
 
@@ -14,18 +15,15 @@ export async function PUT({ request }) {
 
 	const cookie = request.headers.get('cookie')
 
-	const { userId } = await sf({
-		name: 'session_getSessionData',
-		cookie,
-	})
+	const { id } = await current_user(cookie)
 
-	if (! userId) {
+	if (! id) {
 		throwError('User unknown', 404)
 	}
 
 	await sf({
 		name: 'change_password',
-		args: [userId, password],
+		args: [id, password],
 		cookie,
 	})
 
--- a/next-app/src/routes/password/+server.ts
+++ b/next-app/src/routes/password/+server.ts
@@ -15,7 +15,7 @@ export async function PUT({ request }) {
 
 	const cookie = request.headers.get('cookie')
 
-	const { id } = await current_user(cookie)
+	const { id } = await fetch_current_user(cookie)
 
 	if (! id) {
 		throwError('User unknown', 404)
--- a/next-app/src/routes/password/+server.ts
+++ b/next-app/src/routes/password/+server.ts
@@ -1,5 +1,5 @@
 import { json } from '@sveltejs/kit'
-import { current_user } from '$lib/data/user'
+import { fetch_current_user } from '$lib/data/user'
 import { throwError } from '$lib/error'
 import { sf } from '$lib/fetch/server'
 
--- a/next-app/src/routes/projects/[project_code]/+page.server.ts
+++ b/next-app/src/routes/projects/[project_code]/+page.server.ts
@@ -13,7 +13,7 @@ export async function load({ params: { project_code }, request: { headers }}) {
 		project: await get_project_info(args),
 	}
 
-	const { role } = await current_user(args.cookie)
+	const { role } = await fetch_current_user(args.cookie)
 	if (can_view_activity(role)) {
 		const last_30_days = {
 			start_date: daysAgo(30),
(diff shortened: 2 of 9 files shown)


longrunningprocess:
Some outstanding questions:

1. Should `observer_with_comments` see the "unresolved comments" stat?
2. Should authz also be applied to the rows in the activity table on the dashboard?

myieye:
> # ❓❓❓
> 1. Should `observer_with_comments` see the "unresolved comments" stat?
> 2. Should authz also be applied to the rows in the activity table on the dashboard?

My guess would be yes and yes:
1. `observe_with_comments` means they're a user that contributes to discussions, so I would think it makes sense for them to see the stat.
2. Sounds like comments should be totally invisible for the `observer` role, so any activity related to them would also be hidden.

longrunningprocess:
Thanks!


> 1. `observe_with_comments` means they're a user that contributes to discussions, so I would think it makes sense for them to see the stat.

That's my assumption as well and that's how it's coded right now.

> 2. Sounds like comments should be totally invisible for the `observer` role, so any activity related to them would also be hidden.

Agreed, I'd like @megahirt to weigh in since it will require more work on this.

longrunningprocess:
@megahirt and I discussed this and determined we should not show ANY activity to any "observer"-type role.  I intend to make those changes and ask for another review

next-app/src/routes/projects/[project_code]/+page.server.ts
```diff
@@ -1,24 +1,29 @@
 import { get_activities } from './activities/+server'
 import { get as get_project_info } from './meta/+server'
+import { can_view_activity } from '$lib/auth'
+import { current_user } from '$lib/data/user'
```
Comment (megahirt) [inline next-app/src/routes/projects/[project_code]/+page.server.ts:4]:
```suggestion
import { fetch_current_user } from '$lib/data/user'
```

next-app/src/routes/projects/[project_code]/+page.server.ts
```diff
@@ -1,24 +1,29 @@
...
-		end_date: new Date(),
+	const result = {
+		project: await get_project_info(args),
 	}
-	const activities = await get_activities({ ...last_30_days, ...args })
 
-	return {
-		project,
-		activities,
+	const { role } = await current_user(args.cookie)
```
Comment (megahirt) [inline next-app/src/routes/projects/[project_code]/+page.server.ts:16]:
```suggestion
	const { role } = await fetch_current_user(args.cookie)
```

next-app/src/routes/projects/[project_code]/meta/+server.ts
```diff
@@ -1,32 +1,37 @@
+import { can_view_comments } from '$lib/auth'
+import { current_user } from '$lib/data/user'
```
Comment (megahirt) [inline next-app/src/routes/projects/[project_code]/meta/+server.ts:2]:
```suggestion
import { fetch_current_user } from '$lib/data/user'
```

next-app/src/routes/projects/[project_code]/meta/+server.ts
```diff
@@ -1,32 +1,37 @@
...
-		num_entries_with_audio: entries_with_audio.length,
-		num_entries_with_pictures: entries_with_picture.length,
-		num_unresolved_comments: unresolved_comments.length,
 		num_users: Object.keys(users).length,
+		num_entries: entries.length,
+		num_entries_with_audio: entries.filter(has_audio).length,
+		num_entries_with_pictures: entries.filter(has_picture).length,
 	}
+
+	const { role } = await current_user(cookie)
```
Comment (megahirt) [inline next-app/src/routes/projects/[project_code]/meta/+server.ts:19]:
```suggestion
	const { role } = await fetch_current_user(cookie)
```

megahirt (approved):
This is really great work - thanks @longrunningprocess !  

I suggested naming a function with `fetch_` but it's non-blocking.

I notice that while this is a TS file I don't see any typing going on, meaning that everything is implicitly typed.  I would like to have a separate conversation about our use of types in our next app, which we can schedule separately.  Thanks!

next-app/src/routes/password/+server.ts
```diff
@@ -14,18 +15,15 @@ export async function PUT({ request }) {
 
 	const cookie = request.headers.get('cookie')
 
-	const { userId } = await sf({
-		name: 'session_getSessionData',
-		cookie,
-	})
+	const { id } = await current_user(cookie)
```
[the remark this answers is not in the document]
longrunningprocess: it's certainly easy enough to alias it if the file does in fact get more complicated, e.g., `const { id: userId }`

megahirt:
Just FYI I believe there is a way to batch accept suggestions into a single commit - something to look for in the future.

longrunningprocess:
> Just FYI I believe there is a way to batch accept suggestions into a single commit - something to look for in the future.

Thanks, that would be more convenient.

laineyhm:
# Testing # 
3/3
## Observer ##
✅ No comments or activity section
image

## Observer with comment ##
✅ Comments but no activity section
image

## Other role (here, contributor-non-manager) ##
✅ Comments and activity section
image

PR #1611, opened 2022-11-22
update PHP dependencies and remove communication feature

longrunningprocess:
## Description

This PR seeks to update our PHP deps.  While in the process of upgrading, it was noticed that our SMS feature was never rolled out and was keeping the update from continuing without further research.  It was decided that we could simply remove all the SMS functionality.

### Type of Change

- dependency update and residual unused features

### SMS feature before

image

### SMS feature after

image

## Checklist

- [x] I have performed a self-review of my own code
- [x] I have reviewed the title & description of this PR which I will use as the squashed PR commit message
- [x] I have enabled auto-merge (optional)
- [x]  rebase off `develop` after #1610 is merged

## How to test

- verify automated tests passed
- general smoke test of app
- create an account to ensure emails are unaffected
- ensure communication preference section is gone from "My Profile"

## qa.languageforge.org testing

Testers should add his/her findings to end of the PR in a comment and include screenshots, files, etc that are beneficial.

PR #1632, opened 2022-12-05
Refactor jQuery-dependant selector

myieye:
## Description

While QAing #1611 I bumped into this. `angular.element()` supports [very little if jQuery is not included](https://code.angularjs.org/1.8.2/docs/error/jqLite/nosel). So, it didn't like this selector anymore.

image


longrunningprocess:
# Testing

✅  clean console upon opening share

image

PR #1636, opened 2022-12-06
Bugfix/integrate not uploading playwright results

myieye:
## Description

I noticed on a [failed deployment](https://github.com/sillsdev/web-languageforge/actions/runs/3627940183/jobs/6118378744) that E2E test results were not being uploaded. The reason was that I moved the results, but didn't adapt the action. 😞

I didn't notice this, because we expect there to be no files if tests are all green. But, I think it makes sense to always save results, because:
1) It's a bit confusing to only save results sometimes
2) Saving results is free
3) It can be valuable to be able to compare failed results to successful results

Note: I opted to only **always** save screenshots, **not** traces. The only reason is that the upload takes quite a long time if we include traces. That's because our results our sadly quite messy/bloated due to how we manage pages.

### Type of Change

- Tesitng automation fix

## Checklist

## How to test

Check out the sample PR here: https://github.com/myieye/web-languageforge/actions/runs/3629878852/jobs/6122583829
(Before I opted not to always include traces)

### DIFF
--- a/.github/workflows/integrate-and-deploy.yml
+++ b/.github/workflows/integrate-and-deploy.yml
@@ -79,11 +79,11 @@ jobs:
         run: make e2e-tests-ci
       -
         name: Upload Playwright test results
-        if: always()
         uses: actions/upload-artifact@v3
         with:
           name: test-results
           path: test/e2e/test-results
+          if-no-files-found: error
       -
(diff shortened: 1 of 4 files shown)


megahirt (approved):
LGTM!

PR #1639, opened 2022-12-06
Make label check use current labels

myieye:
## Description

The new label check GHA is problematic in that it currently only respects the labels on the PR at creation time. By adding the token we implicitly tell/enable it to access the "live" labels.

I noticed it here: https://github.com/sillsdev/web-languageforge/pull/1637

The [documenation](https://github.com/marketplace/actions/require-labels) is a bit confusing to me:
image

But I tested it [in my fork](https://github.com/myieye/web-languageforge/pull/17). In that PR I last did it backwards: first suceed and then fail.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -15,6 +15,8 @@ concurrency:
 jobs:
   require-label:
     runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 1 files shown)


megahirt (approved):
wow, this is brilliant.  I couldn't figure it out.  Thanks for this!

PR #1640, opened 2022-12-06
fixed asset versioning

longrunningprocess:
## Description

We noticed during the last release that the cache buster for assets was messed up.  It turns out to be a problem with the timing in which (and how) the E2E tests were being run.  This PR fixes that.

### Type of Change

## Checklist

## How to test

- ensure the `v=` on assets in the devtools is not `9.9.9`.

### DIFF
--- a/.github/workflows/integrate-and-deploy.yml
+++ b/.github/workflows/integrate-and-deploy.yml
@@ -72,11 +72,20 @@ jobs:
         name: Build "next" images
         run: make build-next
       -
+        name: Verify version stamping
+        run: docker run --rm lf-app head build-version.txt version.php
+      -
         name: Unit Tests
         run: make unit-tests-ci
       -
+        name: Verify version stamping
+        run: docker run --rm lf-app head build-version.txt version.php
+      -
         name: Playwright E2E Tests
         run: make e2e-tests-ci
+      -
+        name: Verify version stamping
+        run: docker run --rm lf-app head build-version.txt version.php
       -
         name: Upload Playwright test results
         if: always()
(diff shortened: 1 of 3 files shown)


longrunningprocess:
building the app-for-plawright image again (by the same name as the regular app) was the problem with the cache-buster.  It had me wondering if we _needed_ to rebuild and if needed a separate container for the testing?  There may be opportunities here to simplify things.  The bigger concern for me is that at this time we are not actually testing the same code that was built because the app is being built again...  typically I'd expect the app to be built, started and then have the test runner execute tests against it, not build a new one and test it.

megahirt (approved):
Wow, nice find!  Looks like this was hard to track down.  Thanks for your work.  LGTM

longrunningprocess:
> The purpose of the app-for-playwright container is to volume map the TestControl.php backdoor into the existing built app image. This enables quicker test environment setup. We have already discussed a different approach to test environment setup that involves directly loading environment state into MongoDB, but that would involve a significant e2e testing refactor IMO.

yes, totally agreed, this doesn't warrant loading the db with test data but I'm only envisioning volume mapping that file into the existing app container when running locally or just for the tests... we can look at the possibility of not needing another image built.

> I think this is a case where it makes sense to re-build in a dev environment but not in a CI environment. The purpose of the forced `--build` is to ensure that tests are run against the latest code changes on the dev's machine. One solution is to have the make target `e2e-tests-ci` do everything it needs and not call `make playwright-app`, which has extra stuff e.g. a rebuild.

yep, that could be an option, but if we can simply plug into our existing "watch" capability via the `ui-builder`, there might a way to simply get those updated tests without extra targets, that's my only hope here.

> I disagree that we are testing different code. The rebuild wasn't the cause of the issue as I understand it, it just triggered the issue. The issue was caused by a mis-named image in the compose file, which you fix in this PR.

As I look through lines in the run, anything layer that isn't `CACHED`, in my mind, has the potential of being different:   https://github.com/sillsdev/web-languageforge/actions/runs/3627973894/jobs/6118456362#step:11:119

longrunningprocess:
image

appears to be fixed in QA, no need here for an extra person to verify that I suppose.

PR #1646, opened 2022-12-13
Repair audio seeking

### LINKED ISSUE #1644: bug: seeking audio is broken
**Describe the bug**

When dealing with an audio recording on an element, if you move the slider, neither the playback nor the current time position update. 

**To Reproduce**

1. Go to a project with audio available in entries, e.g., test-chris-02 in staging.
2. move the slider and notice the current time doesn't change.
3. Click play and noticed that the audio starts from the beginning
4. (there is also a console error (`this.$element.find(...).get is not a function`)

**Expected behavior**

Of course, the playback and current time position should be updated.

**Additional context**

This was most likely introduced in #1611 when jquery was removed.

ref:  https://github.com/sillsdev/web-languageforge/blob/develop/src/angular-app/bellows/shared/sound-player.component.ts#L19

laineyhm:
Fixes #1644 

## Description

This PR brings changes to the initialization of the slider input (we are no longer using jquery for that). This ensures that the audio element's current time updates as seeking and playing events occur. By changing the audio element's play/pause logic so that its current time is not reset to 0, we allow for seeking and starting playback in the middle of an audio clip.

### Type of Change

## Screenshots

Current time changes with seeking & play
image

## Checklist

- [N/A] I have commented my code, particularly in hard-to-understand areas

## How to test

Go to a project with audio available in entries, e.g., test-chris-02 in staging.
Move the slider and verify that the current time changes with it.
Click play and verify that the audio starts from the updated time.

### DIFF
--- a/src/angular-app/bellows/shared/sound-player.component.html
+++ b/src/angular-app/bellows/shared/sound-player.component.html
@@ -7,5 +7,5 @@
     </span>
     <!--suppress HtmlFormInputWithoutLabel -->
     <!--The step value is very small to handle a bug encountered where the slider stops after its last full step, before reaching the end 2022-09-->
-    <input id="slider" class="seek-slider" type="range" min="0" max="{{$ctrl.durationInSeconds()}}" step="0.0000000000000001" value="0">
+    <input class="seek-slider" type="range" min="0" max="{{$ctrl.durationInSeconds()}}" step="0.0000000000000001" value="0">
 </span>
--- a/src/angular-app/bellows/shared/sound-player.component.html
+++ b/src/angular-app/bellows/shared/sound-player.component.html
@@ -7,5 +7,5 @@
     </span>
     <!--suppress HtmlFormInputWithoutLabel -->
     <!--The step value is very small to handle a bug encountered where the slider stops after its last full step, before reaching the end 2022-09-->
-    <input class="seek-slider" type="range" min="0" max="{{$ctrl.durationInSeconds()}}" step="0.0000000000000001" value="0">
+    <input id="slider" class="seek-slider" type="range" min="0" max="{{$ctrl.durationInSeconds()}}" step="0.0000000000000001" value="0">
 </span>
--- a/src/angular-app/bellows/shared/sound-player.component.ts
+++ b/src/angular-app/bellows/shared/sound-player.component.ts
@@ -16,7 +16,7 @@ export class SoundController implements angular.IController {
 
   $onInit(): void {
 
-    this.slider = document.getElementById('slider') as HTMLInputElement;
+    this.slider = this.$element[0].querySelector('.seek-slider') as HTMLInputElement;
     this.audioElement.currentTime = 0;
 
     //So that duration appears immediately once it is available
--- a/src/angular-app/bellows/shared/sound-player.component.ts
+++ b/src/angular-app/bellows/shared/sound-player.component.ts
@@ -16,7 +16,8 @@ export class SoundController implements angular.IController {
 
   $onInit(): void {
 
-    this.slider = this.$element.find('.seek-slider').get(0) as HTMLInputElement;
+    this.slider = document.getElementById('slider') as HTMLInputElement;
+    this.audioElement.currentTime = 0;
 
     //So that duration appears immediately once it is available
     this.audioElement.addEventListener('durationchange', () => {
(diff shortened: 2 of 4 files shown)


src/angular-app/bellows/shared/sound-player.component.ts
```diff
@@ -16,7 +16,8 @@ export class SoundController implements angular.IController {
 
   $onInit(): void {
 
-    this.slider = this.$element.find('.seek-slider').get(0) as HTMLInputElement;
+    this.slider = document.getElementById('slider') as HTMLInputElement;
```
megahirt, line 19: There should be a way to select the element by it's class name, since we are already scoped to the component.  I'm thinking something like:
`this.$element.querySelector('.seek-slider')`
but I don't know if $element is usable anymore.  There must be a way to access the HTML element of the component, from inside the component :)

This is preferable to adding an id on the input, since an id is supposed to be unique to the entire page, we are creating a situation where multiple `id="slider"` could be present on the page, one for each audio file.  We can avoid this by using the existing class name `seek-slider`

src/angular-app/bellows/shared/sound-player.component.html
```diff
@@ -7,5 +7,5 @@
     </span>
     <!--suppress HtmlFormInputWithoutLabel -->
     <!--The step value is very small to handle a bug encountered where the slider stops after its last full step, before reaching the end 2022-09-->
-    <input class="seek-slider" type="range" min="0" max="{{$ctrl.durationInSeconds()}}" step="0.0000000000000001" value="0">
+    <input id="slider" class="seek-slider" type="range" min="0" max="{{$ctrl.durationInSeconds()}}" step="0.0000000000000001" value="0">
```
megahirt, line 10: I'm hoping we don't need this extra `id`

megahirt (changes requested):
Looks like you fixed it - nice work!  I'd like to use `querySelector()` instead of `getElementById()`.  Maybe @longrunningprocess can suggest how to do that...

src/angular-app/bellows/shared/sound-player.component.ts
longrunningprocess: > There should be a way to select the element by it's class name, since we are already scoped to the component. I'm thinking something like:
this.$element.querySelector('.seek-slider')
but I don't know if $element is usable anymore. There must be a way to access the HTML element of the component, from inside the component :)

yes, depending on whether `$element` is a standard element or some other jqlite version of an element, see
[querySelector()](https://developer.mozilla.org/en-US/docs/Web/API/Element/querySelector) for the standard api

src/angular-app/bellows/shared/sound-player.component.html
longrunningprocess: agreed

src/angular-app/bellows/shared/sound-player.component.ts
laineyhm: Fixed this morning

src/angular-app/bellows/shared/sound-player.component.html
laineyhm: 👍got it! Made this change.

test/e2e/pages/editor.page.ts
```diff
@@ -50,7 +50,9 @@ export class EditorPage extends BasePage {
     playIconSelector: 'i.fa-play',
     dropdownToggleSelector: 'a.dropdown-toggle',
     uploadButtonSelector: 'button.upload-audio',
-    downloadButtonSelector: 'a.buttonAppend'
+    downloadButtonSelector: 'a.buttonAppend',
+    slider: '#slider',
```
megahirt, line 54: I'm pretty sure the test is failing since the id was removed
```suggestion
    slider: 'input.seek-slider',
```

megahirt (approved):
LGTM!  I think the test is still failing due to the id being removed.  Once that is fixed this is good to go.

myieye:
🧪 Seeking and playingis working great (as well as recording 😁)
https://qa.languageforge.org/app/lexicon/5a8a6da88b2147719239fb33#!/editor/entry/58c0d0fec8f0e77904bacbe4?sortBy=Default&sortReverse=false&wholeWord=false&matchDiacritic=false&filterType=isNotEmpty
image

PR #1662, opened 2022-12-23
Run E2E tests in shards (chromium & firefox)

myieye:
## Description

(1) Splits our E2E tests in to "shards" so that we can run tests in parallel across different environments (Running them in parallel in the same environment doesn't work due to the current implementation of state-management). This way we can scale virtually indefinitely without worrying about workflow execution times.

(2) Activates running E2E tests on firefox instead of only chromium. Due to sharding this doesn't affect workflow execution times. (except for the fact that firefox is slower than chromium)

(3) I split up editor tests into multiple files, which allowed us to go from 4 to 6 shards. If we use more than 6 shards, some of them get 0 tests, which I find quite strange considering we have 16 test files. So, I created [an issue](https://github.com/microsoft/playwright/issues/19681) for them 🤷. The slowest shard still needs about 7 minutes in firefox.

### Type of Change

- Test optimization

## Checklist

- [ ] ~I have added tests that prove my fix is effective or that my feature works~

## How to test

Let 'em rip 🔥


myieye:
The `build-app-run-e2e-tests` check will never turn green, because it doesn't exist anymore.

megahirt:
@myieye this is such a fantastic improvement - I started to do this when we initially created the playwright tests but I wasn't able to achieve it with the current test design.  Well done!!

Now that sharding creates a multitude of e2e workflows, it would be really nice if there was a separate "test results" reporter that we could require.

The alternative is to require every shard, which at current count is 12 workflows?!  There's got to be a better way.  Our existing use of [publish-unit-test-results](https://github.com/sillsdev/web-languageforge/blob/develop/.github/workflows/pull-request.yml#L50) might be appropriate for e2e tests as well.

It looks like [Playwright will create JUnit results files](https://playwright.dev/docs/test-reporters#junit-reporter) that the above action can reference.  @myieye would you wire this up for the PR and then I'll change the PR required actions to only require this new one E2E run?

longrunningprocess (commented):
looks so much more organized now, looks great!  Once you've addressed @megahirt 's findings, I'm good with this one.

longrunningprocess (approved):
Excellent!

PR #1671, opened 2023-01-06
remove GD PHP extension from base-php

megahirt:
## Description

The [PHP GD extension](https://www.php.net/manual/en/book.image.php) is, as far as I can tell, no longer used in our PHP code.  It no doubt was used in the past but I cannot find any reference to methods it provides in our current PHP code, and I have personally run all unit and e2e tests locally without the extension and they all pass.

This feels like a safe removal.  This will also speed up build times for the base-php image, especial for PHP >= 7.4

### Type of Change

- Remove unused dependency

## Checklist

## How to test

This is considered tested with unit and e2e tests all passing


megahirt:
This PR is stuck - the Unit Test Results is required but failed to be reported.  Overriding and merging anyway as all tests pass.

longrunningprocess:
# Testing

✅  confirmed good CI

PR #1705, opened 2023-01-19
LFNext editor components

longrunningprocess:
## Description

This PR will add some of the potential components needed for the LFNext editor pages, see the mock-ups for reference.

## Checklist

## QA testing

There will be nothing to test until these components are actually integrated into live pages.

### DIFF
--- a/next-app/src/lib/Details.svelte
+++ b/next-app/src/lib/Details.svelte
@@ -1,0 +1,23 @@
+<!-- https://daisyui.com/components/collapse -->
+<div class='collapse collapse-arrow'>
+	<input type=checkbox on:click/>
+
+	<div class='collapse-title text-xl font-medium bg-primary text-primary-content z-10 pointer-events-none'>
+		<slot name='summary' />
+	</div>
+
+	<div class=collapse-content>
+		<p><slot /></p>
+	</div>
+</div>
+
+<!--
+	Why "z-10 pointer-events-none" and ":global(.collapse-title button)"?
+
+	see https://github.com/saadeghi/daisyui/discussions/1517
+ -->
+<style>
+:global(.collapse-title button) {
+	pointer-events: all;
+}
+</style>
--- a/next-app/src/routes/+page.svelte
+++ b/next-app/src/routes/+page.svelte
@@ -1,5 +1,6 @@
 <script>
 	import { browser } from '$app/environment'
+    import Details from '$lib/Details.svelte'
 	import { GET } from '$lib/fetch'
 	import { Button, Input, Form } from '$lib/forms'
     import PageHeader from '$lib/PageHeader.svelte'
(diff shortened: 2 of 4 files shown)


longrunningprocess:
I took care of the collapsible section components as well as the left drawer that could house the entry list

next-app/src/lib/Details.svelte
```diff
@@ -0,0 +1,23 @@
...
+		<slot name='summary' />
+	</div>
+
+	<div class=collapse-content>
+		<p><slot /></p>
+	</div>
+</div>
+
+<!--
+	Why "z-10 pointer-events-none" and ":global(.collapse-title button)"?
```
megahirt, line 15: I appreciate this good documentation

megahirt:
Here is a recording of the components in action using lf-next `npm run dev`
lfnext

megahirt (approved):
Looks good so far!

next-app/src/routes/+page.svelte
```diff
@@ -31,93 +32,176 @@
...
+	</div>
+
+	<div class='drawer-side not-prose'>
+		<label for=left_drawer class=drawer-overlay />
+
+		<ul class='menu p-4 w-fit bg-base-100 text-base-content'>
+			<li class='menu-title grid grid-cols-3'>
+				<span on:click={ () => sort_items_by('name') } on:keydown={ () => sort_items_by('name') } class='cursor-pointer w-fit'>Name</span>
+				<span on:click={ () => sort_items_by('prop') } on:keydown={ () => sort_items_by('name') } class='cursor-pointer w-fit'>Prop</span>
+				<span on:click={ () => sort_items_by('number') } on:keydown={ () => sort_items_by('name') } class='cursor-pointer w-fit'>Number</span>
```
myieye, line 191: @longrunningprocess I guess this should be `on:keydown={ () => sort_items_by('number')`. Same on line above it.

hahn-kev: Seems like they should always be the same. Can we create something like an angular directive to make them both the same?

next-app/src/routes/+page.svelte
```diff
@@ -31,93 +32,176 @@
...
+			number: 77777777777,
+		},
+	]
 
 	onMount(() => dark_mode = window.matchMedia('(prefers-color-scheme: dark)').matches)
 
 	$: browser && document.documentElement.setAttribute('data-theme', dark_mode ? 'dark' : 'light')
+
+	function sort_items_by(field) {
+		items = items.sort((a, b) => a[field] < b[field] ? 1 : -1)
```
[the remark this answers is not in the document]
megahirt: https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore

Are we targeting ES6?  I'm not sure...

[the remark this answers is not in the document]
hahn-kev: I agree, I'd rather avoid it. If we really want some utility these are a really good option since they're one offs without any dependencies. 
https://anguscroll.com/just/

next-app/src/routes/+page.svelte
longrunningprocess: we could create what Svelte calls an action to watch both events and then do the same thing but this should be a rare case, i.e., watching a click on a non-interactive element... maybe I should just make the `a`'s.  This was for a11y's sake.

longrunningprocess: this is probably the better solution all around:  [decided on better mark-up](https://github.com/sillsdev/web-languageforge/commit/d7f8bfc76d3f89bb825c8165f9874a27decacc40)

next-app/src/routes/+page.svelte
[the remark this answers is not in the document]
longrunningprocess: I had forgotten about that just lib and do like it but the sort-by's second parm can't handle nested sort out of the box, we'd have to customize the incoming function which is possible of course but at that point is becoming less worth the effort when the code essentially does the same thing (https://github.com/angus-c/just/blob/master/packages/array-sort-by/index.mjs).  I'm certainly open to more discussion.

PR #1706, opened 2023-01-20
Delete User Account feature polish

### LINKED ISSUE #1704: Remove user records of deleted accounts from user list
When users delete their accounts, their user record remains, void of personal information. 

Let's filter the user list in Site Administration to exclude the user records of deleted accounts, like these:

image

megahirt:
## Description

The main purpose of this PR was to fix the problem of seeing deleted users in the user list in the site admin app.  I added a isDeleted flag to the UserModel and filtered out users with this flag.

In the process of fixing the above, I added a few additional small changes:

- remove unused references in UserModel 

- remove ui-dist volume on every build 

This can avoid tricky problems when switching branches and having run `make dev` on a different branch.  Essentially, it is possible to build and run the application using `make`, but the ui-dist volume will come from whatever the previous run of `make dev` was, even if it was on an earlier commit or branch.  This change ensures that each time the application is built, we start with a accurate ui-dist volume.

The ui-dist volume is only really used in the `make dev` use case, but we include the volume in the app docker compose service to make it work together with the ui-builder service.  This has the unfortunately side effect that the ui-dist volume is active on the app service even when no ui-builder service is active (for doing a webpack watch)

- report all errors to the user 

We no longer want to swallow and ignore unknown errors.  We want to know about them so we can fix them.

 
- Additionally, I decided to change the username of a deleted user to something more human readable instead of the ID string.
 
- fix logic error in system admin check 

This fixes a logic error when checking for the system admin role of the current user.  The code was checking the role of the user to be deleted, instead of the role of the current user to determine if the delete operation could proceed.

### Before
image

### After
image

## Note to the reviewer
Because there are several small changes, some unrelated, it might be helpful to review this PR one commit at a time.

## Checklist

~~I have added tests that prove my fix is effective or that my feature works~~

## QA testing

- Ensure that a deleted user does not show up in the user list system admin app
- As an admin, you should now be able to remove multiple users from the user list app.  There was a bug that previously prevented this.
- As an admin, try and delete a user who owns a project.  An error message should display preventing the operation.
- `make` should now behave correctly and contain the latest ui-dist build even if `make dev` was run from a previous commit (local developer testing scenario only)


longrunningprocess (approved):
looks like some nice cleanup, thank you!

megahirt:
Normally I wouldn't test my own stuff in QA, but we'll let this go so that we can move stuff toward shipping.

TESTED IN QA:
- verified that deleted user does not show up in the list
- tested deleting a user who own's a project - got expected error message
- tested deleting multiple users using checkbox - OK

I didn't test the `make` / `make dev` use case but that's not relevant to QA testing.  I'm sure that will become known if there are issues with this change.

PR #1716, opened 2023-01-25
migrate to sveltekit action

longrunningprocess:
## Description

This PR migrates the change password functionality toward a newer SvelteKit feature called [Actions](https://kit.svelte.dev/docs/form-actions).

## Checklist

## QA testing

- ensure passwords are required
- ensure the passwords match
- ensure password is actually updated to a new password

### DIFF
--- a/next-app/src/routes/password/change/+page.server.ts
+++ b/next-app/src/routes/password/change/+page.server.ts
@@ -1,0 +1,33 @@
+import { sf, type Rpc } from '$lib/server/sf'
+import { fetch_current_user } from '$lib/server/user'
+import { fail, redirect } from '@sveltejs/kit'
+import type { RequestEvent } from '../$types'
+
+export const actions = {
+  default: async ({ request }: RequestEvent) => {
+	const submissions = await request.formData()
+
+	const new_password = submissions.get('new_password')
+	const new_password_confirm = submissions.get('new_password_confirm')
+
+	if (! new_password) {
+		return fail(400, { failed: 'Password is required' })
+	}
+
+	if (new_password !== new_password_confirm) {
+		return fail(400, { failed: 'Passwords do not match' })
+	}
+
+	const cookie = request.headers.get('cookie') || ''
+
+	const { id } = await fetch_current_user(cookie)
+
+	await sf({
+		name: 'change_password',
+		args: [id, new_password],
+		cookie,
+	})
+
+	throw redirect(303, '/password/changed')
+  }
+}
--- a/next-app/src/routes/password/change/+page.server.ts
+++ b/next-app/src/routes/password/change/+page.server.ts
@@ -1,4 +1,4 @@
-import { sf, type Rpc } from '$lib/server/sf'
+import { sf } from '$lib/server/sf'
 import { fetch_current_user } from '$lib/server/user'
 import { fail, redirect } from '@sveltejs/kit'
 import type { RequestEvent } from '../$types'
--- a/next-app/src/routes/password/change/+page.server.ts
+++ b/next-app/src/routes/password/change/+page.server.ts
@@ -7,7 +7,7 @@ export const actions = {
 	default: async ({ request }: RequestEvent) => {
 		const submissions = await request.formData()
 
-		const new_password = submissions.get('new_password')
+		const new_password = submissions.get('new_password') as string
 		const new_password_confirm = submissions.get('new_password_confirm')
 
 		if (! new_password) {
--- a/next-app/src/routes/password/change/+page.server.ts
+++ b/next-app/src/routes/password/change/+page.server.ts
@@ -4,30 +4,30 @@ import { fail, redirect } from '@sveltejs/kit'
 import type { RequestEvent } from '../$types'
 
 export const actions = {
-  default: async ({ request }: RequestEvent) => {
-	const submissions = await request.formData()
+	default: async ({ request }: RequestEvent) => {
+		const submissions = await request.formData()
 
-	const new_password = submissions.get('new_password')
-	const new_password_confirm = submissions.get('new_password_confirm')
+		const new_password = submissions.get('new_password')
+		const new_password_confirm = submissions.get('new_password_confirm')
 
-	if (! new_password) {
-		return fail(400, { failed: 'Password is required' })
+		if (! new_password) {
+			return fail(400, { failed: 'Password is required' })
+		}
+
+		if (new_password !== new_password_confirm) {
+			return fail(400, { failed: 'Passwords do not match' })
+		}
+
+		const cookie = request.headers.get('cookie') || ''
+
+		const { id } = await fetch_current_user(cookie)
+
+		await sf({
+			name: 'change_password',
+			args: [id, new_password],
+			cookie,
+		})
+
+		throw redirect(303, '/password/changed')
 	}
-
-	if (new_password !== new_password_confirm) {
-		return fail(400, { failed: 'Passwords do not match' })
-	}
-
-	const cookie = request.headers.get('cookie') || ''
-
-	const { id } = await fetch_current_user(cookie)
-
-	await sf({
-		name: 'change_password',
-		args: [id, new_password],
-		cookie,
-	})
-
-	throw redirect(303, '/password/changed')
-  }
 }
--- a/next-app/src/routes/password/change/+page.svelte
+++ b/next-app/src/routes/password/change/+page.svelte
@@ -6,7 +6,7 @@
 	} from '$lib/forms'
 	import PageHeader from '$lib/PageHeader.svelte'
 	import { error } from '$lib/error'
-    import type { ActionData } from './$types'
+	import type { ActionData } from './$types'
 
 	export let form: ActionData
 
--- a/next-app/src/routes/password/change/+page.svelte
+++ b/next-app/src/routes/password/change/+page.svelte
@@ -1,23 +1,20 @@
-<script>
-	import { goto } from '$app/navigation'
-	import { UPDATE } from '$lib/fetch'
+<script lang=ts>
 	import {
 		Button,
 		Form,
 		Input,
 	} from '$lib/forms'
 	import PageHeader from '$lib/PageHeader.svelte'
+	import { error } from '$lib/error'
+    import type { ActionData } from './$types'
+
+	export let form: ActionData
 
 	let new_password = ''
 	let new_password_confirm = ''
 
-	async function change_password() {
-		await UPDATE({url: '/password', body: {
-			password: new_password,
-			password_confirm: new_password_confirm
-		}})
-
-		goto('/password/changed')
+	$: if (form?.failed) {
+		$error = Error(form.failed)
 	}
 </script>
 
@@ -25,15 +22,15 @@
 	<title>Change your password</title>
 </svelte:head>
 
-<div class='mx-auto max-w-lg'>
+<section class='mx-auto max-w-lg'>
 	<PageHeader>
 		Change your password
 	</PageHeader>
 
-	<Form on:submit={ change_password }>
-		<Input label='New password:' type=password bind:value={ new_password } required autofocus />
-		<Input label='Confirm new password:' type=password bind:value={ new_password_confirm } required />
+	<Form>
+		<Input label='New password:' name=new_password type=password bind:value={ new_password } required autofocus />
+		<Input label='Confirm new password:' name=new_password_confirm type=password bind:value={ new_password_confirm } required />
 
 		<Button>Change my password</Button>
 	</Form>
-</div>
+</section>
--- a/next-app/src/routes/password/change/+page.ts
+++ b/next-app/src/routes/password/change/+page.ts
@@ -1,1 +1,0 @@
-export const prerender = true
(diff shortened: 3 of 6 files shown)


next-app/src/routes/password/change/+page.svelte
```diff
@@ -1,39 +1,36 @@
...
-<div class='mx-auto max-w-lg'>
+<section class='mx-auto max-w-lg'>
 	<PageHeader>
 		Change your password
 	</PageHeader>
 
-	<Form on:submit={ change_password }>
-		<Input label='New password:' type=password bind:value={ new_password } required autofocus />
-		<Input label='Confirm new password:' type=password bind:value={ new_password_confirm } required />
+	<Form>
```
hahn-kev, line 30: I know that by convention since this is under the route 'password/change' that it'll get handled on the corresponding `page.server.ts` but I would like to propose a convention to always specify the action attribute (in this case it should be `/password/change` if I understand correctly) so that if this code gets refactored to another page (potentially via a component) it'll still work.

megahirt: I'm a bit torn on this, because I think one of the selling points of Svelte is that so much is implicit in the architecture that there is simply less code to write.  I can really go either way.  It's good to have conventions especially since we are a team working on shared code.  It's also looks elegant to see a simple `<form>` element that just works.

megahirt (approved):

next-app/src/routes/password/change/+page.server.ts
```diff
@@ -0,0 +1,33 @@
+import { sf, type Rpc } from '$lib/server/sf'
+import { fetch_current_user } from '$lib/server/user'
+import { fail, redirect } from '@sveltejs/kit'
+import type { RequestEvent } from '../$types'
+
+export const actions = {
+  default: async ({ request }: RequestEvent) => {
+	const submissions = await request.formData()
```
myieye, line 8: [nit] Looks like this function-body should be indented

next-app/src/routes/password/change/+page.server.ts
```diff
@@ -0,0 +1,33 @@
...
+
+	const new_password = submissions.get('new_password')
+	const new_password_confirm = submissions.get('new_password_confirm')
+
+	if (! new_password) {
```
myieye, line 13:
```suggestion
	if (!new_password?.trim()) {
```

next-app/src/routes/password/change/+page.server.ts
```diff
@@ -0,0 +1,33 @@
...
+
+	const { id } = await fetch_current_user(cookie)
+
+	await sf({
+		name: 'change_password',
+		args: [id, new_password],
+		cookie,
+	})
+
+	throw redirect(303, '/password/changed')
```
myieye, line 31: `throw` is part of the natural flow? 😲

next-app/src/routes/password/change/+page.server.ts
```diff
@@ -0,0 +1,33 @@
...
+		return fail(400, { failed: 'Passwords do not match' })
+	}
+
+	const cookie = request.headers.get('cookie') || ''
+
+	const { id } = await fetch_current_user(cookie)
+
+	await sf({
+		name: 'change_password',
+		args: [id, new_password],
```
myieye, line 27: I get a compile error here that is fairly confusing, but ultimately it's because `new_password` isn't definitely a string and `args` needs to be either all strings or all objects. `submissions.get('new_password') as string` would be one way to go.
image

next-app/src/routes/password/change/+page.svelte
```diff
@@ -1,39 +1,36 @@
...
-	import { UPDATE } from '$lib/fetch'
+<script lang=ts>
 	import {
 		Button,
 		Form,
 		Input,
 	} from '$lib/forms'
 	import PageHeader from '$lib/PageHeader.svelte'
+	import { error } from '$lib/error'
+    import type { ActionData } from './$types'
```
myieye, line 9: The indentation here is spaces. Everything else is tabs 🤔

next-app/src/routes/password/change/+page.ts
```diff
@@ -1 +0,0 @@
-export const prerender = true
```
myieye, line 1: I'm curious why this got deleted 🙂

myieye (changes requested):
Just a few nits mostly 😉

I'm looking at the PHP...is there actually no password validation when changing your pasword 😲? In the long run I assume this validation will land in the new API, but this is maybe the best place for it now.

next-app/src/routes/password/change/+page.server.ts
megahirt: I wasn't familiar with the use of `?` until I looked it up - it is a pre-check for null/undefined on the object.

next-app/src/routes/password/change/+page.server.ts
megahirt: I also find this concerning, but apparently this is how you do it in SvelteKit.  See https://kit.svelte.dev/docs/load#redirects

next-app/src/routes/password/change/+page.svelte
longrunningprocess: > to always specify the action attribute

I like that kind of defensive coding thought.  In my mind these are so tightly coupled, any refactoring, especially of a route, is going to come with changes to both.  I'm open to addressing this if/when a situation comes up and thinking through convention again at that point.

next-app/src/routes/password/change/+page.server.ts
longrunningprocess: thanks, [0f3b636](https://github.com/sillsdev/web-languageforge/pull/1716/commits/0f3b63692394cccd007966ec38d26bf878a03a28)

next-app/src/routes/password/change/+page.svelte
hahn-kev: alright that sounds good. To me it was less refactoring the route, and more moving the form code into a shared component that might be used on another page.

next-app/src/routes/password/change/+page.server.ts
hahn-kev: yeah I almost commented on it. It is weird, but it does mean that you can redirect from deep inside some code that might be checking something. At least we're on v1 now so we don't have to worry about this kind of thing changing.

next-app/src/routes/password/change/+page.svelte
longrunningprocess: at that point, we might be more inclined to move toward an "endpoint" rather than an "action"

next-app/src/routes/password/change/+page.server.ts
megahirt: So I am still bothered by `throw` being part of the normal code path here.  So I think here's what's going on:
SvelteKit's `redirect()` is treated like `error()` and so you are supposed to use `throw`.  I think the logic goes something like "redirects are an unexpected error state where the user asks for one thing, but it's moved and so you have to go somewhere else (that you didn't expect) to get it"

I wonder if instead it should use:
`import { goto } from '$app/navigation';`
then
`goto('/password/changed');`

next-app/src/routes/password/change/+page.server.ts
longrunningprocess: it might time for us to integrate zxcvbn, trim this feels like the same sort of thing as making sure the provided password is a good one.  I'm not immediately inclined to get rid of those trailing spaces in case it was intentional... I guess we'll need more discussion here.

next-app/src/routes/password/change/+page.server.ts
longrunningprocess: yes, it's meant to follow the paradigm of "breaking" normal code flow just like a redirect does.  Similar to the way Svelte uses a valid construct to achieve new concepts, e.g., `export let var` and `$:`.  I'm ok with the mental model shift.

megahirt: ok, so it is a mental model shift - I can handle that.

Would you agree that we should `goto` /password/changed for the success message instead of `redirect`ing there?  It seems more natural to goto with a 200 instead of redirecting with a 303 See Other.

longrunningprocess: > I wonder if instead it should use:
> `import { goto } from '$app/navigation';`
> then
> `goto('/password/changed');`

that would probably accomplish the same thing but not necessarily follow the HTTP flow as well and I sort of like the throw in this case because there are many times in the past where I've had to setup a redirect response and then explicitly return the response to early-exit the current code or function.  This approach felt very natural to me the first time I saw it because setting up a response and then return it later always felt like a leak to me when it should just be finished at that point.

next-app/src/routes/password/change/+page.server.ts
longrunningprocess: yep, the `as string` is fine, corrected in [bf77994](https://github.com/sillsdev/web-languageforge/pull/1716/commits/bf77994b9fe002062e8e25cdbfac501aa3ddadc3)

next-app/src/routes/password/change/+page.ts
longrunningprocess: constraint of utilizing the "action" paradigm, some early discussion here: [Prerendered page alongside non-GET request handler should be an error](https://github.com/sveltejs/kit/issues/4252#top).

doc reference: https://kit.svelte.dev/docs/page-options#prerender-when-not-to-prerender

> Pages with [actions](https://kit.svelte.dev/docs/form-actions) cannot be prerendered, because a server must be able to handle the action POST requests.

next-app/src/routes/password/change/+page.server.ts
longrunningprocess: > ok, so it is a mental model shift - I can handle that.
> 
> Would you agree that we should `goto` /password/changed for the success message instead of `redirect`ing there? It seems more natural to goto with a 200 instead of redirecting with a 303 See Other.

I'd say the 303 is tailer-made for this scenario (https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/303), where the resource in this case was `/password/change` and the new one is `passowrd/changed`.

next-app/src/routes/password/change/+page.svelte
longrunningprocess: probably an auto-import, I'll have to check and see if that's configurable.  At any rate, thank you, corrected in [f0b8a45](https://github.com/sillsdev/web-languageforge/pull/1716/commits/f0b8a45bacb4ffa1296ec880c7d3473becbdf47f)

longrunningprocess:
> Just a few nits mostly 😉
> 
> I'm looking at the PHP...is there actually no password validation when changing your pasword 😲? In the long run I assume this validation will land in the new API, but this is maybe the best place for it now.

it depends where you look 😆 , but yes, we should have something like this in place:  https://svelte.dev/repl/2d597987426345d697c5c6729dc2cacc?version=3.55.1

hahn-kev (approved):

next-app/src/routes/password/change/+page.server.ts
megahirt: From the MDN docs, it would seem that 302 is appropriate for a success confirmation page.  I have never seen a 302 used in this way before, but a lot of this is new to me so I'll continue to learn.

next-app/src/routes/password/change/+page.svelte
megahirt: we should revisit Prettier or another code formatter for Svelte so that we are not dealing with these types of changes, or have inconsistencies like tabs and spaces.

next-app/src/routes/password/change/+page.server.ts
megahirt: I don't know if zxcvbn is still a good choice for password strength these days...would be good to re-evaluate.

megahirt:
> I'm looking at the PHP...is there actually no password validation when changing your password 😲?

I would suggest that password validation (making sure the user typed it the same twice) should be done client-side and doesn't belong in the API layer.  I can see making a case for password strength being enforced at the API layer, but we don't currently do that in PHP.

myieye (approved):

myieye:
> > I'm looking at the PHP...is there actually no password validation when changing your password 😲?
> 
> I would suggest that password validation (making sure the user typed it the same twice) should be done client-side and doesn't belong in the API layer. I can see making a case for password strength being enforced at the API layer, but we don't currently do that in PHP.

Oh, of course you're right 🙈. We only pass 1 password to the server. Not sure what I was thinking there.

next-app/src/routes/password/change/+page.server.ts
longrunningprocess: I think that's a really good point @megahirt ... maybe we should abandon the idea then.

longrunningprocess:
> > I'm looking at the PHP...is there actually no password validation when changing your password 😲?
> 
> I would suggest that password validation (making sure the user typed it the same twice) should be done client-side and doesn't belong in the API layer. I can see making a case for password strength being enforced at the API layer, but we don't currently do that in PHP.

yep, excellent point, I'll change that.

myieye:
🧪 See my test results on [#1719](https://github.com/sillsdev/web-languageforge/pull/1719#issuecomment-1413356079).

laineyhm:
> 🧪 See my test results on [#1719](https://github.com/sillsdev/web-languageforge/pull/1719#issuecomment-1413356079).

Ditto

PR #1719, opened 2023-01-30
refactor password checks to client-side

longrunningprocess:
## Description

This is a follow-through on PR feedback from #1716 

## Checklist

## QA testing

- While trying to change your password, ensure you cannot submit the form unless you've provided a password
- While trying to change your password, ensure you cannot submit the form unless you've provided a confirmation password
- While trying to change your password, ensure you cannot submit the form unless you've provided a confirmation password that matches the new password
- While trying to change your password, ensure you can submit the form and your password has actually changed.

### DIFF
--- a/next-app/src/lib/debounce.ts
+++ b/next-app/src/lib/debounce.ts
@@ -1,0 +1,7 @@
+let id: ReturnType<typeof setTimeout>
+
+export function debounce(fn: () => void, delay_in_ms: number) {
+	clearTimeout(id) // cancels any previous ones that may not have fired the fn yet, i.e., it resets the delay so it's a delay from the last call not the first
+
+	id = setTimeout(fn, delay_in_ms)
+}
--- a/next-app/src/routes/password/change/+page.svelte
+++ b/next-app/src/routes/password/change/+page.svelte
@@ -6,7 +6,7 @@
 	} from '$lib/forms'
 	import PageHeader from '$lib/PageHeader.svelte'
 	import { dismiss, error } from '$lib/error'
-        import { debounce } from '$lib/debounce'
+	import { debounce } from '$lib/debounce'
 	import type { ActionData } from './$types'
 
 	export let form: ActionData
--- a/next-app/src/routes/password/change/+page.svelte
+++ b/next-app/src/routes/password/change/+page.svelte
@@ -6,6 +6,7 @@
 	} from '$lib/forms'
 	import PageHeader from '$lib/PageHeader.svelte'
 	import { dismiss, error } from '$lib/error'
+    import { debounce } from '$lib/debounce'
 	import type { ActionData } from './$types'
 
 	export let form: ActionData
@@ -24,14 +25,6 @@
 
 	function verify(password_confirm: string, password: string) {
 		password_confirm !== password ? $error = Error('Passwords are not the same') : dismiss()
-	}
-
-	function debounce(fn: () => void, delay: number) {
-		const id = setTimeout(() => {
-			fn()
-
-			clearTimeout(id)
-		}, delay)
 	}
 </script>
 
--- a/next-app/src/routes/password/change/+page.svelte
+++ b/next-app/src/routes/password/change/+page.svelte
@@ -6,7 +6,7 @@
 	} from '$lib/forms'
 	import PageHeader from '$lib/PageHeader.svelte'
 	import { dismiss, error } from '$lib/error'
-    import { debounce } from '$lib/debounce'
+        import { debounce } from '$lib/debounce'
 	import type { ActionData } from './$types'
 
 	export let form: ActionData
--- a/next-app/src/routes/password/change/+page.svelte
+++ b/next-app/src/routes/password/change/+page.svelte
@@ -5,7 +5,7 @@
 		Input,
 	} from '$lib/forms'
 	import PageHeader from '$lib/PageHeader.svelte'
-	import { error } from '$lib/error'
+	import { dismiss, error } from '$lib/error'
 	import type { ActionData } from './$types'
 
 	export let form: ActionData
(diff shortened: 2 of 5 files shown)


next-app/src/routes/password/change/+page.svelte
```diff
@@ -16,6 +16,23 @@
...
+
+	function verify(password_confirm: string, password: string) {
+		password_confirm !== password ? $error = Error('Passwords are not the same') : dismiss()
+	}
+
+	function debounce(fn: () => void, delay: number) {
+		const id = setTimeout(() => {
+			fn()
+
+			clearTimeout(id)
```
hahn-kev, line 33: why the clear after? a timeout is only going to fire once and if you call the debounce function multiple times it'll setup a new scope each time with a new id. Should the id be stored in a variable outside the scope of the function so that we can cancel any pending timeouts whenever this function is called?

next-app/src/routes/password/change/+page.svelte
```diff
@@ -16,6 +16,23 @@
 	$: if (form?.failed) {
 		$error = Error(form.failed)
```
hahn-kev, line 17: how does `$error` work? I don't see it defined anywhere.

next-app/src/routes/password/change/+page.svelte
longrunningprocess: good catch, thank you, corrected in [6b7cca3](https://github.com/sillsdev/web-languageforge/pull/1719/commits/6b7cca322ac7a50dfcadbbcf1c202d6ccee123f8)

next-app/src/lib/debounce.ts
```diff
@@ -0,0 +1,7 @@
+let id: ReturnType<typeof setTimeout>
+
+export function debounce(fn: () => void, delay_in_ms: number) {
```
myieye, line 3: #lodash 😄

longrunningprocess: I looked into the just-debounce I believe but still needed to "reset" with my own code so it didn't make sense but you're right, if we had pulled in lodash, it might've already been there to serve the purpose.  FWIW, if I'd been able to get it to bundle properly, it would be in already!

next-app/src/routes/password/change/+page.svelte
longrunningprocess: I'm going to presume you'd resolve this so I can get it merged.  Please let me know if you would've felt differently.

next-app/src/routes/password/change/+page.svelte
```diff
@@ -16,6 +17,15 @@
...
 	}
+
+	$: new_password_confirm && debounce(() => verify(new_password_confirm, new_password), 400)
```
myieye, line 21: I find this 400ms delay a bit quirky. It also puts the error message out of sync with whether the button is enabled or disabled. Considering the comparison check doesn't cost us anything, I think it make sense to reduce it or even just remove it.

myieye: I also don't like getting error messages as I type the confirmation password 🤔. What do you think about always leaving the Submit button enabled? Can we only call verify on blur and submit instead?

next-app/src/routes/password/change/+page.svelte
```diff
@@ -16,6 +17,15 @@
...
+
+	$: submittable = new_password && new_password_confirm && new_password_confirm === new_password
+	$: disabled = ! submittable
+
+	function verify(password_confirm: string, password: string) {
+		password_confirm !== password ? $error = Error('Passwords are not the same') : dismiss()
```
myieye, line 27: Instead of an error bubble, I think it would ultimately be nicer to have the error directly under the relevant field. This is a special case, where it could ultimately be either field that's "wrong", but I think it would still be fine to just put it under the confirmation field.

myieye:
## 🧪 QA testing

- Started some more conversations around the UX ☝️
- ❌ I can submit an invalid form using Enter

There doesn't seem to be a standard form toolkit in svelte yet. Is that right? I feel like all the things I've brought up, hinge on that.

next-app/src/routes/password/change/+page.svelte
longrunningprocess: yeah, based on our recent conversation, I'll just get rid of the bells and whistles here and stick with an error once the user submits (and leave the button enabled the whole time)

next-app/src/routes/password/change/+page.svelte
longrunningprocess: agreed, when it comes to forms, not utilizing the app-wide error solution would be better for the user.

longrunningprocess:
> There doesn't seem to be a form toolkit in the standard svelte libs. Is that right? I feel like all the things I've brought up, hinge on that. Perhaps we should consider adding something like [svelte-forms](https://chainlist.github.io/svelte-forms)?

correct, there is not a built-in for that yet but as you found already, there are some nice solutions out there already. I can consider that.

> Being able to submit invalid forms with Enter is the only "bug" I found. So, if you want, you can just fix that, move this along and then we look at the UX in a later PR.

I need to look closer, I would've expected a standard form with a single button to automatically get the "submit" behavior even on enter but maybe I'm remembering something wrong there.  Thanks for checking that though.

myieye:
> > Being able to submit invalid forms with Enter is the only "bug" I found. So, if you want, you can just fix that, move this along and then we look at the UX in a later PR.
> 
> I need to look closer, I would've expected a standard form with a single button to automatically get the "submit" behavior even on enter but maybe I'm remembering something wrong there. Thanks for checking that though.

@longrunningprocess To prevent submitting, you're currently disabling the button. I guess Svelte[Kit] doesn't read the disabled property to determine whether the user is allowed to submit the form. So, I think what we want to do is:
1) Always leave the submit button enabled
2) Make use of [`use:enhance`](https://kit.svelte.dev/docs/form-actions#progressive-enhancement-use-enhance). Something like this:
```
use:enhance={({ cancel }) => { 
	if (passwordsDontMatch) {
		showError();
		cancel();
	}
}}
```

laineyhm:
### While trying to change your password, ensure you cannot submit the form unless you've provided a password
✅ 
image

### While trying to change your password, ensure you cannot submit the form unless you've provided a confirmation password
✅ 
image

### While trying to change your password, ensure you cannot submit the form unless you've provided a confirmation password that matches the new password
✅ 
image

### While trying to change your password, ensure you can submit the form and your password has actually changed.
✅ 
image

PR #1728, opened 2023-02-15
Redirect to projects and clear session project on project exceptions

### LINKED ISSUE #1708: bug: a user's `lastUsedProjectId:` might not exist anymore
**Describe the bug**
I was having a problem logging into staging.  I'm not sure whether this was the underlying issue but after I changed my password I began noticing a new error.

**To Reproduce**
I presume these steps are what led to my issue:

1. Go to a project
2. delete that project
3. log out
4. try to log in again

**Expected behavior**
When a user comes into the app, the app should not fail to get him/her to a project

**Additional context**
Either the `lastUsedProjectId:` for a user needs to be updated when a project is deleted or the upfront algorithm needs to handle the possibility that the is might not exist anymore.

This might be related to #1600 , not totally sure.

### LINKED ISSUE #1726: bug: OAuth login sometimes redirects users to homescreen
**Describe the bug**
Several devs (@myieye & @longrunningprocess) in staging as well as a [user in production](https://community.software.sil.org/t/google-sign-in-issues/7238) have reported that they are redirected to the home page when they try to login in OAuth.

**To Reproduce**
Not all users/accounts are affected. We still need to determine what causes this to happen. For users with the bug:

1. Go to the login page: /auth/login
2. Click on 'Sign in with Google'
3. Select the prolematic account
4. After a few redirects: tada you're on the home page and not logged in 😢 (in my recording I land on the login page? 🤔 I think I've landed on the homepage before...weird)

**Expected behavior**
The login should work.

**Watch it in real life 🍿**
https://user-images.githubusercontent.com/12587509/218985894-cfc453ef-4598-436f-8358-8925da7fdba9.mp4

Network log from the moment I select the Google account I want to login with:
image

myieye:
## Description

Ensure that deleted projects are removed from users entirely (including `lastProjectUsed`. Also change many redirects from `/logout` to `/projects`. Ultimately this is a bit overkill, but will hopefully minimize similar hassles in the future. Invalid project-IDs in URLs now land on the projects page.

We commonly redirect to /logout when an unexpected Exception occurs. The thinking is presumably: we'll clear the session and maybe it will work next time. But, I think it often makes just as much sense to assume that the problem is project-specific, so going to the projects page removes the hassle of having to log in again. The user is authenticated, so there's no security issue to worry about. Worst case is that the projects page doesn't work and the user lands in an infinite redirect loop. 🤷

## Checklist

## Testing

1) Create 2 users
2) User 1 creates a project and invites user 2
3) User 2 joins project and logs out (or not 🤷)
4) User 1 delete project
5) User 3 can still log in and lands on the project list page

### DIFF
--- a/src/Site/Controller/Redirect.php
+++ b/src/Site/Controller/Redirect.php
@@ -29,7 +29,7 @@ class Redirect extends App
         } catch (\Exception $e) {
             // Don't know what went wrong, so go to logout route to clear the session
             // This will then redirect to the login page
-            return $app->redirect("/auth/logout");
+            return $this->redirectToProjects($app);
         }
         try {
             // Get most recent project ID, either from PHP session or from user's lastUsedProjectID in MongoDB
@@ -38,13 +38,13 @@ class Redirect extends App
             if (SilexSessionHelper::getUserId($app)) {
                 // User tried to access project they're not a member of, so show them projects view so they can pick a different one
                 // This can happen if the user was removed from the project by a manager between their last login and now
-                return $app->redirect("/app/projects");
+                return $this->redirectToProjects($app);
             }
             // Session somehow persisted despite user being logged out, so go to logout route to clear the session
             return $app->redirect("/auth/logout");
         } catch (\Exception $e) {
-            // Don't know what went wrong, so go to logout route to clear the session
-            return $app->redirect("/auth/logout");
+            // Don't know what went wrong, so we'll assume it's project specific and go to the list
+            return $this->redirectToProjects($app);
         }
         if ($projectId) {
             try {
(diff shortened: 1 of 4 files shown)


src/Site/Controller/Redirect.php
```diff
@@ -29,7 +29,7 @@ public function view(
         } catch (\Exception $e) {
             // Don't know what went wrong, so go to logout route to clear the session
             // This will then redirect to the login page
```
megahirt, line 31:
```suggestion
            // Don't know what went wrong, so redirect to projects page
```

megahirt (approved):
Well done on this refactor!  Looks like we've run into this now that our delete project feature is getting some use.

laineyhm:
✅ Went through steps 1-5 with no surprises.

PR #1731, opened 2023-02-16
Update field names to be consistent with FLEx

longrunningprocess:
## Description

| LF labels | FLEx labels |
| --------- | ----------- |
| Word       | Lexeme Form |
| Meaning | Sense              |
| Part of Speech | Grammatical Category |
| Type | Sense Type |

## Checklist

- [x] I have labeled my PR with: bug, feature, engineering, security fix or testing
- [x] I have performed a self-review of my own code
- [x] I have reviewed the title & description of this PR which I will use as the squashed PR commit message
- [x] I have updated tests that prove my fix is effective or that my feature works
- [x] I have enabled auto-merge (optional)

## Testing

Testers, use the following instructions against our staging environment. Post your findings as a comment and include any meaningful screenshots, etc.

- Ensure field names referenced in table above have changed to the correct label
- Ensure there are no more old field name references in the project configuration area

PR #1753, opened 2023-05-23
Revert "Update field names to be consistent with FLEx (#1731)"

longrunningprocess:
This reverts commit 3927735a1779516f65c003d8f7d8f2f72c42f344.

## Description

We came to a realization about the significance of the field names and the fact that some were stored in the database that led us to need to back this change out until a later time.

## Checklist

## Testing

CI tests should be sufficient.

### DIFF
--- a/src/Api/Model/Languageforge/Lexicon/Config/LexConfig.php
+++ b/src/Api/Model/Languageforge/Lexicon/Config/LexConfig.php
@@ -127,14 +127,14 @@ class LexConfig extends ObjectForEncoding
      * @var array
      */
     private static $flexOptionlistNames = [
-        "grammatical-info" => "Grammatical Category",
+        "grammatical-info" => "Part of Speech",
         "semantic-domain-ddp4" => "Semantic Domain",
         "domain-type" => "Academic Domains",
         self::ENVIRONMENTS => "Environments",
         "location" => "Location",
         "usage-type" => "Usages",
         "reversal-type" => "Reversal Entries",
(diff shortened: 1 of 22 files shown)


hahn-kev:
It looks fine to me, but it's a lot of code so it's hard to tell if something was missed or not, but I'd assume you just did a revert so It's not like you made these changes by hand this time. As for testing I never setup my local environment. I'll wait and try to test it out on our qa environment.

myieye:
The only 2 strings hardcoded into LFMerge are "Part of Speech" and "Type" and they're both used as the names of option lists.
I just did a sync of `test-chris-02` into a fresh local instance of LF and both of those option lists are working fine for me. So, the only other case I can think of that might be worth testing would be to create a **new** LD project (with no existing option lists?) and sync that, but I don't think that that would be more fragile than what I've already tried.

myieye:
I was also able to sync changes to those option lists back to LD/FLEx.

longrunningprocess:
> Actually...it's not obvious to me why we want to revert this. The change didn't fix existing projects, but as long as it didn't break anything then at least it makes **new** projects consistent with FLEx. If we're asked again why the labels don't match we could say that we did everything we can easily do.

@hahn-kev  requested it and I do believe it had to do with concerns around some inconsistencies in the code... maybe you and he can discuss it more and decide whether you all want to approve this or not?

hahn-kev:
Go ahead and continue @longrunningprocess, we discussed it and agreed to rollback the change.

PR #1763, opened 2023-06-21
Prevent UI updates when there is an unresolved recording in progress

### LINKED ISSUE #1762: bug: UI refreshes often cause audio loss
**Describe the bug**
We have multiple mechanisms in place for refreshing entires/the UI due to various reasons. But there is no safety measure to prevent that from happening while a user is recording or after they have recorded, but haven't yet saved the recording.

**To Reproduce**

1. Start recording
2. Make a change in a text field
3. Wait... 💥

One could argue that a user isn't likely to make a change while they're recording. But there are various things that can trigger the same problem that are much less obvious.

**Expected behavior**
Recording should never be interrupted or lost unless the user explicitly navigates away from the entry/page without resolving it.

myieye:
## Description

See #1762 

## Checklist

## Testing

1) Make various changes to entries while recording, while uploading or anywhere in between and make sure the recording never gets lost.

### DIFF
--- a/src/angular-app/languageforge/lexicon/editor/editor.component.ts
+++ b/src/angular-app/languageforge/lexicon/editor/editor.component.ts
@@ -34,6 +34,7 @@ import { LexOptionList } from '../shared/model/option-list.model';
 import { FieldControl } from './field/field-control.model';
 import { OfflineCacheUtilsService } from '../../../bellows/core/offline/offline-cache-utils.service';
 import { IDeferred } from 'angular';
+import { RecordingStateService } from './recording-state.service';
 
 class Show {
   more: () => void;
@@ -94,6 +95,7 @@ export class LexiconEditorController implements angular.IController {
     'lexRightsService',
     'lexSendReceive',
     'offlineCacheUtils',
+    'recordingStateService',
   ];
 
   constructor(private readonly $filter: angular.IFilterService,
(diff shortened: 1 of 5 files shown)


src/angular-app/languageforge/lexicon/editor/editor.component.ts
```diff
@@ -369,6 +372,10 @@ export class LexiconEditorController implements angular.IController {
     // entry and is NOT going to a different entry (as is the case with editing another entry.
     let newEntryTempId: string;
 
+    // We might be saving, because the user is navigating away from this entry,
+    // in which case we don't want to lose the upload.
+    await this.recordingStateService.uploading$();
```
myieye, line 377: @rmunn So, we either: (1) ditch the audio and hope as much gets saved as possible (2) wait for the audio at the risk of the save then never getting run at all or (3) put effort into configuring when we wait.

myieye: It's worth mentioning that this is actually just a bonus. The `return` on 1141 is much more significant and effective.

rmunn (approved):
LGTM. The bit where we block UI updates if an upload is in progress might in theory have a better solution, but implementing that better solution would involve a LOT more work than it's worth considering that nobody will notice the UI updates being blocked in 99.8% of cases. So let's get this merged.

PR #1772, opened 2023-09-15
Treat missing isDeleted field as false

### LINKED ISSUE #1770: User admin page only lists a fraction of the users
Tried to find a user today and noticed that the user admin page says there's only 290 users in Language Forge. That's not correct, there's a lot more. It also means we can't search for those other users.

rmunn:
Fixes bug where missing isDeleted field was causing users to *not* show up in active-user lists.

Fixes #1770.

### DIFF
--- a/src/Api/Model/Shared/UserListModel.php
+++ b/src/Api/Model/Shared/UserListModel.php
@@ -10,7 +10,7 @@ class UserListModel extends MapperListModel
     {
         parent::__construct(
             UserModelMongoMapper::instance(),
-            ["username" => ['$regex' => ""], "isDeleted" => ["$in" => [false, null]]],
+            ["username" => ['$regex' => ""], "isDeleted" => ['$in' => [false, null]]],
             ["username", "email", "name", "avatar_ref", "role", "projects", "active"]
         );
     }
--- a/src/Api/Model/Shared/UserListModel.php
+++ b/src/Api/Model/Shared/UserListModel.php
@@ -10,7 +10,7 @@ class UserListModel extends MapperListModel
     {
         parent::__construct(
             UserModelMongoMapper::instance(),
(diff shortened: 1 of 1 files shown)


src/Api/Model/Shared/UserListModel.php
```diff
@@ -10,7 +10,7 @@ public function __construct()
     {
         parent::__construct(
             UserModelMongoMapper::instance(),
-            ["username" => ['$regex' => ""], "isDeleted" => false],
+            ["username" => ['$regex' => ""], "isDeleted" => ["$in" => [false, null]]],
```
megahirt, line 13: This looks correct, however a failing unit test reveals that this likely needs to be in single quotes, otherwise $in is interpreted as a PHP variable instead of a mongo operator.
```suggestion
            ["username" => ['$regex' => ""], "isDeleted" => ['$in' => [false, null]]],
```

megahirt (changes requested):
Thanks for fixing this, Robin.  It looks like it just needs single quotes.  Also needs testing as I didn't try and test the change.

src/Api/Model/Shared/UserListModel.php
megahirt: Unit test failure:
```
There was 1 error:

1) MultipleModelTest::testUserList_HadOnlyUsers
Undefined variable: in

/var/www/html/Api/Model/Shared/UserListModel.php:13
/var/www/test/php/model/shared/MultipleModelTest.php:38
phpvfscomposer:///var/www/html/vendor/phpunit/phpunit/phpunit:97
```

rmunn:
Code formatting check is failing on three SCSS files not touched in this PR. All other checks passing now that Chris caught the single-quotes issue.

PR #1779, opened 2023-09-25
Fix 500 server error on project dashboard

### LINKED ISSUE #1773: Viewing project page throws a 500
open this page and there's a 500 error
https://languageforge.org/projects/smk-flex

### LINKED ISSUE #1775: bug: PHP lex_stats call shouldn't return entire entries list
See https://github.com/sillsdev/web-languageforge/issues/1773#issuecomment-1725240972 for details. On a project with more than 30,000 entries, `lex_stats` is trying to return *all* the entries and running out of memory. It should only be returning an entry count, plus a count of how many entries have pictures and how many have audio, both of which would be easy to get with a simple Mongo query.

rmunn:
## Description

Large projects were returning an HTTP 500 on the project dashboard, because PHP was running out of memory trying to return every entry for stats to be calculated client-side. This changes the stats to be calculated server-side instead. We get rid of the "number of entries with audio" stat because it's not possible to calculate in a simple Mongo query and would require serializing the entire entry database server-side, causing the very memory error we're trying to get rid of.

Fixes #1773.
Fixes #1775.

## Checklist

## Testing

- Load a large project (via Send/Receive) into Language Forge
- In that project, click on the project name in "My Projects -> (project name)" in the upper left

If the bug is not fixed, you'll get a 500 Internal Server Error page. If the bug is fixed, you'll get the dashboard.

### DIFF
--- a/src/Api/Model/Shared/Mapper/MongoQueries.php
+++ b/src/Api/Model/Shared/Mapper/MongoQueries.php
@@ -4,16 +4,14 @@ namespace Api\Model\Shared\Mapper;
 
 class MongoQueries
 {
-    public static function countEntries($databaseName, $collectionName)
+    public static function countEntries($db, $collectionName)
     {
-        $db = MongoStore::connect($databaseName);
         $coll = $db->selectCollection($collectionName);
         return $coll->count();
     }
 
-    public static function countEntriesWithPictures($databaseName, $collectionName)
+    public static function countEntriesWithPictures($db, $collectionName)
     {
-        $db = MongoStore::connect($databaseName);
         $coll = $db->selectCollection($collectionName);
         $query = [
             "senses" => ['$exists' => true, '$ne' => []],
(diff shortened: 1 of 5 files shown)


src/Api/Model/Shared/Mapper/MongoQueries.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace Api\Model\Shared\Mapper;
+
+class MongoQueries
+{
+    public static function countEntries($databaseName, $collectionName)
+    {
+        $db = MongoStore::connect($databaseName);
```
hahn-kev, line 9: not sure what the cost is, but should we be connecting to the db in each of these functions? would it be better to connect once in the encode function and pass the db into each count method?

hahn-kev (approved):
Looks good to me, just one minor suggestion as it seems weird to me to open a connection 3 times for 3 queries happening one after another.

src/Api/Model/Shared/Mapper/MongoQueries.php
rmunn: Good idea. Done.

PR #1781, opened 2023-11-03
Add reset cache button to advanced project configuration tab

### LINKED ISSUE #1774: Interupting loading the dictionary the first time leaves users in a state they can't recover from
When you load a project for the first time it loads all the entries in chunks. If this process is interrupted then the number of entries will not match what's in the database. The only way for a user to recover from this state is to delete the cache for the language forge site. This is both difficult to explain and invalidates any other projects that have been cached successfully.

Potential solutions:

1. add a clear cache button which would clear the cache for the current project and reload the page.
2. allow resuming from an interrupted first load. It may be possible if the entries are loaded by a stable value like created date for the frontend to just resume by requesting values created after the last entree received when loading.

One way we could detect this state automatically is by comparing the total entry count locally with what's in mongo. If we implement solution 2 we also need solution 1 as a fallback.

rmunn:
## Description

Add a button to reset the browser storage cache, in case users need to (e.g., an interrupted download of the project dictionary).

## Screenshots

image

## Checklist

## Testing

- Go to a project's Configuration options
- Go to the advanced configuration tab
- Open Developer Tools, look at Local Storage in Application tab
- Verify there are entries and comments
- Click the red "Reset browser storage" button
- Verify the entries and comments are gone from local storage

### DIFF
--- a/src/angular-app/languageforge/lexicon/settings/configuration/configuration-advanced-options.component.html
+++ b/src/angular-app/languageforge/lexicon/settings/configuration/configuration-advanced-options.component.html
@@ -4,5 +4,15 @@
         <p>Please do not change these settings unless asked to by tech support.</p>
         <label for="updateIntervalSeconds">Update interval (in seconds)</label>
         <input type="number" ng-model="$ctrl.accPollUpdateTimerSecondsDirty">
+        <br>
+        <p>If something went wrong with downloading the dictionary and it only loaded a partial copy (i.e. some dictionary entries are missing), the button below can let you reset the browser storage, removing the partial copy of the dictionary stored in your browser so that you can download it again.</p>
+        <p><strong>CAUTION:</strong> Make sure you are online before doing this, otherwise you won't be able to redownload the dictionary and you'll see an empty project until you can go online again.</p>
+        <button
+            class="btn btn-danger"
+            type="button"
+            data-ng-click="$ctrl.resetLocalStorage()"
+            >
+            Reset browser storage
+        </button>
     </div>
 </div>
--- a/src/angular-app/languageforge/lexicon/settings/configuration/configuration-advanced-options.component.html
+++ b/src/angular-app/languageforge/lexicon/settings/configuration/configuration-advanced-options.component.html
@@ -7,12 +7,12 @@
         <br>
         <p>If something went wrong with downloading the dictionary and it only loaded a partial copy (i.e. some dictionary entries are missing), the button below can let you reset the browser storage, removing the partial copy of the dictionary stored in your browser so that you can download it again.</p>
         <p><strong>CAUTION:</strong> Make sure you are online before doing this, otherwise you won't be able to redownload the dictionary and you'll see an empty project until you can go online again.</p>
-        <button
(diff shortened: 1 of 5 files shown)


src/angular-app/languageforge/lexicon/settings/configuration/configuration-advanced-options.component.ts
```diff
@@ -18,6 +24,11 @@ export class AdvancedOptionsConfigurationController implements angular.IControll
     );
   }
 
+  async resetLocalStorage() {
+    await this.editorOfflineCache.deleteAllEntries();
+    await this.commentsOfflineCache.deleteAllComments();
```
hahn-kev, line 29: should we reload the page after the reset? That'll safely clear any memory that's still holding on to data that was just cleared and it'll start loading the dictionary again.

myieye: It seemed OK to me. When you navigate back to the editor it downloads the dictionary.
I couldn't break it by going to other configuration tabs either.

So, I think the implemented feature is working well 👍 🚀.

However, there's (aaaalmost) zero feedback when you click the button.
The simplest way to get feedback would be (1) an alert or (2) just reload the page as Kevin suggested, as a result the user will see the dialog indicating that the dictionary is being downloaded, which is probably more meaningful and satisfying than an alert. So, I think reloading the page is a good idea.

I'm sure you could start the download without reloading the page, but if it takes more than 2 minutes to implement it's probably not worth it 😆.

Can we also put a pointer cursor on that button? Pretty please 😁

myieye (approved):

src/angular-app/languageforge/lexicon/settings/configuration/configuration-advanced-options.component.html
```diff
@@ -4,5 +4,15 @@ <h2>Caution!</h2>
         <p>Please do not change these settings unless asked to by tech support.</p>
         <label for="updateIntervalSeconds">Update interval (in seconds)</label>
         <input type="number" ng-model="$ctrl.accPollUpdateTimerSecondsDirty">
+        <br>
+        <p>If something went wrong with downloading the dictionary and it only loaded a partial copy (i.e. some dictionary entries are missing), the button below can let you reset the browser storage, removing the partial copy of the dictionary stored in your browser so that you can download it again.</p>
+        <p><strong>CAUTION:</strong> Make sure you are online before doing this, otherwise you won't be able to redownload the dictionary and you'll see an empty project until you can go online again.</p>
+        <a href="#"
```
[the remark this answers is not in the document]
myieye: I now see that this actually links back to the editor...also cool 😎 👍

src/angular-app/languageforge/lexicon/settings/configuration/configuration-advanced-options.component.ts
rmunn: Done. The reset now navigates back to the project root then does a browser reload, so that the redownload is automatically triggered.

PR #1785, opened 2023-11-08
use kustomize instead of sed for environment configuration

### LINKED ISSUE #1783: change k8s yaml parameters (like WEBSITE) to use kustomize instead of sed at the commandline to substitue values
right now the only want to have different config values between staging and prod is by using sed in the deployment make file. This sucks because if we want to add another parameter we need to dig through those scripts and modify them. It would be much simpler to use kustomize like we do in lexbox to just patch the differences we want instead.

hahn-kev:
## Description
refactored our k8s resources to use kustomize to simplify having different configs per environment

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works

to test this I used `kubectl --context dallas-rke diff -k .` in the staging folder (and similar for prod) to verify that nothing unexpected changes.


rmunn:
Still reading through this, but one comment I have is that you have commands like `kubectl --context aws-rke` in the Makefile. That's a good idea, but we'll need to add something to the README saying "You need to name your kubectl contexts by these names" or else people who have a different name in their .kubectl will have the command fail by default. And I'd also suggest changing the context names to something that includes "languageforge" or "lf" in it, because in my .kubectl I have the following:

```plaintext
- context:
    cluster: aws-rke
    namespace: languagedepot
    user: aws-rke
  name: languagedepot-rke
- context:
    cluster: aws-rke
    namespace: languageforge
    user: aws-rke
  name: languageforge-rke
```

I should actually rename those to something like `language{forge,depot}-prod` or `{lf,ld}-prod` but haven't gotten around to it. The point is that the kubectl context can carry a default namespace along with it, and I use that so I don't see languageforge pods when I'm deploying languagedepot, and vice versa.

rmunn (approved):
Looks good to me. I believe the GHA workflows will continue to work with no modifications with this approach, though of course we'll need to see it in action to be 100% certain. But it looks like you have everything covered that I could think of.

hahn-kev:
That's a good point about the context names. I'll update the readme.

PR #1809, opened 2024-05-06
Fix one failing E2E test on Firefox, skip the other one

### LINKED ISSUE #1801: debt: 2 failing Firefox playwright E2E tests after npm packages
Curiously, after [updating our npm dependencies](https://github.com/sillsdev/web-languageforge/pull/1799) there are two Firefox playwright E2E tests that consistently [fail](https://github.com/sillsdev/web-languageforge/actions/runs/8890540339/job/24415644965):

> 2 failed
> [firefox] › editor/editor.entries.spec.ts:207:7 › Editor entries › Create new word, modify new word, autosaves changes, new word visible in editor and list
> [firefox] › editor/editor.pictures.spec.ts:53:7 › Editor pictures › Showing and hiding captions
> 17 passed (3.1m)

rmunn:
## Description

We currently have two E2E tests that fail on Firefox while they pass on Chrome.

One of them appears to be caused by `reload()` not triggering our page autosave when it happens immediately after a `locator.fill()`. I suspect that the `reload()` isn't triggering the `blur` event which would allow our autosave to trigger, but only on Firefox (for reasons that would take so long to determine that I figure it's not worth spending the time). I solved this one by clicking on another entry before running the `reload()`, which forces the just-entered data to autosave and thus the rest of the test passes as it should.

The other test appears to be running into [a Playwright bug with `locator.fill()` on Firefox](https://www.github.com/microsoft/playwright/issues/29614), where the locator thinks the input field (or textarea in this case) is not editable. This happens even if you put `expect(...).toBeEditable()` immediately before the `.fill()`. The `expect` passes, but then the locator waits forever (or rather, until the test times out) for the text area to be editable, for some reason not realizing that it is editable. Since this appears to be caused by a Playwright and/or Firefox bug rather than by our code, and the test passes in Chromium, I decided to simply have this test skipped on Firefox until such time as the `locator.fill()` bug gets resolved.

## Checklist

## Testing

- Run `npm run test-e2e`; all E2E tests should now pass.

### DIFF
--- a/test/e2e/tests/editor/editor.pictures.spec.ts
+++ b/test/e2e/tests/editor/editor.pictures.spec.ts
@@ -51,6 +51,7 @@ test.describe('Editor pictures', () => {
   });
 
   test('Showing and hiding captions', async ({ managerTab, browserName }) => {
+    test.slow(browserName === 'firefox');
     const configurationPage = new ConfigurationPageFieldsTab(managerTab, project());
 
     await test.step('Hide empty captions', async () => {
--- a/test/e2e/tests/editor/editor.pictures.spec.ts
+++ b/test/e2e/tests/editor/editor.pictures.spec.ts
@@ -7,7 +7,7 @@ import { Project } from '../../utils';
 
 test.describe('Editor pictures', () => {
 
-  const { project } = defaultProject();
+  const { project, entryIds } = defaultProject();
   const newProject = projectPerTest(true);
 
   let editorPageManager: EditorPage;
@@ -51,8 +51,6 @@ test.describe('Editor pictures', () => {
   });
 
   test('Showing and hiding captions', async ({ managerTab, browserName }) => {
-    test.slow(browserName === 'firefox');
-
     const configurationPage = new ConfigurationPageFieldsTab(managerTab, project());
 
     await test.step('Hide empty captions', async () => {
@@ -64,7 +62,7 @@ test.describe('Editor pictures', () => {
     });
 
     const caption = await test.step('Non-empty caption is visible', async () => {
-      await editorPageManager.goto();
+      await editorPageManager.goto({ entryId: entryIds()[0] });
       await editorPageManager.showExtraFields(false);
       const picture = editorPageManager.picture(entries.entry1.senses[0].pictures[0].fileName);
       const caption = editorPageManager.caption(picture);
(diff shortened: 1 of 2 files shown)


test/e2e/tests/editor/editor.pictures.spec.ts
```diff
@@ -51,8 +51,6 @@ test.describe('Editor pictures', () => {
   });
 
   test('Showing and hiding captions', async ({ managerTab, browserName }) => {
-    test.slow(browserName === 'firefox');
```
myieye, line 54: Why did you remove this? We really don't want to have to debug this again, because it sometimes takes too long on Firefox. I'd say just leave it unless you're confident it's really not necessary.

test/e2e/tests/editor/editor.pictures.spec.ts
```diff
@@ -76,7 +74,11 @@ test.describe('Editor pictures', () => {
       await expect(caption).toBeVisible(); // it disappears immediately which could be annoying
       await caption.fill('');
       await expect(caption).not.toBeVisible(); // it disappears immediately which could be annoying
+      // Navigate away to force changes to be saved
```
myieye, line 77: Did you try just doing something like `caption.blur()`? (I'm not sure if that method exists, but surely there's an API somewhere that'll do it)

megahirt (commented):
This is great detective work, Robin!  Once you've resolved @myieye 's comments I'd like to get this merged so that future PRs can benefit from the green this will bring back.

test/e2e/tests/editor/editor.pictures.spec.ts
rmunn: Restored in commit dba024ceb. I removed it because it didn't seem necessary in my test runs on my local machine, but it's true that GitHub Actions runners are often slower, so I put it back.

test/e2e/tests/editor/editor.pictures.spec.ts
rmunn: There is a `locator.blur()` method, and I didn't think to try that. I'll try it and see if that's enough.

rmunn: Interestingly, doing `caption.blur()` immediately after `caption.fill('')` caused the test to fail because "Error: locator.blur: Target page, context or browser has been closed". Which I think is incorrect; I think the actual problem is that as soon as you empty the caption, it gets hidden, so `.blur()` suddenly can't find the locator. I'll try again with `if (caption.isVisible()) { caption.blur() }` and see what that does.

rmunn: Well, `caption.blur()` didn't help. Probably because the caption immediately goes invisible and so the Playwright locator can't call `.blur()` on it, and so Firefox isn't sending the blur event. I'll try one more thing: calling .focus() on a different locator, which should cause the blur event to be fired on the caption. If that fails I'll stick with the code as-is.

rmunn: Calling `.focus()` on the lexeme field didn't work on Firefox either. I'm going to leave this as-is, because I have not gotten the simpler alternatives to work and this is simple enough, all things considered.

rmunn: @myieye - Please feel free to resolve this conversation if you're satisfied. I'm leaving it open so it's easier for you to find when you do a re-review.

myieye (approved):

PR #1812, opened 2024-05-16
Base php 7.4

megahirt:
This is a PR to update the base-php image to 7.4.  We will not update to future PHP versions, so it makes sense to just name this base-php

We also add the mongodb-org-tools apt package to the base-php image for use in the backup/restore planned task.

Once this is merged, I can manually create a new base-php image via GHA.

TODO:

- [x] - figure out what debian package has mongoexport in it - which version of debian are we running?
- [x] - `make build-base-php` locally
- [x] - `make` locally
- [x] - run PHP tests locally
- [x] - confirm that `mongoexport` is available in the running app container

PHP tests complete successfully on my local machine, but will fail in this PR build since base-php hasn't been built properly until this is merged.
image

### DIFF
--- a/docker/app/Dockerfile
+++ b/docker/app/Dockerfile
@@ -38,7 +38,7 @@ RUN npm run build:${NPM_BUILD_SUFFIX}
 # COMPOSER-BUILDER
 # download composer app dependencies
 # git - needed for composer install
-FROM sillsdev/web-languageforge:base-php-7.4 AS composer-builder
+FROM sillsdev/web-languageforge:base-php AS composer-builder
 ENV COMPOSER_ALLOW_SUPERUSER=1
 RUN apt-get update && apt-get install -y git && rm -rf /var/lib/apt/lists/* \
     && install-php-extensions @composer
@@ -47,7 +47,7 @@ COPY src/composer.json src/composer.lock /composer/
 RUN composer install
 
 # PRODUCTION IMAGE
-FROM sillsdev/web-languageforge:base-php-7.4 AS production-app
+FROM sillsdev/web-languageforge:base-php AS production-app
 RUN rm /usr/local/bin/install-php-extensions
 RUN apt-get remove -y gnupg2
 RUN mv $PHP_INI_DIR/php.ini-production $PHP_INI_DIR/php.ini
@@ -55,7 +55,7 @@ RUN mv $PHP_INI_DIR/php.ini-production $PHP_INI_DIR/php.ini
 COPY --from=sillsdev/web-languageforge:wait-latest /wait /wait
 
 # DEVELOPMENT IMAGE
-FROM sillsdev/web-languageforge:base-php-7.4 AS development-app
+FROM sillsdev/web-languageforge:base-php AS development-app
 RUN install-php-extensions xdebug-^3.1
 COPY docker/app/docker-php-ext-xdebug.ini /usr/local/etc/php/conf.d
 RUN mv $PHP_INI_DIR/php.ini-development $PHP_INI_DIR/php.ini
--- a/docker/app/Dockerfile
+++ b/docker/app/Dockerfile
@@ -40,8 +40,6 @@ RUN npm run build:${NPM_BUILD_SUFFIX}
 # git - needed for composer install
 FROM sillsdev/web-languageforge:base-php AS composer-builder
 ENV COMPOSER_ALLOW_SUPERUSER=1
-RUN apt-get update && apt-get install -y git && rm -rf /var/lib/apt/lists/* \
-    && install-php-extensions @composer
 WORKDIR /composer
 COPY src/composer.json src/composer.lock /composer/
 RUN composer install
@@ -49,7 +47,8 @@ RUN composer install
 # PRODUCTION IMAGE
 FROM sillsdev/web-languageforge:base-php AS production-app
 RUN rm /usr/local/bin/install-php-extensions
-RUN apt-get remove -y gnupg2
+RUN apt-get remove -y gnupg2 git
+# TODO: remove composer - how do we do this?
 RUN mv $PHP_INI_DIR/php.ini-production $PHP_INI_DIR/php.ini
 # had to add /wait into prod image so the prod image could be run through E2E tests.
 COPY --from=sillsdev/web-languageforge:wait-latest /wait /wait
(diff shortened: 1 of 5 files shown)


megahirt:
This PR tests are not expected to pass because we are changing back to the `base-php` tag from the current `base-php-7.4` tag.  This has the effect that unit tests are now running against the old `base-php` image which is PHP 7.3 and fail.  Once this PR is merged, I can manually push a new base-php image that is updated to 7.4 and then tests will begin passing again.

docker/app/Dockerfile
```diff
@@ -38,24 +38,22 @@ RUN npm run build:${NPM_BUILD_SUFFIX}
 # COMPOSER-BUILDER
 # download composer app dependencies
 # git - needed for composer install
-FROM sillsdev/web-languageforge:base-php-7.4 AS composer-builder
+FROM sillsdev/web-languageforge:base-php AS composer-builder
 ENV COMPOSER_ALLOW_SUPERUSER=1
-RUN apt-get update && apt-get install -y git && rm -rf /var/lib/apt/lists/* \
```
megahirt, line 43: this step has been moved to base-php

docker/app/Dockerfile
```diff
@@ -38,24 +38,22 @@ RUN npm run build:${NPM_BUILD_SUFFIX}
...
 WORKDIR /composer
 COPY src/composer.json src/composer.lock /composer/
 RUN composer install
 
 # PRODUCTION IMAGE
-FROM sillsdev/web-languageforge:base-php-7.4 AS production-app
-RUN rm /usr/local/bin/install-php-extensions
-RUN apt-get remove -y gnupg2
+FROM sillsdev/web-languageforge:base-php AS production-app
+RUN rm /usr/local/bin/install-php-extensions /usr/local/bin/composer
```
megahirt, line 49: we should remove composer from the production image - should have done this earlier but didn't realize it

docker/base-php/Dockerfile
```diff
@@ -1,16 +1,27 @@
...
+# mongodb-mongosh - contains mongosh which is useful for querying the db from the app container
+RUN curl -fsSL https://www.mongodb.org/static/pgp/server-6.0.asc | gpg -o /usr/share/keyrings/mongodb-server-6.0.gpg --dearmor
+RUN echo "deb [ signed-by=/usr/share/keyrings/mongodb-server-6.0.gpg] http://repo.mongodb.org/apt/debian bullseye/mongodb-org/6.0 main" | tee /etc/apt/sources.list.d/mongodb-org-6.0.list
+RUN apt-get update && apt-get -y install mongodb-org-tools mongodb-mongosh && rm -rf /var/lib/apt/lists/*
 
 # see https://github.com/mlocati/docker-php-extension-installer
 # PHP extensions required by the LF application
 COPY --from=mlocati/php-extension-installer /usr/bin/install-php-extensions /usr/local/bin/
-RUN install-php-extensions mongodb intl
+RUN install-php-extensions mongodb intl @composer
```
megahirt, line 24: composer is needed by all PHP images (test-php and app) and so it makes sense to move it into base-php

docker/test-php/Dockerfile
```diff
@@ -1,15 +1,13 @@
...
 # ----- LINES BELOW COPIED FROM APP DOCKERFILE ----------
 WORKDIR /var/www/html
 ENV COMPOSER_ALLOW_SUPERUSER=1
-RUN apt-get update && apt-get install -y git && rm -rf /var/lib/apt/lists/* \
-    && install-php-extensions @composer
 COPY src/composer.json src/composer.lock /var/www/html/
 RUN composer install
 
-RUN install-php-extensions xdebug
+RUN install-php-extensions xdebug-^3.1
```
megahirt, line 10: For whatever reason this didn't specify the xdebug version as we do in the development app build

docker/base-php/Dockerfile
rmunn: If we do this, then the composer PHP extension will be available in production as well. Is that a security issue? What could an attacked do with the extension if the command-line tool isn't there?

rmunn (approved):
LGTM. One concern about the composer extension, but I don't think that should delay getting this merged into develop; we can back that one change out later if needed.

docker/base-php/Dockerfile
```diff
@@ -1,16 +1,27 @@
...
 # curl - used by LF application
 # ffmpeg - used by LF audio upload method
-RUN apt-get update && apt-get -y install p7zip-full unzip curl tini ffmpeg && rm -rf /var/lib/apt/lists/*
+RUN apt-get update && apt-get -y install gnupg git p7zip-full unzip curl tini ffmpeg
+
+# Install MongoDB shell and tools (must be done after gnupg is installed)
+# mongodb-org-tools - contains mongoimport/mongoexport used by LF backup/restore project commands
+# mongodb-mongosh - contains mongosh which is useful for querying the db from the app container
+RUN curl -fsSL https://www.mongodb.org/static/pgp/server-6.0.asc | gpg -o /usr/share/keyrings/mongodb-server-6.0.gpg --dearmor
+RUN echo "deb [ signed-by=/usr/share/keyrings/mongodb-server-6.0.gpg] http://repo.mongodb.org/apt/debian bullseye/mongodb-org/6.0 main" | tee /etc/apt/sources.list.d/mongodb-org-6.0.list
```
rmunn, line 18:
```suggestion
RUN echo "deb [ signed-by=/usr/share/keyrings/mongodb-server-6.0.gpg ] http://repo.mongodb.org/apt/debian bullseye/mongodb-org/6.0 main" | tee /etc/apt/sources.list.d/mongodb-org-6.0.list
```

I don't know if the space before the `]` is required, but it's possible that this is causing the failed .deb package.

docker/base-php/Dockerfile
megahirt: as it turns out, it is already available in production.  I have removed it in this PR on line 52 of the app dockerfile.  See note above.  So this PR improves security by removing composer from our production image.

docker/base-php/Dockerfile
megahirt: it didn't cause the failure (that was due to trying to build arm64) but this is a good addition so I will accept the suggestion.

PR #1815, opened 2024-05-21
Get rid of symfony-exceptions.patch

### LINKED ISSUE #1814: debt: Get rid of Symfony exception-handling patch
A while back we were having problems with the exception handler in Symfony expecting exceptions to have base class Exception, but receiving base class Error instead. So we created `docker/app/symfony-exceptions.patch` to fix that. Now we may be able to get rid of that rather hacky solution, as it seems that the exceptions PHP ends up throwing are finally going to be base class Exception.

Get rid of the hackish symfony-exceptions.patch, finally.

Leaving the file in the repo and just commenting out the Dockerfile line that uses it.

rmunn:
## Description

We can finally get rid of this hacky "patch a vendor library in place" solution now that PHP exceptions are deriving from the Exception class rather than the Error class.

## Screenshots

None; no behavior change.

## Checklist

## Testing

_Describe how to verify your changes and provide any necessary test data._

- Run `make` and make sure the site still works.
- If MongoDB auth is enabled, try changing the `MONGODB_PASS` env var to an incorrect value to see an exception when you try to go to http://localhost:8080/auth/login


megahirt (changes requested):
How about we just comment out this feature instead of removing it.  That way we can still use it when debugging PHP and need to see what the internal Error is.

rmunn:
@megahirt - Done. I've brought the patch file back, and simply commented out the lines in the Dockerfile that would apply it. So if this exception-swallowing behavior comes back in the future, it'll be obvious from reading the Dockerfile what to do about it.

megahirt (approved):
LGTM!

PR #1817, opened 2024-05-22
Fix "implicit any" Typescript errors to prepare for Typescript 5.5

### LINKED ISSUE #1800: debt: suppressImplicitAnyIndexErrors will be disallowed in Typescript 5.5
**Describe the code debt**
We currently have `suppressImplicitAnyIndexErrors` in our tsconfig.json since we have a lot of implicit `any` types scattered throughout our code. That option [will be going away in Typescript 5.5](https://www.github.com/microsoft/TypeScript/issues/51909), so we'll need to turn `suppressImplicitAnyIndexErrors` off, see where these errors are occurring in our code, and fix them. Typescript says the fix is to "Add an index signature to the relevant type (or use a type assertion at the indexing location)."

**To Fix**
Steps to fix this:

1. Remove `"suppressImplicitAnyIndexErrors": true`, from tsconfig.json
2. Run `make`
3. See what Typescript complains about
4. Fix each error location individually

rmunn:
## Description

We currently have `suppressImplicitAnyIndexErrors` in our tsconfig.json since we have a lot of implicit `any` types scattered throughout our code. That option [will be going away in Typescript 5.5](https://www.github.com/microsoft/TypeScript/issues/51909), so we need to fix them now so that we can continue to upgrade Typescript in the future (5.5 is currently in beta).

## Screenshots

No behavioral changes.

## Checklist

## Testing

- Run `make`, make sure site still loads.
- Check that there aren't more Typescript errors showing up in VS Code.

### DIFF
--- a/src/angular-app/bellows/apps/activity/activity-container.component.ts
+++ b/src/angular-app/bellows/apps/activity/activity-container.component.ts
@@ -42,10 +42,10 @@ class Activity {
   userHrefRelated: string;
   icon: string;
 
-  constructor(data: object = {}) {
+  constructor(data: Record<string, any> = {}) {
     if (data != null) {
       for (const property of Object.keys(data)) {
-        this[property] = data[property];
+        this[property as keyof Activity] = data[property];
       }
     }
   }
@@ -57,6 +57,7 @@ class Activity {
       for (const index in this.content) {
         if (this.content.hasOwnProperty(index)) {
           if (index.startsWith('oldValue')) {
+            // @ts-expect-error Typescript should understand .hasOwnProperty, but doesn't
             return Activity.parseValue(this.content[index]);
           }
         }
@@ -72,6 +73,7 @@ class Activity {
       for (const index in this.content) {
         if (this.content.hasOwnProperty(index)) {
           if (index.startsWith('newValue')) {
+            // @ts-expect-error Typescript should understand .hasOwnProperty, but doesn't
             return Activity.parseValue(this.content[index]);
           }
         }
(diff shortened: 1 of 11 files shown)


src/angular-app/bellows/apps/activity/activity-container.component.ts
```diff
@@ -42,10 +42,10 @@ class Activity {
   userHrefRelated: string;
   icon: string;
 
-  constructor(data: object = {}) {
+  constructor(data: Record<string, any> = {}) {
```
hahn-kev, line 45: if these are supposed to be keys of activity (due to cast below), then why don't we change `string` to `keyof Activity`?

rmunn: Typescript doesn't like that: `Type '{}' is missing the following properties from type 'Record<keyof Activity, any>': summary, action, content, changes, and 22 more.`

hahn-kev: oh right, the properties are optional. It would have to be `Partial<Record...>` then

myieye (approved):

src/angular-app/bellows/apps/activity/activity-container.component.ts
rmunn: Trying `Partial<Record<keyof Activity, any>>` does work for that line, but Typescript still types the `for (const property of Object.keys(data))` has having properties of type `string`, rather than typing `property` as type `keyof Activity`. Which means that the `keyof Activity` typecases further down are still required. Going to leave this as-is since the further type refinements to `data` don't provide any benefit in practice.

PR #1813, opened 2024-05-20
Require username and password for mongodb

rmunn:
## Description

Add env vars `MONGODB_USER` and `MONGODB_PASS` to optionally specify authentication when connecting to MongoDB. Some manual setup on the MongoDB server(s) should be done before this is deployed; see https://github.com/sillsdev/web-languageforge/issues/1787#issuecomment-2119674540 for details. Once the appropriate admin user has been created and Kubernetes secrets set, this PR can then be deployed to staging and production.

Note that https://github.com/sillsdev/LfMerge/pull/336 will need to be merged first, and a new LfMerge version built. The version number in `lfmerge/Dockerfile` can then be bumped as part of this PR.

## Screenshots

None

## Checklist

- [X] I have labeled my PR with: bug, feature, engineering, security fix or testing
- [ ] I have performed a self-review of my own code
- [ ] I have reviewed the title & description of this PR which I will use as the squashed PR commit message
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have enabled auto-merge (optional)

## Testing

Testers, use the following instructions against our staging environment. Post your findings as a comment and include any meaningful screenshots, etc.

- Log in
- Send/Receive a test project if you already have one
- Delete the test project if it exists
- Clone it afresh from Language Depot

If those all continue to pass, then the necessary setup steps (creating an admin user and giving it a password, and creating Kubernetes secrets for that username and password) were correctly completed.

PR #1826, opened 2024-06-27
Kubernetes deployment now requires Mongo auth

rmunn:
This is a followup to https://github.com/sillsdev/web-languageforge/pull/1813.

The `MONGO_INITDB_ROOT_USERNAME` and `MONGO_INITDB_ROOT_PASSWORD` variables, if they exist when the MongoDB container is started up, will be used to tell MongoDB to start in auth-required mode. If there is an existing database then the value of these env vars doesn't matter and the usernames and passwords from the existing database will be used instead, but if the database is empty then these two env vars are used to set up an initial administrator account that can access and change anything, and the DB setup scripts are then expected to create any other required users. Since we now have authentication defined in Mongo, we can turn this on by default and everything will continue to work.

This has already been done manually (via the Rancher control panel) on staging and production; this PR simply adds the required environment variables to the depoyment file so that if anyone runs `make deploy-staging` or `make deploy-prod` in the future, the deployment will not *remove* those environment variables.

### DIFF
--- a/docker/deployment/base/db-deployment.yaml
+++ b/docker/deployment/base/db-deployment.yaml
@@ -69,6 +69,17 @@ spec:
             cpu: 0.1
           limits:
             memory: 400Mi
+        emv:
+          - name: MONGO_INITDB_ROOT_USERNAME
+            valueFrom:
+              secretKeyRef:
+                key: MONGODB_USER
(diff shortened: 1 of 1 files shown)


megahirt (approved):
LGTM

PR #1831, opened 2024-07-05
remove old project board workflow

megahirt:
This workflow is no longer necessary, now that GH implemented the workflow GUI which supports auto-adding issues to projects

It's also currently failing since I removed the ProductOwner view that it references.
image


hahn-kev (approved):
Looks good

PR #1838, opened 2024-10-07
Make project backup script auth-aware

myieye:
It looks like we added MongoDB auth right after the backup script was created, so it was missing this.


rmunn (approved):
Sure, that works. I had another solution in my local repo that wasn't as good, and hadn't figured out a better way to do it, so let's go with this instead. In theory this could expose the password (command-line parameters are visible to other processes, at least on Linux) so in theory this should be done with a password-file arg instead (e.g., read the password from the file provided on the command line), but in practice that's not a serious concern as this only gets run on dev machines.

PR #1842, opened 2025-02-03
Use chg in LfMerge, only in local dev LF for now

### LINKED ISSUE #1841: LF: Use chg in LfMerge, only in local dev for now
Now that LfMerge has built a version that uses `chg` instead of `hg` to speed up Send/Receive, we will want to deploy that. We'll start by *only* changing the `CHORUS_HG_EXE` environment variable in Docker Compose, so that it applies to local LF but doesn't get pushed to any of our k8s deployments. Then once we've tested it and ensured everything still works, we'll take the next step and add it to k8s so the change goes to staging and then prod.

rmunn:
## Description

We'll start using `chg` for LfMerge in local LF development; once we know that it works, we'll add the relevant environment variable to the k8s deployment so that staging and then prod also start getting the same Send/Receive speedups.

## Checklist

## Testing

- Run some Send/Receives on local LF and make sure everything still works.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -107,6 +107,7 @@ services:
       - WAIT_HOSTS=db:27017
       - ENVIRONMENT=development
       - DATABASE=scriptureforge
+      - CHORUS_HG_EXE=chg
       - MONGODB_CONN=mongodb://db:27017
       - MONGODB_AUTHSOURCE=admin
       - LANGUAGE_DEPOT_API_TOKEN=bogus-development-token
--- a/docker/app/Dockerfile
+++ b/docker/app/Dockerfile
@@ -10,6 +10,8 @@ WORKDIR /data
 # Copy in files needed for compilation, located in the repo root
 COPY package.json pnpm-lock.yaml ./
 
+# Install latest corepack version to work around a bug with signatures
+RUN npm install -g corepack@latest
 RUN corepack enable
 RUN pnpm install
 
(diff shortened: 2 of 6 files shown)


hahn-kev (approved):

hahn-kev:
looks like our PNPM install is failing in LF too, would you mind looking into that? I'm fine if you merge this PR without fixing it as they're totally unrelated

rmunn:
> looks like our PNPM install is failing in LF too, would you mind looking into that? I'm fine if you merge this PR without fixing it as they're totally unrelated

Huh. So there's some kind of problem with NPM's signing infrastructure and it's unrelated to the specific PNPM 9.15.5 release (LF is still on 9.11.0). Weird; hopefully they sort it out soon because that's going to affect millions of projects.

I'll try rerunning the failed GHA builds once in case it's already been sorted out, then if it's still failing across the board I'll just merge this.

rmunn:
https://www.github.com/nodejs/corepack/issues/612

rmunn:
Corepack bug present in the corepack versions distributed with NPM (corepack 0.30.0). Running `npm install --global corepack@latest` before running `corepack enable` will update corepack to 0.31.0 which has the bugfix. (PNPM rotated a key, so for a while both the old and new keys were returned from the "send me the key" API endpoint, and IIUC corepack 0.30.0 didn't handle that correctly and only checked against the first key returned, which was the old key.

rmunn:
That solved the PNPM install issue, only to run into a different issue: "Package `libasound2` has no installation candidate" which trying to install browsers for `pnpm exec playwright install`. I'll work on solving that separately, but I'll go ahead and merge this PR since that error is unrelated to this one and is probably already present on `develop`.

